PR #27, opened 2017-05-18
Running Linter as Part of CI Jobs

hiranya911:
* Returning non-zero exit codes from `lint.sh` when lint errors are encountered.
* Configuring Travis to run linter before running tests (it will fail the build, on non-zero exit codes)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ python:
   - "3.5"
 # command to install dependencies
 install: "pip install -r requirements.txt"
-before_script: ./lint.sh all
+before_script: 
+  - export PY_VERSION=`python -c 'import sys; print sys.version_info.major'`
+  - if [ $PY_VERSION == '2' ]; then ./lint.sh all; fi
 # command to run tests
 script: pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,6 @@ python:
   - "3.5"
 # command to install dependencies
 install: "pip install -r requirements.txt"
+before_script: ./lint.sh all
 # command to run tests
 script: pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ python:
 install: "pip install -r requirements.txt"
 before_script: 
   - export PY_VERSION=`python -c 'import sys; print sys.version_info.major'`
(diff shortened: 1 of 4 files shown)


hiranya911:
This turned out to be harder that I expected. 

1. Had to anchor pylint version to 1.6.4. Otherwise Travis picks up the latest version of Pylint (1.7.1) which has a bug where it incorrectly generates warnings for some of our imports. This seems like a [known issue](https://github.com/PyCQA/pylint/issues/1084) which has been fixed for a future release. I also created [another ticket](https://github.com/PyCQA/pylint/issues/1489) for further follow up.
2. Pylint `disable` directives were ignored when running on Python 3.3. This triggered a build break in Travis. We test on Python 2.7, 3.3 and 3.5 on Travis. As a workaround I added some logic in place so that linting will only be performed on Python 2.7.

lint.sh
```diff
@@ -17,6 +17,11 @@
 function lintAllFiles () {
   echo "Running linter on module $1"
   pylint --disable=$2 $1
+  rc=$?
```
wilhuff: If all you want to do is exit if the linter fails then you should just set that mode at the top of the script, e.g. with

```bash
set -o errexit
```

That will get rid of both of these extra blocks.

You might also consider adding `set -o nounset` as well, which will help catch typos in variable names (that would otherwise expand to empty strings).

.travis.yml
```diff
@@ -5,5 +5,8 @@ python:
   - "3.5"
 # command to install dependencies
 install: "pip install -r requirements.txt"
+before_script: 
+  - export PY_VERSION=`python -c 'import sys; print sys.version_info.major'`
+  - if [ $PY_VERSION == '2' ]; then ./lint.sh all; fi
```
wilhuff: If this script is bash you should always perform tests in double brackets to prevent accidental interaction with globbing.

Also, always quote variable expansions to prevent accidental syntax errors if you get something unexpected.

```bash
if [[ "$PY_VERSION" == '2' ]]; then ./lint.sh all; fi
```

lint.sh
hiranya911: Done. I added some more argument validation to make it work better with `nounset` option.

.travis.yml
hiranya911: Done.

PR #34, opened 2017-06-08
Added integration tests for auth module

hiranya911:
#31 introduced an integration test suite to the SDK. But until now we have only had tests for the (experimental) `database` module. This PR adds integration tests for the `auth` module. Additionally:

* Updated `CONTRIBUTING.md` with information about running integration tests.
* Updated `prepare_release.sh` script to run integration tests when creating release builds. The credentials for tests are read from the `scripts/` directory itself.

### DIFF
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -77,10 +77,12 @@ fi
 
 if [[ ! -e "cert.json" ]]; then
     echo "[ERROR] cert.json file is required to run integration tests."
+    exit 1
 fi
 
 if [[ ! -e "apikey.txt" ]]; then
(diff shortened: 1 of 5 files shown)


scripts/prepare_release.sh
```diff
@@ -75,6 +75,14 @@ if [[ `git status --porcelain` ]]; then
     exit 1
 fi
 
+if [[ ! -e "cert.json" ]]; then
+    echo "[ERROR] cert.json file is required to run integration tests."
```
bjornick: Don't you want to exit here?  Same thing in the if clause below

hiranya911: Done

Issue #28, opened 2017-05-24
FR: Realtime Database API


galuszkak:
### FR: Realtime Database API

  * Operating System version: Ubuntu
  * Firebase SDK version: 2.0
  * Firebase Product: database

### Describe the problem

Lack of support for Realtime Database API now in firebase-admin.

hiranya911:
You might be interested to know, that a database API around the existing REST support is currently bring developed at https://github.com/firebase/firebase-admin-python/tree/hkj-simple-db

joshbabb:
@hiranya911, I'm having trouble using your merged (#31) Database API, do you have any examples of using db.py?  For example, just to read?

joshbabb:
PS @galuszkak to access db.py you'll need to use `pip install git+https://github.com/firebase/firebase-admin-python` (and not just `pip install firebase-admin`) until the pip package is updated

hiranya911:
Try something like:

```
app = firebase_admin.initialize_app(cred, {'dbURL' : 'https://your-db-url.firebase.com'})
ref = db.get_reference()
```

However, note that we are still working on this and the API is subject to change.

joshbabb:
Thank you, I'll try this after lunch!

Once it gets closer, I can help provide examples or flesh out documentation
if you'd like. The Python SDK is central to a current work project so I'd
love to give back.

joshbabb:
Also, and sorry if I'm being a bother, but it doesn't appear that there is an implementation in this python SDK for subscribing to and handling events (like .on('child_added') in the node SDK, for example.)  Could I be of any help in extending this for that?

hiranya911:
@joshbabb We are not going to implement realtime event handling capabilities in this iteration. We will consider it for a future release, if there's enough demand for it. But feel free to document your requirements/suggestions as issues in this repo. That will help us to prioritize the features. 

We will also gladly accept any code contributions you can provide us. But if you're going to implement realtime support I'd recommend you to wait until we release what we have already implemented (should be any day now -- we are just finalizing the internal approval processes).

hiranya911:
Also please note that I've just merged #35 into master. You should update your test code as follows:

```
app = firebase_admin.initialize_app(cred, {'databaseURL' : 'https://your-db-url.firebase.com'})
ref = db.reference()
```

joshbabb:
Thanks for the note and all your responses, @hiranya911 

I'll make an issue for the request at a later time.  We're going to hack what we can together for the time being :)

[...]

PR #36, opened 2017-06-21
Implemented support for databaseAuthVariableOverride option

hiranya911:
* Removed `Query.querystr` from the public API
* Reading the `databaseAuthVariableOverride` option when initializing the database client
* Passing the override value as a JSON string to backend server.
* Passing query parameters (the `params` argument) as a string to the `requests` library. This gives us full control over the order and format of parameter encoding.

Related to #28


firebase_admin/db.py
```diff
@@ -544,10 +542,16 @@ class _Client(object):
     marshalling and unmarshalling of JSON data.
     """
```
bjornick: You should document what the arguments are now.  The signature no longer conveys this information

hiranya911: Done

hiranya911:
Made the suggested changes.

PR #38, opened 2017-06-22
Setting the User-Agent header when making HTTP calls.

hiranya911:
Setting the User-Agent header with the following format:

```
Firebase/HTTP/<SDK Version>/<Python Version>AdminPython
```


hiranya911:
#28

PR #39, opened 2017-06-23
Fixing Some Nits in Tests

hiranya911:
* Added `integration/` directory to linter
* Skipping `unused-argument` check for tests. This condition occurs far too often in tests (when setting up pytest fixtures)
* Updated a db integration test case to check for multiple updates.

### DIFF
--- a/tests/test_credentials.py
+++ b/tests/test_credentials.py
@@ -81,7 +81,7 @@ class TestApplicationDefault(object):
 
     @pytest.mark.parametrize('app_default', [testutils.resource_filename('service_account.json')],
                              indirect=True)
-    def test_init(self, app_default): # pylint: disable=unused-argument
+    def test_init(self, app_default):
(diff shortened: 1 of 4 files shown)


tests/test_credentials.py
```diff
@@ -81,7 +81,7 @@ class TestApplicationDefault(object):
 
     @pytest.mark.parametrize('app_default', [testutils.resource_filename('service_account.json')],
                              indirect=True)
-    def test_init(self, app_default): # pylint: disable=unused-argument
+    def test_init(self, app_default):
```
wilhuff: You can suppress these warnings without needing to disable this check for all methods in tests:

For example, here you can
```python
def test_init(self, app_default):
    del app_default
```
If that's just too annoying, carry on.

hiranya911: Done

Issue #53, opened 2017-07-29
ETA for storage or transaction support?


alexanderwhatley:
Hi, I was wondering if/when this API will support writing files to storage, and transaction support. Either of these things would be very useful for the work I am doing. Thanks.

rockwotj:
The Realtime Database just added support for conditional requests (AKA transactions) via the REST API.

The docs for this are here: https://firebase.google.com/docs/reference/rest/database/#section-conditional-requests

We're open to Pull Requests 😄 

/cc @caseycrogers

rockwotj:
Also to track Google Cloud Storage for Firebase integration, you'll want to watch this PR: https://github.com/firebase/firebase-admin-python/pull/44

PR #54, opened 2017-08-02
Add Transaction Support 

alexanderwhatley:
I added in 3 new functions that support the transaction functionality that was added in the REST API recently. These are "get_with_etag", "update_with_etag", and "transaction" in db.py, and added tests for all 3 of these functions. "get_with_etag" is able to retrieve the data at the reference as well as its ETag, and "update_with_etag" takes in data and an ETag value as parameters, and tries to write to the reference using these two.

"transaction" is based off of the transaction function that is in the Admin SDK for Nodejs/Java here: https://firebase.google.com/docs/reference/js/firebase.database.Reference#transaction. I added in the first two parameters, transactionUpdate and onComplete, but I am not sure what purpose the third one could serve in the REST API.

### DIFF
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 """Firebase Realtime Database module.
-
 This module contains functions and classes that facilitate interacting with the Firebase Realtime
 Database. It supports basic data manipulation operations, as well as complex queries such as
 limit queries and range queries. However, it does not support realtime update notifications. This
@@ -42,16 +41,12 @@ _USER_AGENT = 'Firebase/HTTP/{0}/{1}.{2}/AdminPython'.format(
 
 def reference(path='/', app=None):
     """Returns a database Reference representing the node at the specified path.
-
     If no path is specified, this function returns a Reference that represents the database root.
-
     Args:
       path: Path to a node in the Firebase realtime database (optional).
       app: An App instance (optional).
-
     Returns:
       Reference: A newly initialized Reference.
-
     Raises:
       ValueError: If the specified path or app is invalid.
     """
@@ -73,7 +68,6 @@ class Reference(object):
 
     def __init__(self, **kwargs):
         """Creates a new Reference using the provided parameters.
-
         This method is for internal use only. Use db.reference() to obtain an instance of
         Reference.
         """
@@ -102,16 +96,12 @@ class Reference(object):
 
     def child(self, path):
         """Returns a Reference to the specified child node.
-
         The path may point to an immediate child of the current Reference, or a deeply nested
         child. Child paths must not begin with '/'.
-
         Args:
           path: Path to the child node.
-
         Returns:
           Reference: A database Reference representing the specified child node.
-
         Raises:
           ValueError: If the child path is not a string, not well-formed or begins with '/'.
         """
@@ -126,23 +116,27 @@ class Reference(object):
 
     def get(self):
         """Returns the value at the current location of the database.
-
         Returns:
           object: Decoded JSON value of the current database Reference.
-
         Raises:
           ApiCallError: If an error occurs while communicating with the remote database server.
         """
         return self._client.request('get', self._add_suffix())
 
+    def get_with_etag(self):
+        """Returns the value at the current location of the database, along with its ETag.
+        Returns:
+            object: Tuple of the ETag value corresponding to the Reference, and the Decoded JSON value of the current database Reference.
+        Raises:
+            ApiCallError: If an error occurs while communicating with the remote database server.
+        """
+        return self._client.request('get', self._add_suffix(), headers={'X-Firebase-ETag': 'true'})
+
     def set(self, value):
         """Sets the data at this location to the given value.
-
         The value must be JSON-serializable and not None.
-
         Args:
           value: JSON-serialable value to be set at this location.
-
         Raises:
           ValueError: If the value is None.
           TypeError: If the value is not JSON-serializable.
@@ -154,16 +148,12 @@ class Reference(object):
 
     def push(self, value=''):
         """Creates a new child node.
-
         The optional value argument can be used to provide an initial value for the child node. If
         no value is provided, child node will have empty string as the default value.
-
         Args:
           value: JSON-serializable initial value for the child node (optional).
-
         Returns:
           Reference: A Reference representing the newly created child node.
-
         Raises:
           ValueError: If the value is None.
           TypeError: If the value is not JSON-serializable.
@@ -177,10 +167,8 @@ class Reference(object):
 
     def update(self, value):
         """Updates the specified child keys of this Reference to the provided values.
-
         Args:
           value: A dictionary containing the child keys to update, and their new values.
-
         Raises:
           ValueError: If value is empty or not a dictionary.
           ApiCallError: If an error occurs while communicating with the remote database server.
@@ -191,26 +179,80 @@ class Reference(object):
             raise ValueError('Dictionary must not contain None keys or values.')
         self._client.request_oneway('patch', self._add_suffix(), json=value, params='print=silent')
 
+    def update_with_etag(self, value, etag):
+        """Updates the specified child keys of this Reference to the provided values and uses ETag to make sure data is up to date.
+        Args:
+            value: A dictionary containing the child keys to update, and their new values.
+            etag: ETag value for the Reference.
+        Returns:
+            ValueError: If value is empty or not a dictionary, or if etag is not a string.
+        """
+        if not value or not isinstance(value, dict):
+            raise ValueError('Value argument must be a non-empty dictionary.')
+        if None in value.keys() or None in value.values():
+            raise ValueError('Dictionary must not contain None keys or values.')
+        if not isinstance(etag, str):
+            raise ValueError('ETag must be a string.')
+
+        try:
+            self._client.request_oneway('put', self._add_suffix(), json=value, headers={'if-match': etag})
+        except ApiCallError as error:
+            detail = error.detail
+            snapshot = detail.response.json()
+            etag = detail.response.headers['ETag']
+            return etag, snapshot
+
     def delete(self):
         """Deleted this node from the database.
-
         Raises:
           ApiCallError: If an error occurs while communicating with the remote database server.
         """
         self._client.request_oneway('delete', self._add_suffix())
 
+    def transaction(self, transaction_update, on_complete=None):
+        """Write to database using a transaction.
+        Args:
+            transaction_update: function that takes in current database data as a parameter.
+            on_complete: function that takes takes in the following parameters:
+                error: Error message, possibly null
+                committed: Whether the transaction_update function committed data to the database
+                data: The data currently in the database
+
+        """
+        if not callable(transaction_update):
+            raise ValueError('transaction_update must be a function.')
+        if on_complete is not None and not callable(on_complete):
+            raise ValueError('on_complete must be a function.')
+
+        error = None
+        committed = False
+        try:
+            tries = 0
+            etag, data = self.get_with_etag()
+            val = transaction_update(data)
+            while tries < _FIREBASE_MAX_RETRIES:
+                resp = self.update_with_etag(val, etag)
+                if resp is None:
+                    committed = True
+                    data = val
+                    break
+                else:
+                    etag, data = resp
+                    tries += 1
+        except Exception as e:
+            error = e
+
+        if on_complete:
+            on_complete(error, committed, data)
+
     def order_by_child(self, path):
         """Returns a Query that orders data by child values.
-
         Returned Query can be used to set additional parameters, and execute complex database
         queries (e.g. limit queries, range queries).
-
         Args:
           path: Path to a valid child of the current Reference.
-
         Returns:
           Query: A database Query instance.
-
         Raises:
           ValueError: If the child path is not a string, not well-formed or None.
         """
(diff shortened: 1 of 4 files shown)


hiranya911:
Thanks for providing this pull request. I took a quick look, but before I get into the details it would be great if you can address the following issues:

1. Please go through the above comment added by @googlebot, and do the needful to address the CLA concerns.
2. Please reformat/re-indent the two source files (`db.py` and `test_db.py`), so any lines you haven't touched remain unchanged, and they do not show up in the diffs. Right now it shows as if the contents of the entire files have been modified.

firebase_admin/db.py
```diff
@@ -37,629 +36,644 @@
...
+			raise ValueError('transaction_update must be a function.')
+		if on_complete is not None and not callable(on_complete):
+			raise ValueError('on_complete must be a function.')
+
+		error = None
+		committed = False
+		try:
+			tries = 0
+			etag, data = self.get_with_etag()
+			val = transaction_update(data)
```
caseycrogers: Hey Alexander, this looks awesome! The one thing I'd say is that you could change transaction_update to return a tuple of (val, retry) where retry is a boolean. Then, you can change you while loop to be:
  while retry == True:
This way the user passes in a transaction_update function that controls retry logic itself. We could also still include _FIREBASE_MAX_RETRIES so that users don't accidentally infinite loop, though I believe our SDKs that already have transactions do not do have this-so it may be best to be consistent.

caseycrogers: Also, transactions in Java and Javascript run asynchronously which is why the on_complete is there. As written, this implementation runs synchronously so the on_complete doesn't really have a use: the values could be directly returned instead of passed into on_complete.

We could either:
1) refactor this function such that it is asynchronous
2) remove the on_complete argument and directly return (error, committed, data)
I heavily lean towards 2 as asynchronous code and callbacks are much less idiomatic in Python than in Java or Javascript.

I'm going to try and grab some time to put a PR against your PR with these recommendations, though I can't guarantee I'll get the chance.

Thanks so much for putting time and effort in to working on this!

alexanderwhatley:
Hi guys, thanks for the feedback! I'll try to get another pull request revising these done by tonight. A few questions/concerns:

1. @hiranya911, I made all of the commits, as you can see from the names, but one was accidentally made with a git account that is not associated with Github or Gmail. Do you think you could manually override the Googlebot on the CLA request?
2. @caseycrogers, I think it's a good idea to have transaction_update return val and boolean. Also, the other SDKs actually do have a retry limit, which is also 25. See here for the Java one: https://github.com/firebase/firebase-admin-java/blob/f9b81d291cd854fab9159582c80f09dc5d561ac0/src/main/java/com/google/firebase/database/core/Repo.java
3. I agree that we can get rid of on_complete.

caseycrogers:
@alexanderwhatley Serves me right for not checking myself!
Strike what I said about removing the max_retries.

hiranya911:
@alexanderwhatley I think the easiest solution would be to get that account associated with your Github account. This can be done via your Github profile settings: https://help.github.com/articles/adding-an-email-address-to-your-github-account/

alexanderwhatley:
@caseycrogers, I got rid of the callback for the transaction function.

@hiranya911: It doesn't seem like the git account I used to commit at first has an associated email account, so I couldn't add it to my GitHub account. I think the only possibilities now are that either you or someone else manually override the googlebot, or you make a PR from this PR yourself.

tests/test_db.py
```diff
@@ -34,14 +37,24 @@ def __init__(self, data, status, recorder):
...
         self._recorder = recorder
+        self._etag = '0'
 
     def send(self, request, **kwargs):
+        if 'if-match' in request.headers and request.headers['if-match'] != self._etag:
+            response = Response()
+            response._content = request.body
+            response.headers = CaseInsensitiveDict({'ETag': self._etag})
+            request_exception = RequestException(response=response)
+            raise db.ApiCallError('', request_exception)
```
hiranya911: I think it's more accurate to raise the `RequestException` here, and let the `db` module handle it. Or better yet, return a response with status code 412, along with the correct headers and payload.

firebase_admin/db.py
```diff
@@ -191,26 +181,72 @@ def update(self, value):
...
+        etag, data = self.get_with_etag()
+        val = transaction_update(data)
+        while tries < _TRANSACTION_MAX_RETRIES:
+            resp = self.update_with_etag(val, etag)
+            if resp is None:
+                break
+            else:
+                etag, data = resp
+                val = transaction_update(data)
+                tries += 1
```
caseycrogers: We can restructure this while loop a little:

```
etag, data = self.get_with_etag()
while tries < _TRANSACTION_MAX_RETRIES:
    val, should_retry = transaction_update(data)
    if not should_retry:
        break
    resp = self.update_with_etag(val, etag)
    if resp is None:
        break
    else:
        etag, data = resp
        tries += 1
```
This also lets the user specify when to stop retrying in transaction_update. Note that this code will raise an exception if transaction_update does not return a tuple. We probably want to catch that exception and raise a one with a context specific message in its place.

Note that should_retry need not be a boolean. We could either assert that it is a boolean and raise a clear exception to prevent users from accidentally passing unintended values through as should_retry, or we could leave it as is, assuming that in some cases a user may be passing through things like an empty list versus a non-empty list as their retry criteria.

hiranya911:
Btw, I think your first commit is associated with the email address "awhatley-mbp@awhatley-mbp.ad.quora.com". Not sure if that's a real email address, or some automatically managed account. But see if you can get it added to your github profile.

hiranya911:
Related to #53

hiranya911:
Hi @alexanderwhatley. It seems you can amend the author of that commit via git rebase: https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

Give that a try, and see if you can change the author of the commit.

alexanderwhatley:
@hiranya911 and @caseycrogers, I did most of the stuff you guys requested. I also fixed the issue with the CLA request by following the StackOverflow link's directions. A few things:

1. For consistency, I think it makes more sense for the `_update_with_etag` to return True, the current value of the etag, and the value of the just updated data if the request goes through. 

2. @caseycrogers, I don't think the way you want the while loop refactored is a good idea, especially given the refactored `_update_with_etag` function, which always returns a tuple of three items. In particular, I would think that in most cases if the `transaction_update` functions returns a boolean as a second argument, it would be false, and if it were false, the while loop would break before 'resp = self.update_with_etag(val, etag)' executes, and thus the transaction would not go through.

3. It's not clear to me at all how to run the integration tests, and I couldn't find any directions, so I didn't add any. It looks like I have to specify the certificate somehow using `--cert`, but none of the files in the tests/data folder worked....

hiranya911:
Thanks @alexanderwhatley. So glad the cla  issue got resolved. I'll take another look at your code tomorrow. I'm also working on getting this new feature/api approved internally. 

As for integration tests, there are some instructions at https://github.com/firebase/firebase-admin-python/blob/master/CONTRIBUTING.md#integration-testing

firebase_admin/db.py
```diff
@@ -191,6 +201,29 @@ def update(self, value):
...
         self._client.request_oneway('patch', self._add_suffix(), json=value, params='print=silent')
 
+    def _update_with_etag(self, value, etag):
+        """Sets the data at this location to the specified value, if the etag matches.
+        """
+        if not value or not isinstance(value, dict):
+            raise ValueError('Value argument must be a non-empty dictionary.')
+        if None in value.keys() or None in value.values():
+            raise ValueError('Dictionary must not contain None keys or values.')
+        if not isinstance(etag, str):
```
hiranya911: Use`isinstance(etag, six.string_types)` here. Behavior of `str` is vastly different between Python 2 and 3.

firebase_admin/db.py
```diff
@@ -191,6 +201,29 @@ def update(self, value):
...
+
+        success = True
+        snapshot = value
+        try:
+            self._client.request_oneway('put', self._add_suffix(), json=value,
+                                        headers={'if-match': etag})
+        except ApiCallError as error:
+            detail = error.detail
+            success = False
+            etag = detail.response.headers['ETag']
```
hiranya911: I think, to be in the safe side:

```
if detail.response and 'ETag' in detail.response.headers:
  etag = detail.response.headers['ETag']
  snapshot = detail.response.json()
  return False, etag, snapshot
else:
  # Possible connection error or other unexpected backend error.
  # Simply raise the original error.
  raise error
```

firebase_admin/db.py
```diff
@@ -199,6 +232,30 @@ def delete(self):
         """
         self._client.request_oneway('delete', self._add_suffix())
 
+    def transaction(self, transaction_update):
```
hiranya911: Lets return a boolean from this method to indicate success/failure. That way the caller has a deterministic way of knowing whether the transaction was committed or not:

```
if ref.transaction(update_func):
  do_more_stuff()
else:
  raise DatabaseError('Failed to commit txn')
```

tests/test_db.py
```diff
@@ -34,14 +36,23 @@ def __init__(self, data, status, recorder):
         self._data = data
         self._status = status
         self._recorder = recorder
+        self._etag = '0'
 
     def send(self, request, **kwargs):
+        if request.headers.get('if-match') is not None and \
```
hiranya911: Suggest you break this into two lines:

```
if_match = request.headers.get('if-match')
if if_match and if_match != self._etag:
   ...
```

alexanderwhatley:
@hiranya911, I just pushed some integration tests and fixed the other things. I feel pretty strongly that the API should expose the functionality for accessing the ETag, in case the user wishes to do some low level operations, such as accessing data and writing data to multiple parts of the database. I additionally think that we should add two functions, `get_etag()` and `check_etag(etag)` that retrieve the ETag without the corresponding data, and check to see if the ETag is correct, respectively. What do you think?

hiranya911:
Hi @alexanderwhatley. Given that the etag functionality is not available in our other SDKs, I think it's a good idea to exercise some restraint when adding it to our public API. At very least I'd like to hear from our Realtime Database team, and get their opinion on the matter. Therefore I'd like to simply release the transaction support you've implemented in its present state, and continue the discussion about exposing more etag-related functionality from the SDK. I suggest you open a new issue for that, documenting some of the use cases the proposed methods can be used for.  In the meantime, developers who really want to access etag values can call the private `_get_with_etag()` and `_update_with_etag()`  methods. 

If we choose to make this functionality public, I think we can probably expose them as extensions to our existing get and set methods:

```
get(self, etag=False) # When etag is True, behaves like _get_with_etag()
set(self, value, etag=None) # When etag is not None, behaves like _update_with_etag()
```

integration/test_db.py
```diff
@@ -149,6 +149,30 @@ def test_update_nested_children(self, testref):
...
+        edward = python.child('users').push(push_data)
+        etag, data = edward._get_with_etag()
+        assert data == push_data
+
+        update_data = {'name' : 'Jack Horner', 'since' : 1940}
+        failed_update = edward._update_with_etag(update_data, '')
+        assert failed_update == (False, etag, push_data)
+
+        successful_update = edward._update_with_etag(update_data, etag)
+        assert successful_update[0]
```
alexanderwhatley: Remember, the method returns three values, one of which is the ETag, and the ETag is not known ahead of time

alexanderwhatley:
Cool. I made the changes that you requested. I'll open up an issue and discuss the possibility of allowing access to the ETag object itself.

hiranya911:
Haven't forgotten this. Just waiting for the API review process to end. Should happen later this week.

firebase_admin/db.py
```diff
@@ -199,6 +235,35 @@ def delete(self):
...
+        Raises:
+            ValueError: If transaction_update is not a function.
+
+        """
+        if not callable(transaction_update):
+            raise ValueError('transaction_update must be a function.')
+
+        tries = 0
+        etag, data = self._get_with_etag()
+        val = transaction_update(data)
```
caseycrogers: Can we rename 'val' to something like 'new_snapshot' or 'snapshot_to_write'?

'val' doesn't seem to communicate that this is the snapshot after applying transaction_update.

hiranya911: We should not move `_get_with_etag()` into the loop as it makes a REST call. Perhaps:

```
etag, data = self._get_with_etag()
while tries < _TRANSACTION_MAX_ENTRIES:
    new_data = transaction_update(data)
    success, etag, data = self._update_with_etag(new_data, etag)
    if success:
        return True
    tries += 1
```

firebase_admin/db.py
```diff
@@ -199,6 +235,35 @@ def delete(self):
...
+        while tries < _TRANSACTION_MAX_RETRIES:
+            success, etag, snapshot = self._update_with_etag(val, etag)
```
hiranya911: Good point. It seems Node and Java SDKs support a similar feature. In Node the update function can return `undefined` to abort, and in Java there's a `Transaction.abort()` method. In any case lets hold off on implementing this change. I'll run this by the API review team, and update this thread with the outcome.

firebase_admin/db.py
caseycrogers: Ah sorry-that was a mistake. I had intended for get_with_etag to remain outside of the loop.

hiranya911:
Changes proposed during the final review are at #56. I will merge this PR as it is, and implement the necessary changes in a new PR.

hiranya911:
Thanks @alexanderwhatley 

Hope you continue to contribute to this project.

alexanderwhatley:
@hiranya911, thanks, I will definitely contribute more in the future, this library is very useful for the work I am doing. Do you have an ETA when the transaction features will be released on PyPI?

hiranya911:
Hopefully next week. I'm working on getting the docs in order.

hiranya911:
This was released today: https://firebase.google.com/support/release-notes/admin/python#2.2.0

Issue #56, opened 2017-08-11
Support for Aborting Transactions


hiranya911:
#54 implements transaction support where the `ref.transaction()` method accepts an `update_function()`. Following changes are proposed to this API:

* `update_function()` may signal abort by raising an exception. This is probably how this API works today. Verify and fix if needed.
* `transaction()` method returns the updated value when it commits. If retry attempts are exhausted it should raise an error.

hiranya911:
These changes are now implemented.

PR #57, opened 2017-08-11
Improvements to the Transaction API

hiranya911:
Follow up to #54. Implements #56.

* Returning the updated value from `transaction()` method.
* Updated API docs.

### DIFF
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -208,21 +208,18 @@ class Reference(object):
         if not isinstance(etag, six.string_types):
             raise ValueError('ETag must be a string.')
 
-        success = True
-        snapshot = value
         try:
             self._client.request_oneway(
                 'put', self._add_suffix(), json=value, headers={'if-match': etag})
+            return True, etag, value
         except ApiCallError as error:
             detail = error.detail
-            if detail.response.headers and 'ETag' in detail.response.headers:
+            if detail.response is not None and 'ETag' in detail.response.headers:
                 etag = detail.response.headers['ETag']
                 snapshot = detail.response.json()
                 return False, etag, snapshot
             else:
                 raise error
-
-        return success, etag, snapshot
 
     def delete(self):
         """Deletes this node from the database.
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -137,11 +137,9 @@ class Reference(object):
         return self._client.request('get', self._add_suffix())
 
     def _get_with_etag(self):
-        """Returns the value at the current location of the database, along with its ETag.
-        """
-        data, headers = self._client.request('get', self._add_suffix(),
-                                             headers={'X-Firebase-ETag' : 'true'},
-                                             resp_headers=True)
+        """Returns the value at the current location of the database, along with its ETag."""
+        data, headers = self._client.request(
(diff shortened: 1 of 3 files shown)


firebase_admin/db.py
```diff
@@ -228,23 +225,40 @@ def _update_with_etag(self, value, etag):
...
+        value of this reference into a new value. If another client writes to this location before
+        the new value is successfully saved, the update function is called again with the new
+        current value, and the write will be retried. In case of repeated failures, this method
+        will retry the transaction up to 25 times before giving up and raising a TransactionError.
+        The update function may also force an early abort by raising an exception instead of
+        returning a value.
 
         Args:
-            transaction_update: function that takes in current database data as a parameter.
+            transaction_update: A function which will be passed tbe current data stored at this
```
rockwotj: `s/tbe/the`

hiranya911: Fixed.

firebase_admin/db.py
```diff
@@ -228,23 +225,40 @@ def _update_with_etag(self, value, etag):
...
+
+        Unlike a normal `set()`, which just overwrites the data regardless of its previous state,
+        `transaction()` is used to modify the existing value to a new value, ensuring there are
+        no conflicts with other clients simultaneously writing to the same location.
+
+        This is accomplished by passing an update function which is used to transform the current
+        value of this reference into a new value. If another client writes to this location before
+        the new value is successfully saved, the update function is called again with the new
+        current value, and the write will be retried. In case of repeated failures, this method
+        will retry the transaction up to 25 times before giving up and raising a TransactionError.
```
bjornick: Do we want to make this number configurable?

firebase_admin/db.py
```diff
@@ -253,16 +267,13 @@ def transaction(self, transaction_update):
 
         tries = 0
         etag, data = self._get_with_etag()
-        val = transaction_update(data)
         while tries < _TRANSACTION_MAX_RETRIES:
-            success, etag, snapshot = self._update_with_etag(val, etag)
+            new_data = transaction_update(data)
+            success, etag, data = self._update_with_etag(new_data, etag)
```
bjornick: Are we using the etag?

firebase_admin/db.py
hiranya911: I think that's reasonable. I've created #58 to implement it. Since this requires an API review, I'll defer it to a future development cycle.

firebase_admin/db.py
hiranya911: Yes, in the next iteration of the loop `etag` gets passed to the `_update_with_etag()` call.

firebase_admin/db.py
```diff
@@ -253,16 +264,13 @@ def transaction(self, transaction_update):
...
             if success:
-                return True
-            else:
-                val = transaction_update(snapshot)
-                tries += 1
-
-        return False
+                return new_data
+            tries += 1
+        raise TransactionError('Transaction aborted after failed retries.')
```
caseycrogers: Should we add the number of retries attempted to this error?

hiranya911: Lets report that at #58 to be fixed in the future. We are trying to get a release build out at the moment.

Issue #55, opened 2017-08-07
API for accessing ETag values


alexanderwhatley:
As discussed in https://github.com/firebase/firebase-admin-python/pull/54, I think it would be a good idea of exposing API functions that can explicitly access the value of the ETag of the firebase db. 

We could have the following API for the Reference class:
```
get(self, etag=False) # When etag is True, behaves like _get_with_etag()
set(self, value, etag=None) # When etag is not None, behaves like _update_with_etag()
get_etag(self) # Retrieves the etag without the corresponding data
check_etag(self, etag) # Returns a boolean indicating whether or not the etag is correct
```

These API functions could come in handy for low level operations that the user may want to perform that the proposed transaction functionality may not be able to handle. An example would be if the user keeps a cache of some of the firebase data, the user will likely want to keep track of the ETag for each part of the database, so he can check whether the data in some part of the cache is up to date.

alexanderwhatley:
@caseycrogers, you said in the comments for https://github.com/firebase/firebase-admin-python/pull/54 that 

> As for your interest in adding low-level access to ETags-this seems reasonable to me. However, there is a caveat:
> 
> As-is, the ETag of a given value at one path in the database will be the same as the ETag of the same value at any other path.
> 
> Additionally, the ETag of values will not change with time. However, we do not document these two properties and do not guarantee that EITHER of them will remain true in the future. As such, we strongly discourage developers from fetching ETag values and then saving them to be used with later transactions at the same or different locations.

I don't think that this is a problem in the case of caching. If the data is the same, and hence the ETag is the same, then the cache is not stale. Also, it does not matter here if the ETag does not change as a function of time (I would certainly think that we would not want it to!)

caseycrogers:
By change as a function time I mean that in the future we may choose to change our algorithm for producing ETags. We have no such plans at the moment but want to keep the option open for ourselves.

For caching it should be fine. Another note that you might find very helpful-while implementing the conditional requests logic I also implemented full support for the [if-none-match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) header. However, since this feature was well out of scope of the project (allowing transactions in REST) we decided to keep the functionality but not document it. If you're doing any caching related stuff I highly recommend you use if-none-match.

It may be even worthwhile to add caching functionality with if-none-match to this library's default GET implementation. Preferably with a flag to turn it on or off.

alexanderwhatley:
[...]

PR #59, opened 2017-08-16
Low Level ETag API

alexanderwhatley:
Implemented low level ETag API discussed in https://github.com/firebase/firebase-admin-python/issues/55. I decided not to implement `check_etag`, as it's easiest just to call `get_etag` and then check. I also renamed `_update_with_etag` to `_set_with_etag` as the name makes more sense.

### DIFF
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -125,27 +125,16 @@ class Reference(object):
         full_path = self._pathurl + '/' + path
         return Reference(client=self._client, path=full_path)
 
-    def _get_with_etag(self):
-        """Returns the value at the current location of the database, along with its ETag.
-        """
-        data, headers = self._client.request('get', self._add_suffix(),
-                                             headers={'X-Firebase-ETag' : 'true'},
-                                             resp_headers=True)
-        etag = headers.get('ETag')
-        return etag, data
-
-    def get_etag(self):
+    def etag(self):
         """Returns the ETag at the current location of the database.
-
         Returns:
-            object: ETag of the current database Reference.
+            str: ETag of the current database Reference.
         """
         headers = self._client.request('get', self._add_suffix(),
                                        headers={'X-Firebase-ETag' : 'true'},
                                        resp_headers=True,
                                        params='print=silent')
-        etag = headers.get('ETag')
-        return etag
+        return headers.get('ETag')
 
     def get(self, etag=False):
         """Returns the value, and possibly the ETag, at the current location of the database.
@@ -158,35 +147,13 @@ class Reference(object):
           ApiCallError: If an error occurs while communicating with the remote database server.
         """
         if etag:
-            return self._get_with_etag()
+            data, headers = self._client.request('get', self._add_suffix(),
+                                                 headers={'X-Firebase-ETag' : 'true'},
+                                                 resp_headers=True)
+            etag = headers.get('ETag')
+            return data, etag
         else:
             return self._client.request('get', self._add_suffix())
-
-    def _set_with_etag(self, value, etag):
-        """Sets the data at this location to the specified value, if the etag matches.
-        """
-        if not value or not isinstance(value, dict):
-            raise ValueError('Value argument must be a non-empty dictionary.')
-        if None in value.keys() or None in value.values():
-            raise ValueError('Dictionary must not contain None keys or values.')
-        if not isinstance(etag, six.string_types):
-            raise ValueError('ETag must be a string.')
-
-        success = True
-        snapshot = value
-        try:
-            self._client.request_oneway('put', self._add_suffix(), json=value,
-                                        headers={'if-match': etag})
-        except ApiCallError as error:
-            detail = error.detail
-            if detail.response.headers and 'ETag' in detail.response.headers:
-                etag = detail.response.headers['ETag']
-                snapshot = detail.response.json()
-                return False, etag, snapshot
-            else:
-                raise error
-
-        return success, etag, snapshot
 
     def set(self, value, etag=None):
         """Sets the data at this location to the given value.
@@ -197,8 +164,12 @@ class Reference(object):
           value: JSON-serialable value to be set at this location.
           etag: Value of ETag that we want to check.
 
+        Returns:
+          object: Tuple of False, current location's etag, and snapshot of location's data
+                  if passed in etag does not match. 
+
         Raises:
-          ValueError: If the value is None.
+          ValueError: If the value is None, or if etag is not a string.
           TypeError: If the value is not JSON-serializable.
           ApiCallError: If an error occurs while communicating with the remote database server,
                         or if the ETag does not match.
@@ -206,7 +177,20 @@ class Reference(object):
         if value is None:
             raise ValueError('Value must not be None.')
         if etag is not None:
-            self._set_with_etag(value, etag)
+            if not isinstance(etag, six.string_types):
+                raise ValueError('ETag must be a string.')
+            try:
+                self._client.request_oneway(
+                    'put', self._add_suffix(), json=value, headers={'if-match': etag})
+                return True, etag, value
+            except ApiCallError as error:
+                detail = error.detail
+                if detail.response is not None and 'ETag' in detail.response.headers:
+                    etag = detail.response.headers['ETag']
+                    snapshot = detail.response.json()
+                    return False, etag, snapshot
+                else:
+                    raise error
         else:
             self._client.request_oneway('put', self._add_suffix(), json=value,
                                         params='print=silent')
@@ -252,7 +236,7 @@ class Reference(object):
                                     params='print=silent')
 
     def delete(self):
-        """Deleted this node from the database.
+        """Deletes this node from the database.
 
         Raises:
           ApiCallError: If an error occurs while communicating with the remote database server.
@@ -261,6 +245,7 @@ class Reference(object):
 
     def transaction(self, transaction_update):
         """Atomically modifies the data at this location.
+
         Unlike a normal `set()`, which just overwrites the data regardless of its previous state,
         `transaction()` is used to modify the existing value to a new value, ensuring there are
         no conflicts with other clients simultaneously writing to the same location.
@@ -287,10 +272,10 @@ class Reference(object):
             raise ValueError('transaction_update must be a function.')
 
         tries = 0
-        etag, data = self._get_with_etag()
+        data, etag = self.get(etag=True)
         while tries < _TRANSACTION_MAX_RETRIES:
             new_data = transaction_update(data)
-            success, etag, data = self._set_with_etag(new_data, etag)
+            success, etag, data = self.set(new_data, etag=etag)
             if success:
                 return new_data
             tries += 1
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 """Firebase Realtime Database module.
-
 This module contains functions and classes that facilitate interacting with the Firebase Realtime
 Database. It supports basic data manipulation operations, as well as complex queries such as
 limit queries and range queries. However, it does not support realtime update notifications. This
@@ -42,16 +41,12 @@ _USER_AGENT = 'Firebase/HTTP/{0}/{1}.{2}/AdminPython'.format(
 
 def reference(path='/', app=None):
     """Returns a database Reference representing the node at the specified path.
-
     If no path is specified, this function returns a Reference that represents the database root.
-
     Args:
       path: Path to a node in the Firebase realtime database (optional).
       app: An App instance (optional).
-
     Returns:
       Reference: A newly initialized Reference.
-
     Raises:
       ValueError: If the specified path or app is invalid.
     """
@@ -73,7 +68,6 @@ class Reference(object):
 
     def __init__(self, **kwargs):
         """Creates a new Reference using the provided parameters.
-
         This method is for internal use only. Use db.reference() to obtain an instance of
         Reference.
         """
@@ -102,16 +96,12 @@ class Reference(object):
 
     def child(self, path):
         """Returns a Reference to the specified child node.
-
         The path may point to an immediate child of the current Reference, or a deeply nested
         child. Child paths must not begin with '/'.
-
         Args:
           path: Path to the child node.
-
         Returns:
           Reference: A database Reference representing the specified child node.
-
         Raises:
           ValueError: If the child path is not a string, not well-formed or begins with '/'.
         """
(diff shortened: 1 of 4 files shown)


firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
         full_path = self._pathurl + '/' + path
         return Reference(client=self._client, path=full_path)
 
-    def get(self):
-        """Returns the value at the current location of the database.
+    def _get_with_etag(self):
```
hiranya911: How about we just move this logic straight into the `get()` method, and get rid of this helper? `transaction()` can call `self.get(etag=True)`.

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
+        """Returns the value at the current location of the database, along with its ETag.
+        """
+        data, headers = self._client.request('get', self._add_suffix(),
+                                             headers={'X-Firebase-ETag' : 'true'},
+                                             resp_headers=True)
+        etag = headers.get('ETag')
+        return etag, data
+
+    def get_etag(self):
```
hiranya911: Perhaps call it `etag()`? That sounds more consistent with the other methods we have in this module like `child()` and `reference()`.

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
 
         Returns:
-          object: Decoded JSON value of the current database Reference.
+            object: ETag of the current database Reference.
+        """
+        headers = self._client.request('get', self._add_suffix(),
+                                       headers={'X-Firebase-ETag' : 'true'},
+                                       resp_headers=True,
+                                       params='print=silent')
+        etag = headers.get('ETag')
```
hiranya911: `return headers.get('ETag')`

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
-    def set(self, value):
-        """Sets the data at this location to the given value.
+        success = True
+        snapshot = value
+        try:
+            self._client.request_oneway('put', self._add_suffix(), json=value,
+                                        headers={'if-match': etag})
+        except ApiCallError as error:
+            detail = error.detail
+            if detail.response.headers and 'ETag' in detail.response.headers:
```
hiranya911: Check for `detail.response is not None` here (check the code in master now).

firebase_admin/db.py
```diff
@@ -197,32 +248,11 @@ def update(self, value):
...
-                snapshot = detail.response.json()
-                return False, etag, snapshot
-            else:
-                raise error
+        self._client.request_oneway('patch', self._add_suffix(), json=value,
+                                    params='print=silent')
 
     def delete(self):
-        """Deletes this node from the database.
+        """Deleted this node from the database.
```
hiranya911: Revert.

firebase_admin/db.py
```diff
@@ -231,33 +261,27 @@ def delete(self):
 
     def transaction(self, transaction_update):
         """Atomically modifies the data at this location.
-
```
hiranya911: Add the empty lines back.

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
-    def get(self):
-        """Returns the value at the current location of the database.
+    def _get_with_etag(self):
+        """Returns the value at the current location of the database, along with its ETag.
+        """
+        data, headers = self._client.request('get', self._add_suffix(),
+                                             headers={'X-Firebase-ETag' : 'true'},
+                                             resp_headers=True)
+        etag = headers.get('ETag')
+        return etag, data
```
hiranya911: Recommend changing the order here to `return data, etag`

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
+
+    def get_etag(self):
+        """Returns the ETag at the current location of the database.
 
         Returns:
-          object: Decoded JSON value of the current database Reference.
+            object: ETag of the current database Reference.
```
hiranya911: `str: ETag of the....`

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
-    def _get_with_etag(self):
-        """Returns the value at the current location of the database, along with its ETag."""
-        data, headers = self._client.request(
-            'get', self._add_suffix(), headers={'X-Firebase-ETag' : 'true'}, resp_headers=True)
-        etag = headers.get('ETag')
-        return etag, data
+    def _set_with_etag(self, value, etag):
+        """Sets the data at this location to the specified value, if the etag matches.
+        """
+        if not value or not isinstance(value, dict):
```
hiranya911: The value doesn't have to be a `dict` right? I think any JSON-serializable value should work here.

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
+            raise ValueError('Dictionary must not contain None keys or values.')
+        if not isinstance(etag, six.string_types):
+            raise ValueError('ETag must be a string.')
 
-    def set(self, value):
-        """Sets the data at this location to the given value.
+        success = True
+        snapshot = value
+        try:
+            self._client.request_oneway('put', self._add_suffix(), json=value,
```
hiranya911: I had done a bit of refactoring here. Make sure some of those changes remain.

firebase_admin/db.py
```diff
@@ -125,40 +125,91 @@ def child(self, path):
...
+        else:
+            return self._client.request('get', self._add_suffix())
 
-    def _get_with_etag(self):
-        """Returns the value at the current location of the database, along with its ETag."""
-        data, headers = self._client.request(
-            'get', self._add_suffix(), headers={'X-Firebase-ETag' : 'true'}, resp_headers=True)
-        etag = headers.get('ETag')
-        return etag, data
+    def _set_with_etag(self, value, etag):
```
hiranya911: I wonder if this should also be made part of the `set()` method. That would allow callers of `set(value, etag=True)` to access the return values.

hiranya911:
Looks pretty good overall. Just a few things to improve, and a couple of questions.

alexanderwhatley:
Ok pushed some changes. There are a couple of lint errors in db.py that I'm not sure how to fix, perhaps you can weigh in? Thanks.

firebase_admin/db.py
```diff
@@ -125,40 +125,75 @@ def child(self, path):
...
         The value must be JSON-serializable and not None.
+        If etag != None, checks to make sure that the ETag value matches that of the location.
 
         Args:
           value: JSON-serialable value to be set at this location.
+          etag: Value of ETag that we want to check.
+
+        Returns:
+          object: Tuple of False, current location's etag, and snapshot of location's data
+                  if passed in etag does not match.
```
hiranya911: Indent by 4 spaces. Lint error is possibly due to a trailing whitespace at the end of the line.

firebase_admin/db.py
```diff
@@ -125,40 +125,75 @@ def child(self, path):
...
+          object: Tuple of False, current location's etag, and snapshot of location's data
+                  if passed in etag does not match. 
 
         Raises:
-          ValueError: If the value is None.
+          ValueError: If the value is None, or if etag is not a string.
           TypeError: If the value is not JSON-serializable.
-          ApiCallError: If an error occurs while communicating with the remote database server.
+          ApiCallError: If an error occurs while communicating with the remote database server,
+                        or if the ETag does not match.
```
hiranya911: Indent by 4 spaces. Might want to add a lint ignore directive here to fix the 2nd lint error.

`# pylint: disable=missing-raises-doc`

Might have to experiment a bit to get the placement of the directive right.

firebase_admin/db.py
```diff
@@ -235,38 +249,33 @@ def transaction(self, transaction_update):
         Unlike a normal `set()`, which just overwrites the data regardless of its previous state,
         `transaction()` is used to modify the existing value to a new value, ensuring there are
         no conflicts with other clients simultaneously writing to the same location.
-
```
hiranya911: Please add these lines back. Here and below.

firebase_admin/db.py
```diff
@@ -125,40 +125,75 @@ def child(self, path):
         full_path = self._pathurl + '/' + path
         return Reference(client=self._client, path=full_path)
 
-    def get(self):
-        """Returns the value at the current location of the database.
+    def etag(self):
+        """Returns the ETag at the current location of the database.
+        Returns:
```
hiranya911: Add an empty line before `Returns:`.

firebase_admin/db.py
```diff
@@ -125,40 +125,75 @@ def child(self, path):
...
+                                       params='print=silent')
+        return headers.get('ETag')
+
+    def get(self, etag=False):
+        """Returns the value, and possibly the ETag, at the current location of the database.
 
         Returns:
-          object: Decoded JSON value of the current database Reference.
+          object: Decoded JSON value of the current database Reference if etag=False, otherwise
+                  the decoded JSON value and the corresponding ETag.
```
hiranya911: Indent by 4 spaces.

firebase_admin/db.py
```diff
@@ -678,8 +687,11 @@ def from_app(cls, app):
 
     def request(self, method, urlpath, **kwargs):
         resp_headers = kwargs.pop('resp_headers', False)
+        params = kwargs.get('params', None)
         resp = self._do_request(method, urlpath, **kwargs)
-        if resp_headers:
+        if resp_headers and params == 'print=silent':
```
hiranya911: Can we just return `resp.json(), resp.headers` in this case? Does that trigger an error? If not I'd say simply return the tuple, and ignore the payload value (which I assume would get set to None).

alexanderwhatley: No, if print=silent, resp.json() throws an error message.

alexanderwhatley:
Made some more changes, please take a look.

firebase_admin/db.py
```diff
@@ -125,40 +125,77 @@ def child(self, path):
...
+        Returns:
+            str: ETag of the current database Reference.
+        """
+        headers = self._client.request('get', self._add_suffix(),
+                                       headers={'X-Firebase-ETag' : 'true'},
+                                       resp_headers=True,
+                                       params='print=silent')
+        return headers.get('ETag')
+
+    def get(self, etag=False):
```
hiranya911: Document the etag parameter.

```
Args:
  etag: A boolean indicating whether to return the ETag of the current location along with the value (optional).
```

firebase_admin/db.py
```diff
@@ -125,40 +125,77 @@ def child(self, path):
...
+                                                 resp_headers=True)
+            etag = headers.get('ETag')
+            return data, etag
+        else:
+            return self._client.request('get', self._add_suffix())
 
-    def set(self, value):
+    def set(self, value, etag=None):
         """Sets the data at this location to the given value.
-
```
hiranya911: Add this line back.

firebase_admin/db.py
```diff
@@ -125,40 +125,77 @@ def child(self, path):
...
         The value must be JSON-serializable and not None.
+        If etag != None, checks to make sure that the ETag value matches that of the location.
 
         Args:
           value: JSON-serialable value to be set at this location.
+          etag: Value of ETag that we want to check.
```
hiranya911: Add `(optional).` at the end.

firebase_admin/db.py
hiranya911: Ok. Lets leave it as it is for now. I'll try to think of a better way to refactor this bit out.

hiranya911:
After thinking about this API a bit more deeply, I think I've arrived at the following API:

```
ref.get(etag=False)
ref.etag()
ref.compare_and_set(expected_etag, update_value)
```

I didn't much like the fact that `set()` method not returning anything in one case, and returning a 3-tuple in the other. So I left it unchanged, and introduced a new `compare_and_set()` method. This should make the intent clear and improve readability of user code as well. 

I will use this as a starting point for our internal reviews. But lets not make any code changes just yet. Lets wait for the review process to come to an end. In either case I believe this PR has all the bits of logic necessary to support the functionality we wish to expose. So it should be easy to refactor the public API in anyway we want.

hiranya911:
Hi @alexanderwhatley. Sorry about the delay, but I finally have an update to share with you. After our internal review process, we have reached consensus on the following API:

```
ref.get(etag=False)
ref.update_if_unchanged(expected_etag, value) => (success, value, etag)
ref.get_if_changed(etag) => (success, value, etag)
```
The argument was `get_if_changed()` is more useful in practice than a method that only returns the etag string. For example, in your caching scenario:

```
# Initial read to populate cache
value, etag = ref.get(etag=True)
update_cache(ref.key, value, etag)
```
And after some time...
```
success, value, etag = ref.get_if_changed(etag_from_cache)
if success:
  update_cache(ref.key, value, etag)
```

I believe you already have code necessary for `get()` and `update_if_unchanged()`. To implement `get_if_changed()` you need to use the [if-none-match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) header as mentioned by @caseycrogers in #55. Are you in a position to make these changes in the PR? Otherwise I can try to make the necessary changes on top of what you have implemented so far.

alexanderwhatley:
Sure, I'll try to get something out by tonight.

alexanderwhatley:
@hiranya911, I pushed some changes for the new API. It's not clear to me how to handle the case where the ETag matches in the `get_if_changed` method, however. If the method fails, then it raises a 304 status, which is not an error status, and so it is not caught.
Furthermore, it seems like the response's JSON in this case (accessed by `resp.json()`) is empty, and so it throws a JSONDecodeError when trying to unserialize it. How should I go about handling this case? Thanks.

hiranya911:
@alexanderwhatley I suppose, for now, you can directly call `client._do_request()` from your method, and test for the status code. You will have to add a pylint directive to ignore the warning generated for calling a hidden method. This is ugly, but it's ok for now. I've been separately working on refactoring all the HTTP client code in the SDK, and that will take care of it.

alexanderwhatley:
Hi @hiranya911, I pushed some more changes. It still seems like if the if-none-match header matches the ETag provided, then the server does not return any of the data stored in the current location.

hiranya911:
That is expected with a `304 Not Modified` response. Since the content at server hasn't changed, we can rely on the client-side cached values. You should just return `(False, None, None)` for this case.

firebase_admin/db.py
```diff
@@ -125,31 +125,70 @@ def child(self, path):
         full_path = self._pathurl + '/' + path
         return Reference(client=self._client, path=full_path)
 
-    def get(self):
-        """Returns the value at the current location of the database.
+    def etag(self):
```
hiranya911: Lets remove this method. Our API review team decided to implement `get_if_changed()` instead of `etag()`.

alexanderwhatley:
Done. I also made some changes to the unit tests.

hiranya911:
@alexanderwhatley Thank you for patiently seeing this through. Great to have you as a contributor.

hiranya911:
@alexanderwhatley I'm in the process of refactoring the DB module. However, I'm seeing a weird behavior with the new `get_if_changed()` method. I would appreciate if you can verify this for me independently.

Steps:
1. Checkout the branch https://github.com/firebase/firebase-admin-python/tree/hkj-db-refactor
2. Uncomment line 98 of `integration/test_db.py`
3. Run the integration test suite

At least in my test environment, after executing `test_get_if_changed()`, the test suite seems to hang. It looks like a `304` response from the server leaves the HTTP session in an inconsistent state. Can you or somebody verify this for me?

alexanderwhatley:
@hiranya911, hmm yeah, I'm getting it hanging too. It actually happened yesterday before I pushed the code as well, but I assumed it was due to the slow hotel wifi...

hiranya911:
Thanks for checking. I'm looking into it. This might be a server-side issue.

hiranya911:
This was partially [released today](https://firebase.google.com/support/release-notes/admin/python#2.3.0). We removed the `get_if_changed()` method from the public API temporarily due to a bug in the server-side API. We'll add it back once the server team has fixed the issue.

Issue #62, opened 2017-08-20
Transaction API does not match documentation


andyfangdz:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOS
  * Firebase SDK version: 2.2.0
  * Firebase Product: database (auth, database, storage, etc)

#### Steps to reproduce:

1. Create a database in firebase like 
```json
{
  "jobs": {
    "count": 0
  }
}
```
2. in Python, do `root.child('jobs/count').transaction(lambda x: x + 1 if x else 1)`

Raised error: `ValueError: Value argument must be a non-empty dictionary.`

#### Relevant Code:
https://github.com/firebase/firebase-admin-python/blob/99c48ef28692a263ef4171b1b0c15489584b9aab/firebase_admin/db.py#L204-L207
These lines of code assume that the value passed in is a dictionary, which was not mentioned in the documentation (https://firebase.google.com/docs/database/admin/save-data#section-transactions) at all.

PR #63, opened 2017-08-20
Remove unnecessary checks on update with etag

andyfangdz:
This PR fixes https://github.com/firebase/firebase-admin-python/issues/62 by removing unnecessary checks in function `_update_with_etag`.

## Why remove the checks instead of updating the docs?

1. To be more consistent with Java/Node API
2. It is specified in the docs that transaction is a `set()` operation, however, the current code checks the value as if it's an `update()` operation.

Code of `set()`, notice how it doesn't do this check at all and is using `put` like `transaction()`
https://github.com/firebase/firebase-admin-python/blob/99c48ef28692a263ef4171b1b0c15489584b9aab/firebase_admin/db.py#L146-L161

Code of `update()`, instead of using `put`, it's using `patch`. The original author might have borrowed the code from here:
https://github.com/firebase/firebase-admin-python/blob/99c48ef28692a263ef4171b1b0c15489584b9aab/firebase_admin/db.py#L186-L200

/cc @alexanderwhatley @hiranya911

### DIFF
--- a/firebase_admin/__init__.py
+++ b/firebase_admin/__init__.py
@@ -23,7 +23,7 @@ from firebase_admin import credentials
 
 # Declaring module version as per https://www.python.org/dev/peps/pep-0396/#specification
 # Update this accordingly for each release.
-__version__ = '2.2.0'
+__version__ = '2.2.1'
 
 _apps = {}
 _apps_lock = threading.RLock()
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -201,10 +201,6 @@ class Reference(object):
 
     def _update_with_etag(self, value, etag):
         """Sets the data at this location to the specified value, if the etag matches."""
-        if not value or not isinstance(value, dict):
-            raise ValueError('Value argument must be a non-empty dictionary.')
-        if None in value.keys() or None in value.values():
-            raise ValueError('Dictionary must not contain None keys or values.')
         if not isinstance(etag, six.string_types):
             raise ValueError('ETag must be a string.')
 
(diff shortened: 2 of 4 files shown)


andyfangdz:
I noticed that there's already a PR that changes the API in https://github.com/firebase/firebase-admin-python/pull/59. I'm not sure if that PR fixes this particular problem, but I think it makes sense to merge this PR if public API review takes more time because currently there's no way to update a single leaf element.

hiranya911:
Thanks @andyfangdz. This looks pretty good. Would you be able to provide an integration test for this as well? See `integration/test_db.py` and the [integration testing guide](https://github.com/firebase/firebase-admin-python/blob/master/CONTRIBUTING.md#integration-testing) for more information.

andyfangdz:
@hiranya911 Indeed! Here's the integration test. I also took the liberty of bumping a minor version since this change technically does not break anything already working with `2.2.0`.

hiranya911:
Thanks @andyfangdz for the contribution.

andyfangdz:
@hiranya911 Can you do a release for `2.2.1` on pypi? Thanks!

hiranya911:
@andyfangdz I was waiting to see if we can get #59 also approved, and do a 2.3.0 release sometime next week. I will do a release next week regardless of the status of #59. Sorry about the delay.

andyfangdz:
@hiranya911 No worries! Thanks for the fast review :smile:. Looking forward to #59!

hiranya911:
This was [released today](https://firebase.google.com/support/release-notes/admin/python#2.3.0).

andyfangdz:
Thanks!

PR #69, opened 2017-09-06
Implemented a util for getting project ID from the credential or the environment

hiranya911:
* Renamed the internal `utils` method as `_utils`
* Added a new `project_id()` function
* Updated test cases

### DIFF
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -14,9 +14,8 @@
 
 """Internal utilities common to all modules."""
 
-import os
-
 import firebase_admin
+
 
 def _get_initialized_app(app):
     if app is None:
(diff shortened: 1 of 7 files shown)


firebase_admin/_utils.py
```diff
@@ -32,3 +34,28 @@ def _get_initialized_app(app):
...
+    the App. Finally, it attempts to get the project ID by referencing the GCLOUD_PROJECT
+    environment variable.
+
+    Args:
+      app: A Firebase App instance.
+
+    Returns:
+      str: A project ID value or None.
+    """
+    pid = app.options.get('projectId')
```
bjornick: Wouldn't it make sense to initialize app to get all the values from the environment as opposed doing this lookup every time someone calls project_id(app)?

hiranya911: Done. Now this logic runs as part of `firebase_admin.initialize_app()`. Updated test cases accordingly.

hiranya911:
Made the suggested change. Project ID look up logic is now moved to `initialize_app()`.

Issue #71, opened 2017-09-06
FR: Provide a "timeout" option


mvalota:
I'm using firebase-admin v2.3.0 in a Google AppEngine app (Standard Environment, Python 2.7).
Unfortunately I'm getting a DeadlineExceededError once in a while when accessing my Firebase DB.

On Appengine the request timeout is 5 seconds, but it is possible to increase it up to 60 seconds, as described here:
https://cloud.google.com/appengine/articles/deadlineexceedederrors 

I already tried to follow the official solution proposed by Google for Requests, but it doesn't seem to work:
https://cloud.google.com/appengine/docs/standard/python/issue-requests

I've seen that the google-auth library provides a timeout option in her Requests transport:
https://google-auth.readthedocs.io/en/latest/reference/google.auth.transport.requests.html#module-google.auth.transport.requests

It would be great to have a "timeout" option on Firebase App initialization, in order to avoid the DeadlineExceededError on AppEngine.

have a nice day!
Mauro

hiranya911:
@mvalota Can you share the exact stacktrace/error message you're seeing? The 5 second timeout is for internal RPC calls made by App Engine. URL Fetch calls have a 60 second timeout by default:

```
Making requests to external URLs using URLFetch can also produce DeadlineExceededErrors
if the target website is having performance issues or normally takes more than 60 seconds
to reply. The logged stack trace of the DeadlineExceededErrors should contain calls to the
URLFetch libraries in these cases.
```

I would like to know what is actually triggering the deadline exceeded error.

mvalota:
Here's the stacktrace:

An error occurred during a request. (/base/data/home/apps/s~project-name/20170901t114805.403796533333348168/main.py:94)
Traceback (most recent call last):
  File "/base/data/home/apps/s~project-name/20170901t114805.403796533333348168/lib/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/base/data/home/apps/s~project-name/20170901t114805.403796533333348168/lib/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/base/data/home/apps/s~project-name/20170901t114805.403796533333348168/lib/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/base/data/home/apps/s~project-name/20170901t114805.403796533333348168/lib/flask/app.py", line 1612, in full_dispatch_request
[...]

PR #72, opened 2017-09-07
Implementing a httpTimeout option

hiranya911:
Related to #71.

A new 'httpTimeout' option is introduced which will be used to control the timeout of all outbound requests made by the `db` module. In the future, we can use the same in other modules too.

```
firebase_admin.initialize_app(credential, {
    'databaseURL' : test_url,
    'httpTimeout': 60 # timeout in seconds
})
```

### DIFF
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -686,7 +686,8 @@ class _Client(_http_client.JsonHttpClient):
           auth_override: A dictionary representing auth variable overrides or None (optional).
               Default value provides admin privileges. A None value here provides un-authenticated
               guest privileges.
-          timeout: HTTP request timeout in seconds (optional).
+          timeout: HTTP request timeout in seconds (optional). If not set connections will never
+              timeout, which is the default behavior of the underlying requests library.
         """
         _http_client.JsonHttpClient.__init__(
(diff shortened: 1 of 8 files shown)


firebase_admin/db.py
```diff
@@ -686,6 +686,7 @@ def __init__(self, credential, base_url, auth_override=_DEFAULT_AUTH_OVERRIDE):
           auth_override: A dictionary representing auth variable overrides or None (optional).
               Default value provides admin privileges. A None value here provides un-authenticated
               guest privileges.
+          timeout: HTTP request timeout in seconds (optional).
```
bjornick: What is the behavior if one isn't specified?

hiranya911: In that case we don't set any timeout on the connection, which activates the defaults of the `requests` library. However, their default is essentially "never timeout". Would it be better to default to something like 60 seconds instead?

bjornick: I think we should document the default behavior.  If the old behavior was "never timeout" we should probably leave it be, in case anybody depended on it.

hiranya911: Done

Issue #88, opened 2017-11-05
Unable to query by boolean False


arichen:
* Operating System version: OSX 10.11.4
  * Firebase SDK version: Firebase Admin Python SDK 2.4.0
  * Firebase Product: database

In firebase-admin-python/firebase_admin/db.py , it's unable to use equal_to(), start_at(), and end_at() with a boolean False value. It throws `ValueError: Equal to value must not be empty or None.`

Suggest the condition of False boolean value to be catered.

andremoniz:
Hi @hiranya911 do you know when this fix will be release?

hiranya911:
@andremoniz Planning a patch release later this week.

PR #89, opened 2017-11-05
Fix db.py cannot query by False boolean value

arichen:
Fixes issue https://github.com/firebase/firebase-admin-python/issues/88


hiranya911:
I think we can just simplify these if conditions to `if value is None:`. Empty string can be considered a valid parameter in some cases.

Btw, please complete the CLA if you wish to proceed with this PR.

hiranya911:
I'm closing this in favor of #96

PR #96, opened 2017-11-23
Fixing start_at, end_at, equal_to queries to accept False values

### LINKED ISSUE #88: Unable to query by boolean False

hiranya911:
`start_at()`, `end_at()` and `equal_to()` method in the `Query` class currently does not accept `False` as a valid argument (along with a bunch of other false values such as empty string and empty dict). This PR relaxes the argument checks so these values are supported.

Resolves #88. Related to #89.

### DIFF
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -442,7 +442,7 @@ class Query(object):
           Query: The updated Query instance.
 
         Raises:
-          ValueError: If the value is empty or None.
+          ValueError: If the value is None.
         """
         if start is None:
             raise ValueError('Start value must not be None.')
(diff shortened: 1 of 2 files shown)


firebase_admin/db.py
```diff
@@ -444,8 +444,8 @@ def start_at(self, start):
         Raises:
           ValueError: If the value is empty or None.
```
avishalom: If empty is ok, amend comments here and below

hiranya911: Done. Good catch.

PR #105, opened 2017-12-20
Auto init

avishalom:
Adding the auto init from env var to python.


firebase_admin/__init__.py
```diff
@@ -144,12 +146,29 @@ class _AppOptions(object):
     """A collection of configuration options for an App."""
 
     def __init__(self, options):
-        if options is None:
-            options = {}
-        if not isinstance(options, dict):
-            raise ValueError('Illegal Firebase app options type: {0}. Options '
-                             'must be a dictionary.'.format(type(options)))
-        self._options = options
+        if options is not None:
```
hiranya911: This looks a bit messy. Please refactor as follows:

```
if options is None:
    options = _load_from_environment()
...
```

tests/test_app.py
```diff
@@ -22,10 +22,25 @@
...
 
-
 CREDENTIAL = credentials.Certificate(
     testutils.resource_filename('service_account.json'))
 GCLOUD_PROJECT = 'GCLOUD_PROJECT'
+CONFIG_JSON = firebase_admin._CONFIG_JSON_ENV
+
+# This fixture will ignore the environment variable pointing to the default
+# configuration for the duration of the tests.
+@pytest.fixture(scope="session", autouse=True)
```
hiranya911: What are the tests affected by this fixture? Ideally it shouldn't affect tests in other modules.

tests/test_app.py
```diff
@@ -102,6 +264,10 @@ class TestFirebaseApp(object):
         firebase_admin.App('uninitialized', CREDENTIAL, {})
     ]
 
+    bad_config_file = ['firebase_config_empty.json',
```
hiranya911: Inline this list wherever you're using it. Seems to be used in only one test case.

tests/test_app.py
```diff
@@ -135,11 +301,29 @@ def test_app_init_with_invalid_options(self, options):
         with pytest.raises(ValueError):
             firebase_admin.initialize_app(CREDENTIAL, options=options)
 
+    @pytest.mark.parametrize('bad_file_name', bad_config_file)
+    def test_default_app_init_with_bad_config_from_env(self, bad_file_name):
```
hiranya911: `test_app_init_with_invalid_config_file`

tests/test_app.py
```diff
@@ -135,11 +301,29 @@ def test_app_init_with_invalid_options(self, options):
...
+        config_file_old = os.environ.get(CONFIG_JSON)
+        os.environ[CONFIG_JSON] = testutils.resource_filename(bad_file_name)
+        try:
+            with pytest.raises(ValueError):
+                firebase_admin.initialize_app(CREDENTIAL)
+        finally:
+            if config_file_old:
```
hiranya911: I think you should create a fixture, that sets a given value/file as the FIREBASE_CONFIG variable, and then reverts it back after the test. Then inject that fixture to any tests that is potentially affected by this change. You can probably get rid of that session scoped/autouse fixture that way.

tests/test_app.py
```diff
@@ -70,6 +85,154 @@ def get(self):
...
+        if self.config_file_old:
+            os.environ[CONFIG_JSON] = self.config_file_old
+        elif os.environ.get(CONFIG_JSON):
+            del os.environ[CONFIG_JSON]
+
+def named_option_pairs_for_test(named_id_option_pairs):
+    return dict(zip(("ids", "params"), zip(*named_id_option_pairs)))
+
+
+@pytest.fixture(**named_option_pairs_for_test([
```
hiranya911: This is a readability nightmare. Can we turn these into individual test cases? At very least, set this dictionary up as a global constant. Then use pytest indirect parameters to pass the test parameters from test case to the fixture: https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

tests/test_app.py
avishalom: yes you are right, this is no longer needed.

tests/test_app.py
avishalom: This is exactly what I am doing in the OptionsTest class

tests/test_app.py
```diff
@@ -140,6 +162,127 @@ def test_app_init_with_invalid_name(self, name):
...
+                                               'firebase_config_invalid.json',
+                                               'no_such_file'])
+    def test_app_init_with_invalid_config_file(self, bad_file_name):
+        config_old = set_config_env(bad_file_name)
+        with pytest.raises(ValueError):
+            firebase_admin.initialize_app(CREDENTIAL)
+        revert_config_env(config_old)
+
+
+    OptionsTestCase = namedtuple('OptionsTestCase',
```
hiranya911: This is sitting in the middle of a class definition. Please move to somewhere more appropriate.

tests/test_app.py
```diff
@@ -140,6 +162,127 @@ def test_app_init_with_invalid_name(self, name):
...
+                                      'storageBucket': 'sb1 - .mock'})]
+
+    @pytest.mark.parametrize('test_case', options_test_cases)
+    def test_app_init_with_default_config(self, test_case):
+        """Set the CONFIG env var and test that options are initialized"""
+        config_old = set_config_env(test_case.config_json)
+        app = firebase_admin.initialize_app(CREDENTIAL, options=test_case.init_options)
+        for field in firebase_admin._CONFIG_VALID_KEYS:
+            assert app.options.get(field) == test_case.want_options.get(field), test_case.name
+        revert_config_env(config_old)
```
hiranya911: This won't run if an assertion fails.  See if you can turn it into a self-cleaning fixture like https://github.com/firebase/firebase-admin-python/blob/master/tests/test_credentials.py#L83
  
If that's not possible, at least use a try-finally block.

tests/test_app.py
```diff
@@ -140,6 +162,127 @@ def test_app_init_with_invalid_name(self, name):
...
+        OptionsTestCase(name='full config file is ignored with missing values in options',
+                        config_json='firebase_config.json',
+                        init_options={'databaseAuthVariableOverride': 'davy1 - mock',
+                                      'projectId': 'pid1 - mock',
+                                      'storageBucket': 'sb1 - .mock'},
+                        want_options={'databaseAuthVariableOverride': 'davy1 - mock',
+                                      'projectId': 'pid1 - mock',
+                                      'storageBucket': 'sb1 - .mock'})]
+
+    @pytest.mark.parametrize('test_case', options_test_cases)
```
hiranya911: Pass `ids=` parameter here. Then you don't have to use `test_case.name` in the body of the test case.

tests/test_app.py
```diff
@@ -140,6 +162,127 @@ def test_app_init_with_invalid_name(self, name):
...
+        config_old = set_config_env(bad_file_name)
+        with pytest.raises(ValueError):
+            firebase_admin.initialize_app(CREDENTIAL)
+        revert_config_env(config_old)
+
+
+    OptionsTestCase = namedtuple('OptionsTestCase',
+                                 'name, config_json, init_options, want_options')
+    options_test_cases = [
+        OptionsTestCase(name='no env var, empty options',
```
hiranya911: Can you clean up/clarify the test case names a bit? Perhaps get some ideas from the Java PR?

tests/test_app.py
avishalom: I've put it in the fixture according to your first suggestion, thanks. 
PTAL

firebase_admin/__init__.py
```diff
@@ -144,17 +147,44 @@ class _AppOptions(object):
...
+            json_str = config_file
+        else:
+            try:
+                with open(config_file, 'r') as json_file:
+                    json_str = json_file.read()
+            except Exception as err:
+                raise ValueError('Unable to read file {}. {}'.format(config_file, err))
+        try:
+            json_data = json.loads(json_str)
+        except Exception as err:
```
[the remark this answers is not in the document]
avishalom: we can,
the error message is "Expecting property name: line 1 column 2 (char 1)"
I think it is clearer if we add information, but willing to remove if you want it gone.

tests/test_app.py
```diff
@@ -93,6 +93,107 @@ def init_app(request):
     else:
         return firebase_admin.initialize_app(CREDENTIAL)
 
+@pytest.fixture(scope="function")
+def env_test_case(request):
+    config_old = set_config_env(request.param.config_json)
+    def fin():
```
hiranya911: Can we use `yield ` here? http://pytest.readthedocs.io/en/2.8.7/yieldfixture.html

tests/test_app.py
```diff
@@ -172,116 +273,20 @@ def test_app_init_with_invalid_config_file(self, bad_file_name):
...
-        app = firebase_admin.initialize_app(CREDENTIAL, options=test_case.init_options)
+    @pytest.mark.parametrize('env_test_case', env_options_test_cases,
+                             ids=[x.name for x in env_options_test_cases],
+                             indirect=['env_test_case'])
+    def test_app_init_with_default_config(self, env_test_case):
+        app = firebase_admin.initialize_app(CREDENTIAL, options=env_test_case.init_options)
         for field in firebase_admin._CONFIG_VALID_KEYS:
-            assert app.options.get(field) == test_case.want_options.get(field), test_case.name
-        revert_config_env(config_old)
+            assert app.options.get(field) == env_test_case.want_options.get(field)
```
hiranya911: Can we do a stronger check here? Something like `assert app.options._options == env_test_cast.want_options`?

firebase_admin/__init__.py
```diff
@@ -46,10 +49,14 @@ def initialize_app(credential=None, options=None, name=_DEFAULT_APP_NAME):
       credential: A credential object used to initialize the SDK (optional). If none is provided,
           Google Application Default Credentials are used.
       options: A dictionary of configuration options (optional). Supported options include
-          ``databaseURL``, ``storageBucket`` and ``httpTimeout``. If ``httpTimeout`` is not set,
-          HTTP connections initiated by client modules such as ``db`` will not time out.
+          ``databaseURL``, ``storageBucket``, ``projectId``, ``databaseAuthVariableOverride``
+          and ``httpTimeout``. If ``httpTimeout`` is not set, HTTP connections initiated by client
+          modules such as ``db`` will not time out.
+          If options are not provided an attempt is made to load the options from the environment.
```
hiranya911: This should be part of the method description, rather than the argument description.

Issue #61, opened 2017-08-18
FR: Provide a "Shallow" option when getting data


grennis:
Please add support for the [shallow option](https://firebase.google.com/docs/database/rest/retrieve-data#shallow) when retrieving data from the realtime db.

giusepperj:
@hiranya911 When can we expect this to get merged? thank you 😄

hiranya911:
@giusepperj Unfortunately, this PR is broken, and doesn't have any tests. Until those issues are addressed, we cannot make any progress with it. If you or somebody else can put in some time to clean up the code, and provide some tests, I'll be glad to take it to the next stage of the review process.

hiranya911:
I would like to make some progress with this.

@rodolphopivetta will you be able to provide some test cases for us? Otherwise I'll use your PR as a starting point and roll out my own.

hiranya911:
This was released in v2.8.0

PR #83, opened 2017-10-17
#61 add support to shallow param

rodolphopivetta:
I just added the param to [shallow](https://firebase.google.com/docs/database/rest/retrieve-data#shallow) results

PR #109, opened 2018-01-08
Shallow reads support in database API

hiranya911:
Adds an optional `shallow` argument to the `db.Reference.get()` method. This improves upon #83.

go/firebase-python-shallow-get


hiranya911:
Resolves #61

PR #119, opened 2018-01-19
Create CHANGELOG.md

teskey:
Create changlelog from https://firebase.google.com/support/release-notes/admin/python

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,21 +5,21 @@
 
 ### Initialization
 
-- [added] The [`initialize_app()`](/docs/reference/admin/node/admin#.initializeApp)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,146 @@
+# Unreleased
+- 
+
+# v2.8.0
+
+### Initialization
+
+- [added] The [`initialize_app()`](/docs/reference/admin/node/admin#.initializeApp)
```
hiranya911: We use relative URLs (starting with `/docs/`) in the Firebase devsite, but that won't work in Github. Should we fix that?

teskey: Oh yeah, I thought I had fixed those everywhere by putting in the full path.. I did search and replace so not sure how I missed this one or others, so let me double check and address

PR #130, opened 2018-02-16
Add snippets to the admin repo

samtstern:
They are linted by `lint.sh`, not sure if that fails below a certain score?


hiranya911:
The linter fails the build if it detects at least one error. But looks like the linter passed on this one. From the last build log:

```
$ if [[ "$PY_VERSION" == '2' ]]; then ./lint.sh all; fi
Running linter on module firebase_admin
Running linter on module tests
Running linter on module integration
Running linter on module snippets
```

We have more snippets at https://github.com/firebase/quickstart-python. I think we should move them here as well.

samtstern:
@hiranya911 I added all the relevant `quickstart-python` snippets here and also moved the files around to be in subfolders.  Lint gets 10/10 now.

samtstern:
However the `581a9dc` should have failed due to lint, and it did not...

samtstern:
I'll wait for @hiranya911 to merge.

PR #136, opened 2018-03-13
Added support for mutable-content flag in aps

carsongee:
This adds an additional flag to the Aps class to support mutable notifications: https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/ModifyingNotifications.html

### DIFF
--- a/firebase_admin/messaging.py
+++ b/firebase_admin/messaging.py
@@ -341,16 +341,19 @@ class Aps(object):
         sound: Name of the sound file to be played with the message (optional).
         content_available: A boolean indicating whether to configure a background update
             notification (optional).
+        mutable_content: A boolean indicating whether to have the message be mutable
+            on device before final delivery (optional).
         category: String identifier representing the message type (optional).
         thread_id: An app-specific string identifier for grouping messages (optional).
     """
 
-    def __init__(self, alert=None, badge=None, sound=None, content_available=None, category=None,
-                 thread_id=None):
+    def __init__(self, alert=None, badge=None, sound=None, content_available=None,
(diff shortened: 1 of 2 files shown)


firebase_admin/messaging.py
```diff
@@ -341,16 +341,19 @@ class Aps(object):
...
+        mutable_content: A boolean indicating whether to have the message be mutable
+            on device before final delivery (optional).
         category: String identifier representing the message type (optional).
         thread_id: An app-specific string identifier for grouping messages (optional).
     """
 
-    def __init__(self, alert=None, badge=None, sound=None, content_available=None, category=None,
-                 thread_id=None):
+    def __init__(self, alert=None, badge=None, sound=None, content_available=None,
+                 mutable_content=None, category=None, thread_id=None):
```
hiranya911: `mutable-content` does not seem to be a standard parameter APNS supports but one of the many possible extensions. I think the right way to support this type of options is by adding a `**kwargs` parameter to the `Aps.__init__()` constructor (or may be a regular dict parameter, so we don't have to deal with mapping underscores to hyphens). In any case, this is an API change, and will need to go through our internal review process. Please hang tight.

carsongee: I had thought of that as a good option as well, I'm happy to swap to that if you'd like, just let me know.  I'll tell you I chose it as is because I think it will become official, it is just newish and I think the doc reference hasn't been updated.

hiranya911:
@carsongee thanks again for the PR. I've combined your work with an additional `custom_fields` argument to make the API a bit more future proof (see #143). Let me know what you think.

PR #143, opened 2018-03-22
Implementing support for APNS mutable-content field

hiranya911:
Added `mutable_content` and `custom_data` arguments to `Aps.__init__()`.

Resolves #136

### DIFF
--- a/tests/test_messaging.py
+++ b/tests/test_messaging.py
@@ -537,11 +537,9 @@ class TestAPNSPayloadEncoder(object):
 
 class TestApsEncoder(object):
 
-    def _check_aps(self, aps):
-        with pytest.raises(ValueError) as excinfo:
-            check_encoding(messaging.Message(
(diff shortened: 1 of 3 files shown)


tests/test_messaging.py
```diff
@@ -587,6 +587,28 @@ def test_invalid_thread_id(self, data):
         expected = 'Aps.thread_id must be a string.'
         assert str(excinfo.value) == expected
 
+    @pytest.mark.parametrize('data', ['', list(), tuple(), True, False, 1, 0, ])
+    def test_invalid_custom_data_dict(self, data):
+        if isinstance(data, dict):
+            return
+        aps = messaging.Aps(custom_data=data)
+        excinfo = self._check_aps(aps)
```
avishalom: I have a slight preference, here and elsewhere to inline this check, and get the with .. raises in the same test function as it is easier to read, and only 2 lines longer. But this is ok too. (maybe call it _check_aps_exception())

hiranya911: Refactored as suggested.

PR #148, opened 2018-04-04
Snippets for Session Management

hiranya911:
Snippets demonstrate the Flask webapp environment. These will appear in https://firebase.google.com/docs/auth/admin/manage-cookies


hiranya911:
I think developers will be able to port these samples to their preferred web frameworks. In case of Node.js we show samples assuming the Express framework.

Issue #157, opened 2018-05-06
Validate Python 3 compatible syntax


ncouture:
There are at least some Python >= 3 invalid syntax used in this project (see print statements added in #130).

This suggests there is currently no such validation in place or the snippets are not validated to be compatible with Python >= 3.

ncouture:
As a side note since this ticket discusses syntax validation;  I'm not sure what's the general rule of thumb when it comes to the Python syntax style used in their projects but as there a few inconsistencies with their style guide for instance `.pylintrc` defines line length to be of a maximum of 100 characters and [your style guide](https://google.github.io/styleguide/pyguide.html#Line_length ) prescribes 80 characters.

I would like to suggest using [yapf](https://github.com/google/yapf). For example the following yapf configuration makes it apply many changes based on Google's Python style guide:
```
[style]
based_on_style = google
```

hiranya911:
We test the sources on both Python 2 and 3. The snippets have been upgraded to Python 3 syntax as of #157.

PR #159, opened 2018-05-06
Using Python 3 compatible print statements in snippets (#157)

ncouture:
* Fixed Python 3 incompatible print syntax used in snippets


hiranya911:
If we are to accept this, you also need to provide a solution for linting on Python 2. Otherwise it's going to break the build: https://travis-ci.org/firebase/firebase-admin-python/jobs/375511853

I guess easiest fix would be to skip the corresponding pylint check for snippets.

ncouture:
As per last commit; snippets have been modified to include the following import statement in order to make **pylint** aware of future print syntax when executed under Python < 3:
```python
from __future__ import print_function
```

snippets/messaging/cloud_messaging.py
```diff
@@ -35,7 +37,7 @@ def send_to_token():
     # registration token.
     response = messaging.send(message)
     # Response is a message ID string.
-    print('Successfully sent message:', response)
+    print(('Successfully sent message:', response))
```
hiranya911: This is going to print out a tuple, which is not what we want. 

```
>>> print(('foo', 'bar'))
('foo', 'bar')
```

Please remove the redundant parenthesis -- here and below. Then I can merge this PR.

PR #165, opened 2018-05-17
'?' is now an illegal character for the REST API

Bien-CV:
Having a '?' character in a path raised an ApiCallError (405 Client Error: Method Not Allowed for url: -fakeURLwith?-.json Reason: append .json to your request URI to use the REST API. ).

It is indeed not the right error. 

This pull request allows the correct error to be raised ( ValueError : 'Invalid path: "{0}". Path contains illegal characters.' ).

### DIFF
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -34,7 +34,7 @@ from firebase_admin import _utils
 
 
 _DB_ATTRIBUTE = '_database'
-_INVALID_PATH_CHARACTERS = '[].#$'
+_INVALID_PATH_CHARACTERS = '[].?#$'
 _RESERVED_FILTERS = ('$key', '$value', '$priority')
 _USER_AGENT = 'Firebase/HTTP/{0}/{1}.{2}/AdminPython'.format(
     firebase_admin.__version__, sys.version_info.major, sys.version_info.minor)


Bien-CV:
CLA signed.

hiranya911:
@Bien-CV thanks for the PR. However, `?` is a valid path character in Firebase Realtime Database. I wonder if we should do something special to handle them in Python.

@rockwotj any thoughts on how to fix this? Simple URL encoding of the "?" is not enough it seems.

rockwotj:
@hiranya911 we've got a really bad bug that's been out for 5 years or something. We have broken behavior with URL decoding where we accidentally do it twice. That breaks most of these things, and we don't have a good fix because there are apps that rely on the broken behavior. We're trying to come up with a way to fix it.

Bien-CV:
The '?' issue bothers all developers trying to use URLs as keys in firebase.

hiranya911:
@rockwotj just wanted to ping you again on this. Is it reasonable to not accept `?` as a valid path character (at least until the issue in the server-side gets resolved)?

rockwotj:
There is no harm, I'm fine with it. I'll ping this thread when the issue is deployed.

PR #171, opened 2018-05-29
Updated API docs

hiranya911:
Also renamed `UserImportRecord` to `ImprotUserRecord` as agreed upon in the API proposal.

go/firebase-admin-user-import


firebase_admin/_user_import.py
```diff
@@ -346,10 +435,26 @@ def scrypt(cls, key, rounds, memory_cost, salt_separator=None):
 
     @classmethod
     def bcrypt(cls):
```
avishalom: Might be staticmethod, but good either way.

PR #175, opened 2018-05-31
Support for creating custom tokens without service account credentials

hiranya911:
Currently the SDK must be initialized with service account credentials in order to be able to call FirebaseAuth.createCustomToken(). With this PR, the SDK will attempt to sign custom tokens by calling the IAM service in the cloud when the service account credentials are not provided.

```
# Following will work in GCP managed runtimes where the Metadata service is present
firebase_admin.initialize_app()
token = auth.create_custom_token(uid)
```

```
# Following will work anywhere
options = {
    'service_account_id': 'test-service-account@project-id.iam.gserviceaccount.com',
}
firebase_admin.initialize_app(options=options)
token = auth.create_custom_token(uid)
```

go/firebase-admin-sign
go/firebase-admin-iam-sign

### DIFF
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -119,7 +119,7 @@ class TokenGenerator(object):
 
         # Attempt to discover a service account email from the local Metadata service. Use it
         # with the IAM service to sign bytes.
-        resp = self.request(url=METADATA_SERVICE_URL)
+        resp = self.request(url=METADATA_SERVICE_URL, headers={'Metadata-Flavor': 'Google'})
         service_account = resp.data.decode()
         return _SigningProvider.from_iam(self.request, google_cred, service_account)
 
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -130,10 +130,12 @@ class TokenGenerator(object):
             try:
                 self._signing_provider = self._init_signing_provider()
             except Exception as error:
+                url = 'https://firebase.google.com/docs/auth/admin/create-custom-tokens'
                 raise ValueError(
                     'Failed to determine service account: {0}. Make sure to initialize the SDK '
-                    'with a service account credential. Alternatively, specify a service account '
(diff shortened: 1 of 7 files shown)


firebase_admin/_token_gen.py
```diff
@@ -60,20 +65,79 @@ def __init__(self, code, message, error=None):
...
+    def _init_signing_provider(self):
+        """Initializes a signing provider by following the go/firebase-admin-sign protocol."""
+        # If the SDK was initialized with a service account, use it to sign bytes.
+        google_cred = self.app.credential.get_credential()
+        if isinstance(google_cred, google.oauth2.service_account.Credentials):
+            return _SigningProvider.from_credential(google_cred)
+
+        # If the SDK was initialized with a service account email, use it with the IAM service
+        # to sign bytes.
+        service_account = self.app.options.get('service_account_id')
```
bjornick: So we don't want them to be able to override the default GCE credentials?

firebase_admin/auth.py
```diff
@@ -109,10 +109,13 @@ def create_custom_token(uid, developer_claims=None, app=None):
 
     Raises:
         ValueError: If input parameters are invalid.
+        AuthError: If an error occurs while creating the token using the remote IAM service.
     """
     token_generator = _get_auth_service(app).token_generator
-    return token_generator.create_custom_token(uid, developer_claims)
-
+    try:
+        return token_generator.create_custom_token(uid, developer_claims)
```
bjornick: It looks like we are wrapping some calls with try catch (I assume because we now no longer assume that IAM calls work).  How confident are you that you caught all the cases.  This feels like it would be easy to miss one of these cases.

tests/test_token_gen.py
```diff
@@ -194,6 +198,61 @@ def test_noncert_credential(self, user_mgt_app):
         with pytest.raises(ValueError):
             auth.create_custom_token(MOCK_UID, app=user_mgt_app)
 
+    def test_sign_with_iam(self):
```
bjornick: None of these test exercise the new error cases you added, right?

PR #176, opened 2018-06-26
Reading project ID from GOOGLE_CLOUD_PROJECT

hiranya911:
`GCLOUD_PROJECT` is no longer supported in some managed Google runtimes. `GOOGLE_CLOUD_PROJECT` is the new recommended environment variable.

### DIFF
--- a/firebase_admin/__init__.py
+++ b/firebase_admin/__init__.py
@@ -236,18 +236,18 @@ class App(object):
         Raises:
             ValueError: If a non-string project ID value is specified.
         """
-        pid = options.get('projectId')
-        if not pid:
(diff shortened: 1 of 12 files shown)


firebase_admin/__init__.py
```diff
@@ -242,7 +243,7 @@ def _lookup_project_id(cls, credential, options):
             except AttributeError:
                 pass
         if not pid:
```
bjornick: hmm. in hindsight pid seems like an unfortunate name for project id.  Can we change that?

hiranya911: Done

Issue #173, opened 2018-05-30
FR: Handle multiple databases


bitromedian:
Code from https://firebase.google.com/docs/database/admin/start

```python
import firebase_admin
from firebase_admin import credentials
from firebase_admin import db

# Fetch the service account key JSON file contents
cred = credentials.Certificate('path/to/serviceAccountKey.json')

# Initialize the app with a service account, granting admin privileges
firebase_admin.initialize_app(cred, {
    'databaseURL': 'https://databaseName.firebaseio.com'
})

# As an admin, the app has access to read and write all data, regradless of Security Rules
ref = db.reference('restricted_access/secret_document')
print ref.get()
```

The `db` seems a singleton variable so it's not able to handle multiple databases at the same time.
Any ideas?

hiranya911:
`db` is module. Way to support this is by passing the database URL as an optional parameter to `db.reference()`. This is in our roadmap.

bitromedian:
I managed to use two databases like below:

```
firebase_admin.initialize_app(cred, {
    'databaseURL': 'https://databaseName.firebaseio.com'
})

app2 = firebase_admin.initialize_app(cred, {
    'databaseURL': 'https://databaseName-app2.firebaseio.com'
}, name='app2')

ref = db.reference('restricted_access/secret_document')
ref2 = db.reference('restricted_access/secret_document', app2)

print(ref.get())
print(ref2.get())
```

hiranya911:
This has been implemented and will be included in next release.

PR #178, opened 2018-07-02
Multi DB support

hiranya911:
Makes it easier to access different Firebase Database instances in the same project:

```
import firebase_admin
from firebase_admin import db

firebase_admin.initialize_app(options={
    'databaseURL': 'https://default.firebaseio.com',
})
ref = db.reference()
other_ref = db.reference(url='https://other.firebaseio.com')
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+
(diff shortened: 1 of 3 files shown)


hiranya911:
Resolves #173

PR #50, opened 2017-07-18
Adding stream listeners to sniff change in child

rizasif:
Hello,

I noticed that there (according to my best knowledge) hasn't been a stream listener implemented in the project. Which has a sole responsibility to fire up a callback function (https://firebase.google.com/docs/reference/rest/database/#callback) when any change to a firebase node is occurred. This is essential for developers who want to perform an action when a change to database is observed.

The changes I proposed are inspired by PyRebase (https://github.com/thisbejim/Pyrebase). I have implemented a SSEClient which uses python request library to listen for changes. These changes have been tested using a simulation of 100 users, performing read and write operations. The results, so far, are promising with 0% failure rate. 

I have written a test code for you as well. The following adds a new child node to the already available parent "users". Now, if you visit the firebase console and delete this new child node, then the "call_function" will fire up. Similar, case can be observed by updating, adding or deleting in the child of "users" node through the console.

```
import firebase_admin
from firebase_admin import credentials
from firebase_admin import db

# Fetch the service account key JSON file contents
cred = credentials.Certificate("PATH_TO_JSON_FILE")

# Initialize the app with a service account, granting admin privileges
firebase_admin.initialize_app(cred, {"databaseURL": "DATABASE_URL"})

root = db.reference()
print(root.get())


new_user = root.child("users").push({"name": "lila Annen", "age": 19})
ref = root.child("users")
print(ref.get())

def call_function(message):
  print( message["event"])
  print(message["path"])

my_stream_question_requested = db.reference("users").stream(call_function, stream_id="new stream")
```

Hope to hear from you soon, thanks!

PR #183, opened 2018-07-21
[WIP] Adding stream listener to stream changes in child nodes

the-c0d3r:
Hi, 

This PR's original intent is to add stream listener function to fire a callback function when changes in the database is detected. It is a revival of this https://github.com/firebase/firebase-admin-python/pull/50 1-year-old PR and also based on the code contributed by @rizasif and @Aqsa-K in that PR. This PR is mainly to serve 3 purposes.
1. Update the old PR with a rebase on master
2. Clean up, refactor & fix lint errors
3. Write tests

For now, I have completed the first and second of the above. Please advise me on how to properly write the test in PyTest. I tested the code manually by defining callback and manually modifiying data in my firebase demo project. And it seems to be fired everytime I made changes. I read through the existing test suite and I cannot think of a way to write into it yet.

Changes:
- `lint.sh` : Ignore directive for "protected-access"
- `firebase_admin/sseclient.py` : SSE client module
- `firebase_admin/db.py` : Streaming functionality
- New API call: `database.reference('/xxx').stream(callbackFunc)`

Regards


hiranya911:
Please revert the changes to `lint.sh`. Instead add the directives to the lines in the source files where the lint errors should be ignored.

Rename `sseclient.py` to `_sseclient.py` so it will be considered an internal module.

For unit testing, you can mount an adapter on the `requests.Session` object and have it return some fake events (I think). See how we do something similar in the existing tests:

https://github.com/firebase/firebase-admin-python/blob/70d98173cdb3df6bc494b076717f3970f9029a7e/tests/test_db.py#L128-L132

The [`MockAdapter`](https://github.com/firebase/firebase-admin-python/blob/master/tests/testutils.py#L124) used in the above example responds to requests made by the SDK with a fake response. You will probably end up writing a similar adapter that responds with a fake stream of events.

firebase_admin/db.py
```diff
@@ -155,10 +155,10 @@ def parent(self):
 
     def build_headers(self, token=None):
         headers = {'content-type' : 'application/json; charset=UTF-8'}
-        if not token and self._client._session.credentials:
+        if not token and self._client._session.credentials: # pylint: disable=protected-access
```
hiranya911: I think `self._client.session` should work here. `session` is exposed from the `HttpClient`:

https://github.com/firebase/firebase-admin-python/blob/c420b4f4e50b99ca62b7bc5c71e11d87219c4775/firebase_admin/_http_client.py#L53-L55

That way you won't need the linter directives at all.

the-c0d3r: Oh, I didn't catch that earlier. I have changed it, thanks. I still have yet to write the tests. I'll finish the tests in a few days.

Aqsa-K: great work @the-c0d3r . I am glad that this is finally progressing. Let me know if any help is required.

the-c0d3r:
Hi, I have added the test for SSEClient code. I'm not sure if I did it correctly or not. I added `MockSSEClientAdapter()` which is a subclass of `MockAdapter`. And I captured the actual event raw response and put it in payload. But somehow when I run the test, the MockSSEClient actually sends out Bytes object which is expected, but inside SSEClient code's `response.iter_content` returns bytes instead of string. I compared it with the my program to actually communicate with the server, and somehow `iter_content` returns string. Therefore for now, I added a check for the type in SSEClient, to call `decode()` if the `nextchar` from `response.iter_content` is a bytes object. 

```
                nextchar = next(self.resp_iterator)
                if isinstance(nextchar, bytes):
                    nextchar = nextchar.decode()
                self.buf += nextchar
```
Python will raise `TypeError: Can't convert 'bytes' object to str implicitly` if executed the test withouth this check. Is there anyway to modify the MockSSEClientAdapter or the parent class, to make it's response.iter_content return string instead of bytes?
But I'm not sure about this check yet so I haven't committed this change. Please comment.

hiranya911:
Hi @the-c0d3r. I think this is due to the way `iter_content()` is implemented in `requests`:

http://docs.python-requests.org/en/master/_modules/requests/models/#Response.iter_content

There are couple of things you can try out.
1. Try setting `resp.raw.stream = True` before returning from `send()`.
2. Try setting an `encoding` on the mock response as it is expected [here](https://github.com/requests/requests/blob/master/requests/utils.py#L494).

One of these is likely to fix the issue.

hiranya911:
Thanks @the-c0d3r. This still needs to go through our internal API review process (which I have already initiated). We might get some feedback from that, which will need to be implemented -- but we can do that in steps, after merging this one. I will also take another closer look at all the code here, and post some feedback.

firebase_admin/_sseclient.py
```diff
@@ -0,0 +1,198 @@
+"""SSEClient module to handle streaming of realtime changes on the database
```
hiranya911: Please add the Apache license header here.

firebase_admin/_sseclient.py
```diff
@@ -0,0 +1,198 @@
...
+    def __init__(self, url, session, build_headers, last_id=None, retry=3000, **kwargs):
+        self.should_connect = True
+        self.url = url
+        self.last_id = last_id
+        self.retry = retry
+        self.running = True
+        # Optional support for passing in a requests.Session()
+        self.session = session
+        # function for building auth header when token expires
+        self.build_headers = build_headers
```
hiranya911: Do we really need this? The session we get from the db module is an instance of `google.auth.transport.requests.AuthorizedSession`, which is guaranteed to add the required Authorization header. So this shouldn't be required.

firebase_admin/_sseclient.py
```diff
@@ -0,0 +1,198 @@
...
+        # Any extra kwargs will be fed into the requests.get call later.
+        self.requests_kwargs = kwargs
+
+        # The SSE spec requires making requests with Cache-Control: nocache
+        if 'headers' not in self.requests_kwargs:
+            self.requests_kwargs['headers'] = {}
+        self.requests_kwargs['headers']['Cache-Control'] = 'no-cache'
+
+        # The 'Accept' header is not required, but explicit > implicit
+        self.requests_kwargs['headers']['Accept'] = 'text/event-stream'
```
hiranya911: The above can be written as:

```
headers = self.request_kwargs.get('headers', {})
# add the required values to headers
self.request_kwargs['headers'] = headers
```

firebase_admin/_sseclient.py
```diff
@@ -0,0 +1,198 @@
...
+
+        self._connect()
+
+    def close(self):
+        """Close the SSE Client instance"""
+        # TODO: check if AttributeError is needed to catch here
+        self.should_connect = False
+        self.retry = 0
+        self.resp.close()
+        #  self.resp.raw._fp.fp.raw._sock.shutdown(socket.SHUT_RDWR)
```
hiranya911: Remove the commented out lines?

On a side note, I was testing this branch the other day, and I noticed that calling close() here does not immediately release the underlying socket. As a result the Stream remains active for a while even after calling `stream.close()`. So perhaps we do need the commented out lines?

firebase_admin/_sseclient.py
```diff
@@ -0,0 +1,198 @@
...
+        #  self.resp.raw._fp.fp.raw._sock.close()
+
+
+    def _connect(self):
+        """connects to the server using requests"""
+        if self.should_connect:
```
hiranya911: Should also check for `self.running`.

firebase_admin/_sseclient.py
```diff
@@ -0,0 +1,198 @@
...
+
+    sse_line_pattern = re.compile('(?P<name>[^:]*):?( ?(?P<value>.*))?')
+
+    def __init__(self, data='', event='message', event_id=None, retry=None):
+        self.data = data
+        self.event = event
+        self.event_id = event_id
+        self.retry = retry
+
+    def dump(self):
```
hiranya911: I think this whole method can be removed.

firebase_admin/db.py
```diff
@@ -23,16 +23,25 @@
...
 
 import requests
 import six
 from six.moves import urllib
+from google.auth import transport
 
 import firebase_admin
 from firebase_admin import _http_client
 from firebase_admin import _utils
+from firebase_admin._sseclient import SSEClient, KeepAuthSession
```
hiranya911: As a matter of style, we don't import individual classes and members. Please import the module here as `from firebase_admin import _sseclient`, and change the code below accordingly.

firebase_admin/db.py
```diff
@@ -69,6 +78,49 @@ def _parse_path(path):
...
+
+    def start(self):
+        """Start the streaming by spawning a thread"""
+        self.thread = threading.Thread(target=self.start_stream)
+        self.thread.start()
+        return self
+
+    def start_stream(self):
+        """Streaming function for the spawned thread to run"""
+        self.sse = SSEClient(
```
hiranya911: Can you move this to the start() method above, before the thread is started? Then I think you will be able to drop the sleep() call in close().

firebase_admin/db.py
```diff
@@ -101,6 +153,23 @@ def parent(self):
             return Reference(client=self._client, segments=self._segments[:-1])
         return None
 
+    def build_headers(self, token=None):
```
hiranya911: See if we can remove this altogether (see my comment on _sseclient). If we must keep this, please rename to _build_headers() so it's treated as an internal method.

tests/test_db.py
```diff
@@ -523,7 +523,7 @@ def test_range_query(self):
         assert recorder[0].headers['User-Agent'] == db._USER_AGENT
 
 
-class TestDatabseInitialization(object):
+class TestDatabaseInitialization(object):
```
hiranya911: Good catch 👍

tests/test_sseclient.py
```diff
@@ -0,0 +1,79 @@
...
+
+        assert sseclient.url == self.test_url
+        assert sseclient.running
+        assert sseclient.session != None
+
+    def test_event(self):
+        sseclient = self.init_sse()
+        for msg in sseclient:
+            event = json.loads(msg.data)
+            break
```
hiranya911: Is the break necessary? Shouldn't it exit on its own since the mock response only has one event?

firebase_admin/_sseclient.py
the-c0d3r: I also noticed the delay in calling `stream.close()`. I was calling it in a single thread, and it takes a few seconds to close and return. I'll test it out a bit more after I followed your other suggestions first.

firebase_admin/db.py
the-c0d3r: This `build_headers()` is tested to be unnecessary just like your comment on `_sseclient`. So I will remove it from all subsequent code.

firebase_admin/_sseclient.py
the-c0d3r: Method Removed, and it's accompanying test in `test/test_sseclient.py` removed.

firebase_admin/_sseclient.py
the-c0d3r: `build_header()` and all subsequent function arguments removed.

the-c0d3r:
Hi @hiranya911, I believe I have addressed all of your comments with my latest commits, except the part about the closing of the stream. I wrote a simple python code to just setup stream and close it and record the timing. It returned within 100 ms. But somehow when executing the code line by line manually in python shell, it would take over 10 seconds to close the stream (aka time taken for python to return control to user). It will take me some time to debug this since I don't know much about requests module's internals. If you can shed some light on it, that would be great.

firebase_admin/_sseclient.py
```diff
@@ -23,27 +34,31 @@ def rebuild_auth(self, prepared_request, response):
...
+          session: the requests.session()
+          last_id: optional id
+          retry: the interval in ms
+          **kwargs: extra kwargs will be sent to requests.get
+        """
         self.should_connect = True
         self.url = url
         self.last_id = last_id
         self.retry = retry
         self.running = True
```
hiranya911: Remove this as it gets set to False below.

wamburu:
+1

hiranya911:
@the-c0d3r the internal review process for this is chugging along. I will have an update for you next week.

the-c0d3r:
Hi @hiranya911 thanks for the update. I have synced the PR with the latest master. If there are any further tasks, please let me know.

hiranya911:
@the-c0d3r I have some feedback. Once they are implemented, I can merge this PR:

1. Rename the `stream()` method to `listen()`.
2. Rename the `Stream` type to `ListenerRegistration`.
3. Remove the `stream_id` parameter from the `listen()` method.

Let me know if you have any questions.

the-c0d3r:
Hi, I think I have addressed those feedbacks. Please take a look.

hiranya911:
Thanks @the-c0d3r. This looks pretty solid. I'll merge it shortly. I'll intend to do a bit of work on top it over the next few days, before release.

hiranya911:
Upon taking another look, this might have a problem. I don't see how the credentials are passed into the underlying session.

hiranya911:
Ok, managed to fix the issue. I'll send a separate PR with the fix.

PR #190, opened 2018-08-08
Fixed and refactored the db.Reference.listen() API

hiranya911:
Continuation of #183 

* `KeepAuthSession` now extends from the `AuthorizedSession`. This allows the `listen()` API to properly authorize outgoing requests.
* Added `db.Event` type. The callback gets fired with instances of this class.
* Updated documentation and changelog.
* Cleaned up the code and added more unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [added] The `db.Reference` type now provides a `listen()` API for
+  receiving realtime update events from the Firebase Database.
 - [added] The `db.reference()` method now optionally takes a `url`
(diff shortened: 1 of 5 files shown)


hiranya911:
@the-c0d3r please take a look.

the-c0d3r:
@hiranya911 I have looked at the changes. It looks good, much cleaner than what I wrote. But the problem you mentioned in #183 I do not see the problem. But I have been using the SDK with the authentication json file. So that may be obscuring the problem.

hiranya911:
@the-c0d3r You probably have your database security rules set to public, allowing unauthenticated access to the data. I encountered the issue when testing with a service account json file.

PR #209, opened 2018-10-31
Introduce a new service, the Firebase Project Management Service

weixifan:
An initial CL to establish the presence of the service.

PR #212, opened 2018-11-01
Handling HTTP errors from local metadata service

hiranya911:
It seems some Travis CI nodes now respond to http://metadata URL calls. Therefore we should handle possible errors.


hiranya911:
Related to the build failure mentioned in #209

PR #216, opened 2018-11-07
More Implementation in project_management.py

weixifan:
Implement iOS equivalents of existing methods, including some refactoring; implement `set_display_name` for both platforms.


firebase_admin/project_management.py
```diff
@@ -0,0 +1,495 @@
...
+
+_PROJECT_MANAGEMENT_ATTRIBUTE = '_project_management'
+
+
+def _get_project_management_service(app):
+    return _utils.get_app_service(app, _PROJECT_MANAGEMENT_ATTRIBUTE, _ProjectManagementService)
+
+
+def android_app(app_id, app=None):
+    """Obtains a reference to an Android App in the associated Firebase project.
```
hiranya911: Nit: `s/App/app/` here and elsewhere in the documentation.

firebase_admin/project_management.py
```diff
@@ -0,0 +1,495 @@
...
+    def _extract_message(self, identifier, identifier_label, error):
+        if not isinstance(error, requests.exceptions.RequestException) or error.response is None:
+            return str(error)
+        status = error.response.status_code
+        message = _ProjectManagementService.ERROR_CODES.get(status)
+        if message:
+            return '{0} "{1}": {2}'.format(identifier_label, identifier, message)
+        return '{0} "{1}": Error {2}.'.format(identifier_label, identifier, status)
+
+    def _none_to_empty(self, string):
```
hiranya911: Make this a static or class method. Basically any helper methods that don't need `self` can be static or class method.

integration/test_project_management.py
```diff
@@ -0,0 +1,100 @@
...
+    del default_app
+    ios_apps = project_management.list_ios_apps()
+    for ios_app in ios_apps:
+        if ios_app.get_metadata().display_name.startswith(TEST_APP_DISPLAY_NAME_PREFIX):
+            return ios_app
+    return project_management.create_ios_app(
+        bundle_id=TEST_APP_BUNDLE_ID, display_name=TEST_APP_DISPLAY_NAME_PREFIX)
+
+
+def test_android_set_display_name_and_get_metadata(test_android_app, project_id):
```
hiranya911: Nit: Could be just called `test_android_app_metadata`

firebase_admin/project_management.py
weixifan: Done. I have also made the attribute names "display name" and "app ID" consistently miniscule as well, to avoid the term `App`.

firebase_admin/project_management.py
weixifan: Done.

PR #217, opened 2018-11-08
Implement get_config and all SHA certificate-related methods

weixifan:
This PR concludes the main implementation work of the `project_management` module. New integration tests have been added to cover these new functionalities.

The final few upcoming PRs will add a unit test suite for the `project_management` module.


firebase_admin/project_management.py
```diff
@@ -147,46 +324,130 @@ def __init__(self, name, app_id, display_name, project_id):
...
+        """The permanent, globally unique, user-assigned ID of the parent Firebase project."""
         return self._project_id
 
 
 class AndroidAppMetadata(_AppMetadata):
-    """Android-specific information about an Android Firebase App."""
+    """Android-specific information about an Android Firebase app."""
 
-    def __init__(self, name, app_id, display_name, project_id, package_name):
+    def __init__(self, package_name, name, app_id, display_name, project_id):
```
hiranya911: It seems to me the argument order you had earlier is the more appropriate one. Commons args followed by more type specific args.

firebase_admin/project_management.py
```diff
@@ -214,49 +482,120 @@ def __init__(self, app):
         self._timeout = app.options.get('httpTimeout')
 
     def get_android_app_metadata(self, app_id):
-        """Retrieves detailed information about an Android App."""
+        return self._get_app_metadata(
+            platform_resource_name=_ProjectManagementService.ANDROID_APPS_RESOURCE_NAME,
```
hiranya911: platform_resource_name is a mouthful, but I don't have a shorter suggestion either :)

firebase_admin/project_management.py
```diff
@@ -214,49 +482,120 @@ def __init__(self, app):
...
 
     def list_android_apps(self):
-        """Lists all the Android Apps within the Firebase Project."""
-        path = '/v1beta1/projects/{0}/androidApps?pageSize={1}'.format(
-            self._project_id, _ProjectManagementService.MAXIMUM_LIST_APPS_PAGE_SIZE)
+        return self._list_apps(
+            platform_resource_name=_ProjectManagementService.ANDROID_APPS_RESOURCE_NAME,
+            app_class=AndroidApp)
+
+    def list_ios_apps(self):
```
hiranya911: What if we drop these platform-specific helper functions and directly call `_list_apps()` where it's needed? Same for others?

integration/test_project_management.py
```diff
@@ -34,22 +44,144 @@ def test_android_app(default_app):
         package_name=TEST_APP_PACKAGE_NAME, display_name=TEST_APP_DISPLAY_NAME_PREFIX)
 
 
-def test_get_android_app_metadata(test_android_app, project_id):
+@pytest.fixture(scope='module')
+def test_ios_app(default_app):
```
hiranya911: Lets not prefix fixtures with `test_`. That's confusing. Lets just call this `ios_app`? Same goes for `test_android_app` above.

integration/test_project_management.py
```diff
@@ -34,22 +44,144 @@ def test_android_app(default_app):
...
         if android_app.get_metadata().display_name.startswith(TEST_APP_DISPLAY_NAME_PREFIX):
             found = True
             break
     assert found
+
+
+def test_get_android_app_config(test_android_app, project_id):
+    config = test_android_app.get_config()
+
+    assert '"mobilesdk_app_id": "{0}"'.format(test_android_app.app_id) in config
```
hiranya911: What is the format of the `config` string? Can we do a stronger assertion here than this?

integration/test_project_management.py
```diff
@@ -14,14 +14,24 @@
```
hiranya911: Can you also start cranking out some unit tests? Note that we only run integration tests during releases. We need a set of fast and extensive unit tests, that we run during development and on our CI servers.

integration/test_project_management.py
```diff
@@ -34,22 +44,144 @@ def test_android_app(default_app):
...
+    assert metadata.display_name == new_display_name
+    assert metadata.bundle_id == TEST_APP_BUNDLE_ID
+
+
+def test_list_ios_apps(test_ios_app):
+    del test_ios_app
+
+    ios_apps = project_management.list_ios_apps()
+
+    for ios_app in ios_apps:
```
hiranya911: Can be shortened as follows: 
```
assert any([app.get_metadata().display_name.startswith(TEST_APP_DISPLAY_NAME_PREFIX)
             for app in ios_apps])
```

integration/test_project_management.py
weixifan: Please see the following pull requests #218 + #219.

firebase_admin/project_management.py
weixifan: Acknowledged. Descriptiveness and brevity are always in tension when it comes to readability; here I favour descriptiveness due to the slightly complex list of parameters we need to pass into the helper methods common to both Android and iOS.

firebase_admin/project_management.py
weixifan: This is to avoid the clutter that would otherwise appear at the module level. This way, I am able to localize the long list of platform-specific constants into the `_ProjectManagementService` class.

integration/test_project_management.py
weixifan: Done.

integration/test_project_management.py
weixifan: Done.

firebase_admin/project_management.py
weixifan: I actually wanted to leave it as the last argument as well, but then I will have trouble in `_get_app_metadata` in line 498 because there the parameter name would be different between iOS and Android, forcing me to pass it as the first positional argument and the rest as keyword arguments (which must follow positional arguments). On the other hand, I can remove keywords from all the other arguments. Please let me know if you have any suggestions.

integration/test_project_management.py
weixifan: Done.

integration/test_project_management.py
weixifan: Done.

Agreed. The format of the Android config is a JSON, while the iOS config is a .plist; I have added the relevant tests.

integration/test_project_management.py
```diff
@@ -94,26 +96,33 @@ def test_list_android_apps(test_android_app):
...
-    assert '"project_id": "{0}"'.format(project_id) in config
+    json_config = json.loads(config)
+    assert json_config['project_info']['project_id'] == project_id
+    for client in json_config['client']:
+        client_info = client['client_info']
+        if client_info['mobilesdk_app_id'] == android_app.app_id:
+            assert client_info['android_client_info']['package_name'] == TEST_APP_PACKAGE_NAME
+            break
+    else:
+        assert False, 'Failed to find the test Android app in the Android config.'
```
hiranya911: `pytest.fail(....)`

weixifan: Thanks. Done.

PR #218, opened 2018-11-09
Add an (incomplete) unit test for package_management; add a MockMultiRequestAdapter test utility

weixifan:
Feedback on the layout of the unit test as well as the `MockMultiRequestAdapter` is welcome. Note that I will make the test parameterized in a later PR so that it can be used for iOS as well.

Edit: Actually, I just tried to make the tests parameterized; however, there are just enough parameters per test to make the resulting test a lot less readable than it is now. There are simply too many parameters, I think, for that to be worth it. I'm leaning towards unit test debuggability and readability, so I'll duplicate the unit test for iOS in a later PR.


hiranya911:
I will wait until #217 is merged and this PR is sync'ed against that. As it is, it's difficult to determine the code changes I should review.

weixifan:
I have just merged #217; sorry for the delay.

tests/testutils.py
```diff
@@ -121,19 +121,35 @@ def get_credential(self):
...
         resp.url = request.url
-        resp.status_code = self._status
-        resp.raw = six.BytesIO(self._data.encode())
+        resp.status_code = self._statuses[self._current_response]
+        resp.raw = six.BytesIO(self._responses[self._current_response].encode())
+        self._current_response = min(self._current_response + 1, len(self._responses) - 1)
         return resp
+
+
+class MockAdapter(MockMultiRequestAdapter):
```
hiranya911: Can we expose `status` and `data` properties from this class to keep the API unchanged?

```
@property
def status(self):
    return self._statuses[0]
```

tests/test_sseclient.py
```diff
@@ -30,8 +30,8 @@ def __init__(self, payload, recorder):
     def send(self, request, **kwargs):
         resp = super(MockSSEClientAdapter, self).send(request, **kwargs)
         resp.url = request.url
-        resp.status_code = self._status
-        resp.raw = six.BytesIO(self._data.encode())
+        resp.status_code = self._statuses[0]
```
hiranya911: See if we can do without this. Check my comment below.

tests/testutils.py
```diff
@@ -121,19 +121,35 @@ def get_credential(self):
...
+
+        The lengths of the responses and statuses parameters must match.
+
+        Each incoming request consumes a response and a status, in order. If all responses and
+        statuses are exhausted, further requests will reuse the last response and status.
+        """
         adapters.HTTPAdapter.__init__(self)
-        self._data = data
-        self._status = status
+        assert len(responses) == len(statuses)
```
hiranya911: Throw a ValueError here. We only do asserts inside test cases.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+import firebase_admin
+from firebase_admin import project_management
+from tests import testutils
+
+BASE_URL = 'https://firebase.googleapis.com'
+
+TEST_PROJECT_ID = 'hello-world'
+TEST_ANDROID_APP_ID = '1:12345678:android:deadbeef'
+TEST_ANDROID_APP_NAME = 'projects/{0}/androidApps/{1}'.format(TEST_PROJECT_ID, TEST_ANDROID_APP_ID)
+TEST_ANDROID_APP_DISPLAY_NAME = "My Android App"
```
hiranya911: Use single quotes

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+TEST_ANDROID_APP_NO_DISPLAY_NAME = {
+    'name': TEST_ANDROID_APP_NAME,
+    'appId': TEST_ANDROID_APP_ID,
+    'projectId': TEST_PROJECT_ID,
+    'packageName': TEST_ANDROID_APP_PACKAGE_NAME,
+}
+
+TEST_IOS_APP_ID = '1:12345678:ios:ca5cade5'
+TEST_IOS_APP_NAME = 'projects/{0}/iosApps/{1}'.format(TEST_PROJECT_ID, TEST_IOS_APP_ID)
+TEST_IOS_APP_DISPLAY_NAME = "My iOS App"
```
hiranya911: Single quotes

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+IOS_APP_NO_DISPLAY_NAME_OPERATION_SUCCESSFUL_RESPONSE = json.dumps({
+    'name': OPERATION_NAME,
+    'done': True,
+    'response': TEST_IOS_APP_NO_DISPLAY_NAME,
+})
+
+ERROR_RESPONSE = 'some error'
+
+class TestCreateAndroidApp(object):
+    _CREATION_URL = '{0}/v1beta1/projects/{1}/{2}'.format(BASE_URL, TEST_PROJECT_ID, "androidApps")
```
hiranya911: Single quotes around string literal. Here and elsewhere.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+
+import pytest
+
+import firebase_admin
+from firebase_admin import project_management
+from tests import testutils
+
+BASE_URL = 'https://firebase.googleapis.com'
+
+TEST_PROJECT_ID = 'hello-world'
```
hiranya911: Make the value 'test-project-id' or something similar for clarity.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+    @classmethod
+    def teardown_class(cls):
+        testutils.cleanup_apps()
+
+    def _set_up_mock_responses_and_request_captor_for_project_management_service(
+            self, statuses, responses, app=None):
+        if not app:
+            app = firebase_admin.get_app()
+        project_management_service = project_management._get_project_management_service(app)
+        captor = []
```
hiranya911: We've been using the name 'recorder' for this in other modules. Lets be consistent with that.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+    @classmethod
+    def setup_class(cls):
+        firebase_admin.initialize_app(
+            testutils.MockCredential(), {'projectId': TEST_PROJECT_ID})
+
+    @classmethod
+    def teardown_class(cls):
+        testutils.cleanup_apps()
+
+    def _set_up_mock_responses_and_request_captor_for_project_management_service(
```
hiranya911: `_instrument_service()` -- It's shorter and consistent with the terminology used in other services.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+
+        assert android_app.app_id == TEST_ANDROID_APP_ID
+        assert len(captor) == 3
+        assert captor[0].method == 'POST'
+        assert captor[0].url == TestCreateAndroidApp._CREATION_URL
+        body = {'packageName': TEST_ANDROID_APP_PACKAGE_NAME}
+        assert json.loads(captor[0].body.decode()) == body
+        assert captor[1].method == 'GET'
+        assert captor[1].url == TestCreateAndroidApp._POLLING_URL
+        assert not captor[1].body
```
hiranya911: Feels kind of unnecessary for a GET request.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+                OPERATION_IN_PROGRESS_RESPONSE,  # Creation Operation is still not done.
+                OPERATION_FAILED_RESPONSE,  # Operation is finished, but terminated with an error.
+            ])
+
+        with pytest.raises(project_management.ApiCallError) as excinfo:
+            project_management.create_android_app(
+                package_name=TEST_ANDROID_APP_PACKAGE_NAME,
+                display_name=TEST_ANDROID_APP_DISPLAY_NAME)
+
+        assert 'Polling finished, but the Operation terminated in an error' in str(excinfo.value)
```
hiranya911: Should be "operation" (lowercase) -- Fix where the error is raised.

tests/test_project_management.py
```diff
@@ -0,0 +1,222 @@
...
+from tests import testutils
+
+BASE_URL = 'https://firebase.googleapis.com'
+
+TEST_PROJECT_ID = 'hello-world'
+TEST_ANDROID_APP_ID = '1:12345678:android:deadbeef'
+TEST_ANDROID_APP_NAME = 'projects/{0}/androidApps/{1}'.format(TEST_PROJECT_ID, TEST_ANDROID_APP_ID)
+TEST_ANDROID_APP_DISPLAY_NAME = "My Android App"
+TEST_ANDROID_APP_PACKAGE_NAME = 'com.hello.world.android'
+TEST_ANDROID_APP = {
```
hiranya911: Some of these constants are referenced only once. Can we just inline them and remove the constants?

tests/test_project_management.py

tests/test_project_management.py

tests/test_project_management.py
weixifan: I find that when I read a deeply nested test constant, I lose track of the structure of the outermost object. As an example, if we inlined `ANDROID_APP_NO_DISPLAY_NAME_OPERATION_SUCCESSFUL_RESPONSE`, in which we inline `TEST_ANDROID_APP_NO_DISPLAY_NAME`, then it's going to make the test `test_create_android_app_without_display_name` responses list (lines 125-127) much harder to digest. Currently, I believe that list is really easy to digest, with each response named and commented and still fitting in one line each.

tests/test_project_management.py

tests/test_project_management.py

tests/test_project_management.py
weixifan: The reason I named it this way is that *instrumentation* has a rather broad meaning, and while reading other unit tests, it took me a bit to figure out exactly what role it performs; the name I have describes exactly what it does. For the sake of consistency, however, I'll change it to `_instrument_service`.

tests/test_project_management.py

tests/test_project_management.py
weixifan: While it's true that the integration test will result in a HTTP 400 failure if we ever passed a body to this `GET` request, I think it's good for the unit test to catch this error as well. For example, the `json=request_body` parameter could result from a copy-paste error.

tests/test_project_management.py

tests/test_sseclient.py
weixifan: Done. I still have to change `self._status` to `self.status`, and `self._data` to `self.data`.

tests/testutils.py
weixifan: Done.

tests/testutils.py
weixifan: Done.

tests/test_project_management.py
hiranya911: I still think that we are overdoing it with constants. I often read code on vim and emacs with only basic text editing support, and I find myself jumping back and forth a lot trying to determine the full value of some constants. For instance in the above example, can we get rid of the `TEST_ANDROID_APP_NO_DISPAY_NAME` constant by inlining?

Having said that, I don't mind leaving things as is for now, and revising this later.

weixifan: I agree that inlining the simple, unnested constants is probably not going to hurt readability. This will tone down the level we are naming constants, while still making the test shorter and easier to understand for nested constants. I will, however, do most of this work in the next PR, so that my merges are easier.

PR #219, opened 2018-11-09
Add the bulk of the remaining unit tests for the project_management module

weixifan:
After this PR, only several smaller unit tests remain for `ShaCertificate`. Factored out a common base test class, `BaseProjectManagementTest`, to house shared utilities.


weixifan:
This PR is now ready for review as the previous PR #218 has been merged.

I have finished inlining most of the simple constants in the test file; I'm leaving the inlining of the SHA hash constants until the next (and final) in order to simplify some merges. Please let me know if you need it inlined in this PR.

tests/test_project_management.py
```diff
@@ -113,6 +207,19 @@ def _instrument_service(self, statuses, responses, app=None):
             testutils.MockMultiRequestAdapter(responses, statuses, recorder))
         return recorder
 
+    def _assert_request_is_correct(
+            self, request, expected_method, expected_url, expected_body=None):
+        assert request.method == expected_method
+        assert request.url == expected_url
+        if expected_body is None:
+            assert not request.body
```
hiranya911: Nit: Can this be `assert request.body is None`

tests/test_project_management.py
```diff
@@ -208,9 +309,476 @@ def test_create_android_app_polling_failure(self):
...
+    _GET_CONFIG_URL = ('https://firebase.googleapis.com/v1beta1/projects/-/androidApps/'
+                       '1:12345678:android:deadbeef/config')
+    _ADD_CERT_URL = ('https://firebase.googleapis.com/v1beta1/projects/-/androidApps/'
+                     '1:12345678:android:deadbeef/sha')
+    _LIST_CERTS_URL = ('https://firebase.googleapis.com/v1beta1/projects/-/androidApps/'
+                       '1:12345678:android:deadbeef/sha')
+    _DELETE_SHA_1_CERT_URL = 'https://firebase.googleapis.com/v1beta1/{0}'.format(SHA_1_NAME)
+    _DELETE_SHA_256_CERT_URL = 'https://firebase.googleapis.com/v1beta1/{0}'.format(SHA_256_NAME)
+
+    @pytest.fixture
```
hiranya911: Nit: Can this be a class-level constant instead of a fixture?

tests/test_project_management.py
```diff
@@ -208,9 +309,476 @@ def test_create_android_app_polling_failure(self):
...
+
+class TestIosApp(BaseProjectManagementTest):
+    _GET_METADATA_URL = ('https://firebase.googleapis.com/v1beta1/projects/-/iosApps/'
+                         '1:12345678:ios:ca5cade5')
+    _SET_DISPLAY_NAME_URL = ('https://firebase.googleapis.com/v1beta1/projects/-/iosApps/'
+                             '1:12345678:ios:ca5cade5?updateMask=displayName')
+    _GET_CONFIG_URL = ('https://firebase.googleapis.com/v1beta1/projects/-/iosApps/'
+                       '1:12345678:ios:ca5cade5/config')
+
+    @pytest.fixture
```
hiranya911: Same here.

tests/test_project_management.py
weixifan: Yeah I thought about that too, but then we have to write `TestAndroidApp.ANDROID_APP` instead of just `android_app`. We can make this a module-level constant, but you have to scroll up farther as you said in the previous PR. I actually attempted to make many more things a class-level `@fixture` for this reason, until I realized even that's not free: the parameter list for each test grows. So I chose to use `@fixture` with the only item that every single test in this class uses: the app.

tests/test_project_management.py
weixifan: See above.

tests/test_project_management.py

PR #221, opened 2018-11-14
Fixing a performance issue in SSE client

### LINKED ISSUE #198: Performance issue and copyright violation in `_sseclient.py` (for `db.Reference.listen()`)
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOS 10.13.6
  * Firebase SDK version: 2.13.0
  * Firebase Product: database

#### Steps to reproduce:

1. Have some Firebase Realtime Database path containing a significant amount of data (>200KB, say).
2. Call `db.reference('/path').listen(callback)`.

As the first update, the server will send the entire contents of that path. On my machine this pegs a CPU core for minutes before the callback is finally called. The issue is the following code: https://github.com/firebase/firebase-admin-python/blob/abd39805f83f906296d5c99b6c4943754154d02c/firebase_admin/_sseclient.py#L93-L96
Since `self.resp_iterator` iterates through the stream one character at a time and `self._event_complete()` runs a regular expression over the entire buffer, this is an O(N^2) operation, which takes a long time when N is large.

**Separately, there is a copyright violation here.** I discovered that `_sseclient.py, despite having a Google copyright header, is largely copied from https://github.com/btubbs/sseclient/blob/master/sseclient.py.

#### Relevant Code:

Run the following code to reproduce:
```python
def listen_issue():
    path = '/sandbox/listen'

    print("Inserting data...")
    start = time.time()
    for i in range(10):
        long_data = ''.join(random.choice(string.ascii_uppercase) for _ in range(20000))
        db.reference(os.path.join(path, str(i))).set(long_data)
    end = time.time()
    print("...done (took %f seconds)" % (end - start))

    done_event = threading.Event()
    listener = db.reference(path).listen(lambda db_event: done_event.set())
    start = time.time()
    print("Waiting for first message from listen()...")
    done_event.wait()  # CPU usage is at 100% here
    end = time.time()
    print("...done (took %f seconds)" % (end - start))
    listener.close()
```

Example output on my machine:
```
Inserting data...
...done (took 1.678544 seconds)
Waiting for first message from listen()...
...done (took 109.180737 seconds)
```

hiranya911:
Improves the streaming RTDB API performance by checking for end of field more efficiently. Specifically, instead of matching a regex on each iteration, we keep track of the last 4 characters seen, and perform a simple string comparison.

Here are some preliminary test results to get an idea of the improvement.

| Node Size  | Time to process (Before) | Time to process (After) |
| ------------- | ------------- | ------------- |
| 100K  | 40.83s  | 1.73s |
| 1M  | - | 11.36s |

Fixes #198

### DIFF
--- a/firebase_admin/_sseclient.py
+++ b/firebase_admin/_sseclient.py
@@ -50,8 +50,7 @@ class _EventBuffer(object):
     def append(self, char):
         self._buffer.append(char)
         self._tail += char
-        if len(self._tail) > 4:
-            self._tail = self._tail[1:]
+        self._tail = self._tail[-4:]
 
     def truncate(self):
         head, sep, _ = self.buffer_string.rpartition('\n')
--- a/firebase_admin/_sseclient.py
+++ b/firebase_admin/_sseclient.py
@@ -177,7 +177,7 @@ class Event(object):
           raw: the raw data to parse.
 
         Returns:
-          Event: newly intialized ``Event`` object with the parameters initialized.
+          Event: A new ``Event`` with the parameters initialized.
         """
         event = cls()
         for line in raw.split('\n'):
--- a/firebase_admin/_sseclient.py
+++ b/firebase_admin/_sseclient.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""SSEClient module to stream realtime updates in the Firebase Database."""
+"""SSEClient module to stream realtime updates from the Firebase Database.
+
+Based on a similar implementation from Pyrebase.
+"""
 
 import re
 import time
@@ -58,7 +61,7 @@ class SSEClient(object):
         self.buf = u'' # Keep data here as it streams in
 
         headers = self.requests_kwargs.get('headers', {})
-        # The SSE spec requires making requests with Cache-Control: nocache
+        # The SSE spec requires making requests with Cache-Control: no-cache
         headers['Cache-Control'] = 'no-cache'
         # The 'Accept' header is not required, but explicit > implicit
         headers['Accept'] = 'text/event-stream'
@@ -77,7 +80,7 @@ class SSEClient(object):
             if self.last_id:
                 self.requests_kwargs['headers']['Last-Event-ID'] = self.last_id
             self.resp = self.session.get(self.url, stream=True, **self.requests_kwargs)
-            self.resp_iterator = self.resp.iter_content(decode_unicode=True)
+            self.resp_iterator = self.resp.iter_content(decode_unicode=True, chunk_size=None)
             self.resp.raise_for_status()
         else:
             raise StopIteration()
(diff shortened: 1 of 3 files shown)


daniel-ziegler:
Doesn't work for me at all on Mac OS X -- I simply don't receive any listen events. This definitely needs to be tested with a real HTTP stream.

Also note that while this should hopefully improve the constant factor, it's still O(N^2).

hiranya911:
There seems to be a `requests` bug here, which I've reported at https://github.com/requests/requests/issues/4876

The proposed solution essentially reduces N by reading data in chunks. With the fix I've tested payloads (in the unit tests) as large as 10MB, which can be processed under a second.

hiranya911:
Attached is another solution I've been tinkering with. It doesn't change how we read content from `requests` -- just changes how we handle the incoming data:

* Using a list to buffer incoming data as opposed to string concatenation which is inefficient at large sizes.
* Getting rid of the regex check on each iteration. The alternative check I've implemented assumes '\n\n` as the only acceptable terminator, which seems to be correct for Firebase.

With this I was able to read a 1MB node from a production database in under 10 seconds. Feedback welcome.

[sse.txt](https://github.com/firebase/firebase-admin-python/files/2591340/sse.txt)

daniel-ziegler:
I like that solution better since it is indeed linear time. Of course for maximum performance we'd need to figure out how to read in bigger chunks and not process each character individually in Python, but I'm happy with the improvement.

hiranya911:
@daniel-ziegler I've cleaned up the 2nd solution and committed it. Can you give it a try? My performance numbers are added to the PR description.

firebase_admin/_sseclient.py
```diff
@@ -37,6 +40,35 @@ def rebuild_auth(self, prepared_request, response):
...
+
+    def __init__(self):
+        self._buffer = []
+        self._tail = ''
+
+    def append(self, char):
+        self._buffer.append(char)
+        self._tail += char
+        if len(self._tail) > 4:
+            self._tail = self._tail[1:]
```
daniel-ziegler: You could also just do `self._tail = self._tail[-4:]` instead of the `if`.

daniel-ziegler:
Seems to work; obviously I haven't tested it very thoroughly.

firebase_admin/_sseclient.py
hiranya911: Done

firebase_admin/_sseclient.py
```diff
@@ -150,7 +177,7 @@ def parse(cls, raw):
           raw: the raw data to parse.
 
         Returns:
-          Event: newly intialized ``Event`` object with the parameters  initialized.
+          Event: newly intialized ``Event`` object with the parameters initialized.
```
bklimt: s/intialized/initialized/

hiranya911: Fixed the comment

PR #223, opened 2018-11-17
Complete implementation and unit tests

weixifan:
This should be the final PR for the implementation of the App-scoped features in the Project Management Service.
- Reworked the display name semantics to return `None` when it has never been set rather than the empty string.
- Add `__eq__` and `__hash__` methods to `AndroidAppMetadata` and `IosAppMetadata`; write the remaining unit tests.
- Remove `name` as a publicly visible property from `AndroidAppMetadata` and `IosAppMetadata`.


hiranya911:
I'll wait for #219 to be merged and this branch to be sync'ed.

weixifan:
#219 has been merged; please take a look.

weixifan:
I originally wanted to have the `ANDROID_APP_METADATA` constant as a `@fixture` as with the last PR, but unfortunately I wanted to share them between the Android and iOS tests, so they may as well be a module-level constant at that point.

weixifan:
The coverage is now at 100%.

tests/test_project_management.py
```diff
@@ -1065,6 +1082,12 @@ def test_delete_sha_certificates_not_found(self, android_app):
         assert excinfo.value.detail is not None
         assert len(recorder) == 1
 
+    def test_raises_if_app_has_no_project_id(self):
```
hiranya911: Can we use `testutils.run_without_project_id()` here as well?

PR #238, opened 2018-12-17
Implemented messaging.CriticalSound API

hiranya911:
go/apns-critical-sound

Implemented the `CriticalSound` API. Made empty strings disallowed in the sound config.

This made the `messaging` module slightly longer than `pylint` would like. I've silenced the lint error for now by adding a comment. I'll fix this property in a separate PR (didn't want to pile on here and make the review complicated).

### DIFF
--- a/firebase_admin/messaging.py
+++ b/firebase_admin/messaging.py
@@ -446,6 +446,7 @@ class ApsAlert(object):
     Args:
         title: Title of the alert (optional). If specified, overrides the title set via
             ``messaging.Notification``.
+        subtitle: Subtitle of the alert (optional).
         body: Body of the alert (optional). If specified, overrides the body set via
             ``messaging.Notification``.
         loc_key: Key of the body string in the app's string resources to use to localize the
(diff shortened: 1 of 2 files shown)


firebase_admin/messaging.py
```diff
@@ -754,6 +776,29 @@ def encode_aps(cls, aps):
...
+        if sound and isinstance(sound, six.string_types):
+            return sound
+        if not isinstance(sound, CriticalSound):
+            raise ValueError(
+                'Aps.sound must be a non-empty string or an instance of CriticalSound class.')
+        result = {
+            'name': _Validators.check_string('CriticalSound.name', sound.name, non_empty=True),
+            'volume': _Validators.check_number('CriticalSound.volume', sound.volume),
+        }
+        if sound.critical is True:
```
bklimt: Should we allow any truthy value and not just `True` specifically?

firebase_admin/messaging.py
```diff
@@ -418,6 +421,25 @@ def __init__(self, alert=None, badge=None, sound=None, content_available=None, c
...
+            folder of your app's container directory. Specify the string ``default`` to play the
+            system sound.
+        critical: Set to ``True`` to set the critical alert flag on the sound configuration
+            (optional).
+        volume: The volume for the critical alert's sound. Must be a value between 0.0 (silent)
+            and 1.0 (full volume) (optional).
+    """
+
+    def __init__(self, name, critical=None, volume=None):
+        self.name = name
```
bklimt: What's the rationale for doing input validation when the `ApsAlert` gets built instead of here when the values are first supplied?

hiranya911: This is to guard against something like this:

```
cs = CriticalSound('foo', volume=0.5)
cs.volume = -1
send_message(cs)
```

bklimt: Oh, python. :-)  Sounds reasonable.

firebase_admin/messaging.py
hiranya911: In the past we've been pretty strict with our type checking -- something I plan to relax as we go on. Made the change as suggested.

PR #203, opened 2018-10-05
Add support for "webpush.fcm_options" (WebpushFcmOptions)

RoberMac:
Resolved #202

Issue #202, opened 2018-10-05
FR: Add support for "webpush.fcm_options" (WebpushFcmOptions)


RoberMac:
As FCM HTTP API v1  and Firebase JS SDK already supports "[WebpushFcmOptions](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#WebpushFcmOptions)" and [click action handling](https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/src/controllers/sw-controller.ts#L189) on Web , I forked this repo and will create a PR for this issue.

hiranya911:
I'm on this one. Code being developed in a separate branch at the moment.

hiranya911:
This was included in the 2.15.0 release.

PR #242, opened 2019-01-02
FCM Options Support in WebpushConfig API

hiranya911:
Adding the `fcm_options` parameter to the `WebpushConfig` type. Change ported from https://github.com/firebase/firebase-admin-node/pull/373

Resolves #203


hiranya911:
Resolves #202

firebase_admin/_messaging_utils.py
```diff
@@ -232,6 +234,18 @@ def __init__(self, title=None, body=None, icon=None, actions=None, badge=None, d
         self.custom_data = custom_data
 
 
+class WebpushFcmOptions(object):
+    """Options for features provided by the FCM SDK for Web.
+
+    Args:
+        link: The link to open when the user clicks on the notification. Must be a HTTPS URL
```
bklimt: s/a HTTPS/an HTTPS/

firebase_admin/_messaging_utils.py
```diff
@@ -588,6 +603,20 @@ def encode_webpush_notification_actions(cls, actions):
...
+    def encode_webpush_fcm_options(cls, options):
+        """Encodes a WebpushFcmOptions instance into JSON."""
+        if options is None:
+            return None
+        result = {
+            'link': _Validators.check_string('WebpushConfig.fcm_options.link', options.link),
+        }
+        result = cls.remove_null_values(result)
+        link = result.get('link')
+        if link is not None and not link.startswith('https://'):
```
bklimt: Should we allow alternate capitalization on the protocol? I mean, "HTTPS://" is probably okay, right?

hiranya911: Yes, but we have always checked only for the lower-cased prefix in this codebase (there are similar URL checks in auth, db etc). So I'll keep this consistent with that. Also developers haven't  requested this, so it should be ok.

firebase_admin/_messaging_utils.py
hiranya911: Done

PR #247, opened 2019-01-12
Implementing HTTP retries for the SDK

### LINKED ISSUE #237: Automatically retry 503 errors for RTDB
RTDB occasionally will service [503][1] status codes if the service is temporarily overloaded - currently the HTTPClient doesn't retry those operations, which are safe to be retried.

[1]: https://httpstatuses.com/503

hiranya911:
Implementing HTTP retries for the SDK. This will:

1. Retry once for all low-level errors
2. Retry up to 4 times for HTTP 500 and 503 errors (with exponential backoff)
3. Respect the HTTP Retry-After header when available

Resolves #237

### DIFF
--- a/firebase_admin/_http_client.py
+++ b/firebase_admin/_http_client.py
@@ -19,15 +19,27 @@
 
 from google.auth import transport
 import requests
+from requests.packages.urllib3.util import retry # pylint: disable=import-error
+
+
+# Default retry configuration: Retries once on low-level connection and socket read errors.
+# Retries up to 4 times on HTTP 500 and 503 errors, with exponential backoff. Returns the
+# last response upon exhausting all retries.
+DEFAULT_RETRY_CONFIG = retry.Retry(
+    connect=1, read=1, status=4, status_forcelist=[500, 503],
+    raise_on_status=False, backoff_factor=0.5)
 
 
 class HttpClient(object):
     """Base HTTP client used to make HTTP calls.
 
-    HttpClient maintains an HTTP session, and handles request authentication if necessary.
+    HttpClient maintains an HTTP session, and handles request authentication and retries if
+    necessary.
     """
 
-    def __init__(self, credential=None, session=None, base_url='', headers=None):
+    def __init__(
+            self, credential=None, session=None, base_url='', headers=None,
+            retries=DEFAULT_RETRY_CONFIG):
         """Cretes a new HttpClient instance from the provided arguments.
 
         If a credential is provided, initializes a new HTTP session authorized with it. If neither
@@ -38,6 +50,9 @@ class HttpClient(object):
           session: A custom HTTP session (optional).
           base_url: A URL prefix to be added to all outgoing requests (optional).
           headers: A map of headers to be added to all outgoing requests (optional).
+          retries: A urllib retry configuration. Default settings would retry once for low-level
+              connection and socket read errors, and up to 4 times for HTTP 500 and 503 errors.
+              Pass a False value to disable retries (optional).
         """
         if credential:
             self._session = transport.requests.AuthorizedSession(credential)
@@ -48,6 +63,9 @@ class HttpClient(object):
 
         if headers:
             self._session.headers.update(headers)
+        if retries:
+            self._session.mount('http://', requests.adapters.HTTPAdapter(max_retries=retries))
+            self._session.mount('https://', requests.adapters.HTTPAdapter(max_retries=retries))
         self._base_url = base_url
 
     @property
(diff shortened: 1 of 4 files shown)


firebase_admin/_http_client.py
```diff
@@ -19,15 +19,27 @@
...
-    HttpClient maintains an HTTP session, and handles request authentication if necessary.
+    HttpClient maintains an HTTP session, and handles request authentication and retries if
+    necessary.
     """
 
-    def __init__(self, credential=None, session=None, base_url='', headers=None):
+    def __init__(
+            self, credential=None, session=None, base_url='', headers=None,
+            retries=DEFAULT_RETRY_CONFIG):
         """Cretes a new HttpClient instance from the provided arguments.
```
bklimt: s/Cretes/Creates/

hiranya911:
According to https://cloud.google.com/apis/design/errors#handling_errors clients should retry on 500 and 503 errors. I do understand that RTDB APIs are slightly different from other cloud REST APIs. But I suppose we can try this out in the wild, and if it becomes a problem use a different retry config just for the RTDB client (the API as implemented here supports it).

firebase_admin/_http_client.py
hiranya911: Done

rockwotj:
That sounds good to me @hiranya911

firebase_admin/_http_client.py
```diff
@@ -19,16 +19,28 @@
...
 import requests
+from requests.packages.urllib3.util import retry # pylint: disable=import-error
+
+
+# Default retry configuration: Retries once on low-level connection and socket read errors.
+# Retries up to 4 times on HTTP 500 and 503 errors, with exponential backoff. Returns the
+# last response upon exhausting all retries.
+DEFAULT_RETRY_CONFIG = retry.Retry(
+    connect=1, read=1, status=4, status_forcelist=[500, 503],
+    raise_on_status=False, backoff_factor=0.5)
```
beck-conalgo: I am not finding `raise_on_status` as an argument to Retry object defined in version `2.21.0` of requests module (most recent public version). Perhaps you are getting some other version? This is causing an error during import of `firebase_admin.admin` on my setup.

hiranya911: `retry.Retry` is a class defined in `urllib3`. It has supported `raise_on_status` since v1.15 (released in 2016): https://github.com/urllib3/urllib3/commit/9f48d26cd608b331b86e1e867f0edffee9560f94

You need to upgrade urllib3 to something more recent.

beck-conalgo: Your code reaches into a copy of `urllib3` that's bundled with `requests`. It's not possible to upgrade that independently, AFAIK. I am using the newest `requests` (2.21.0) and it contains this issue.

hiranya911: This has been developed and tested on latest `requests`. I double checked just now to make sure:

```
>>> import requests
>>> requests.__version__
'2.21.0'
>>> import firebase_admin
>>> firebase_admin.__version__
'2.15.1'
>>> from firebase_admin import db
>>>
>>> from requests.packages.urllib3.util import retry
>>> print(retry.Retry.__doc__)
 Retry configuration.

    Each retry attempt will create a new Retry object with updated values, so
    they can be safely reused.
...
    :param bool raise_on_status: Similar meaning to ``raise_on_redirect``:
        whether we should raise an exception, or return a response,
        if status falls in ``status_forcelist`` range and retries have
        been exhausted.
...
```

`requests` does not ship `urllib3` with it (as far as I can tell). It just re-exports whatever it can find already installed in the system: https://github.com/requests/requests/blob/v2.21.0/requests/packages.py

beck-conalgo: `requests` has `urllib3>=1.21.1,<1.25` as a requirement (in [setup.py](https://github.com/requests/requests/blob/c9309e4c59acfa757b34619205794e18f2009c43/setup.py#L47)), and that range of versions will do what we want. However, my build process is using pip to construct zipfile modules (because appengine limitations) and when pip is finished with `requests`, the packages subdirectory inside has an obsolete version of urllib3 captured into the zip. I've since learned that is coming from the virtualenv being used for building, even though I've added `urllib3` to my `requirements.txt`, and pip finds and downloads `urllib3` 1.24, but that is not what gets captured into the zipfile holding `requests`!

If your code used `from urllib3.util import retry` and specified `urllib3` as a dependancy, then all would be well. Since we might hope that `requests` will cleanup and remove the `packages` hack someday, maybe that would be a net improvement regardless?

Thanks for your help today.

hiranya911: That sounds like a reasonable change to make. Although I'm a little concerned about declaring a dependency on `urllib3`. We had to stop declaring `requests` as a dependency to work around a bug related to Firestore and pip. I wonder if adding `urllib3` to our setup file will cause issues. But I can certainly experiment with that option and see if it's possible.

Issue #248, opened 2019-01-12
ApiCallError - OSError("(104, 'ECONNRESET')")


kill4food:
Hi,
I'm a script on an amazon EC2 with operating system Ubuntu Server 18.04 LTS (HVM).
Firebase SDK version:  2.14.0
Python version: 3.7
Firebase product: database

once in 12-48 hours while i try to use a noraml .get() command on different parts of my code i get the following exception:
```
ApiCallError
('Connection aborted.', OSError("(104, 'ECONNRESET')"))
event:   File "/home/script.py", line 29, in script
    database.reference('data').order_by_child('parameters/name').get()
  File "/usr/local/lib/python3.7/site-packages/firebase_admin/db.py", line 612, in get
    result = self._client.body('get', self._pathurl, params=self._querystr)
  File "/usr/local/lib/python3.7/site-packages/firebase_admin/_http_client.py", line 93, in body
    resp = self.request(method, url, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/firebase_admin/db.py", line 884, in request
    raise ApiCallError(_Client.extract_error_message(error), error)
```
sometimes it happens with order_by_child get and sometimes on a normal get.

hiranya911:
A few low-level errors are to be expected when interacting with RTDB over a long period (due to backend server restarts, and other network issues). You can catch the error and retry when this happens. In the future, the SDK will also do some automatic retrying (see #247) when something lik this happens.

hiranya911:
@kill4food we do retry for 503 errors now (it was implemented in the same PR referenced above). We retry up to 4 times with exponential backoff before giving up. That adds up to a retry window of around 8 seconds. In the above case the error probably lasted longer than that. Keep us posted on how often you see this sort of errors. If 503's are frequent, we can look into further extending the retry window.

hiranya911:
Oh, actually it turns out `requests` does not retry some HTTP methods like PATCH by default. Probably because some of those methods are not safe to retry. I'll see if we can override that.

hiranya911:
These are the methods that are retried by default: https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.retry.Retry.DEFAULT_METHOD_WHITELIST

kill4food:
About the frequency of this issue, i got it twice in a 3 months period
But if it could be better if patch exceptions could be retried, since update() is a very common method

kill4food:
[...]

PR #255, opened 2019-01-30
Extending HTTP retries to all methods

hiranya911:
By default `urllib3` only retries certain HTTP methods: https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.retry.Retry.DEFAULT_METHOD_WHITELIST

Lot of our APIs use the POST method and some even use PATCH. Therefore I'm enabling retry support for all HTTP methods.

Related to #248

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
--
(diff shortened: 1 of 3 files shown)


hiranya911:
Related documentation from GCP (adding here for more context): https://cloud.google.com/apis/design/errors#error_retries

PR #263, opened 2019-02-12
Distribution package only including firebase_admin

fgmacedo:
*firebase-admin* is polluting the Python's default package namespace with a *integration* package.

We already have an internal app using *integration* namespace, so the names are in conflict.

I think that *integration* should not be included on the distribution package, since It's related to integration tests only.

This PR changes the pattern for discovering packages to `firebase*`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@ from __future__ import print_function
 from os import path
 import sys
 
-from setuptools import find_packages
 from setuptools import setup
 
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -55,7 +55,7 @@
     license=about['__license__'],
     keywords='firebase cloud development',
     install_requires=install_requires,
-    packages=find_packages(exclude=['tests']),
+    packages=find_packages(include=['firebase*']),
```
hiranya911: Thanks for pointing this out @fgmacedo. Can we drop the wildcard and explicitly make it `['firebase_admin']`?  We don't want anything else exposed from our binaries.

fgmacedo:
@hiranya911 thanks for merging!  Do you have any plans for a new release coming?  We are unable to install the project. Even our fork is not working, because a `pip install -e <fork-url>` clones the project with a high priority path on the `sys.modules` that includes the `integration` module, so the conflict persists. We need a new version on the pypi to be able to install as a source dist.

hiranya911:
@fgmacedo will try to get something out next week.

fgmacedo:
Thanks @hiranya911! :heart:

PR #264, opened 2019-02-19
Added snippets for email action links API

hiranya911:
Snippets to appear on https://firebase.google.com/docs/auth/admin/email-action-links

Also fixing a regression caused by upstream libraries: https://github.com/googleapis/google-auth-library-python/pull/324


snippets/auth/index.py
```diff
@@ -587,6 +587,58 @@ def import_without_password():
...
+    # [END init_action_code_settings]
+    return action_code_settings
+
+def password_reset_link():
+    action_code_settings = init_action_code_settings()
+    display_name = 'Example User'
+    # [START password_reset_link]
+    email = 'user@example.com'
+    link = auth.generate_password_reset_link(email, action_code_settings)
+    # Construct password reset email template, embed the link and send
```
bklimt: This wording is a bit awkward. Maybe something like this would be clearer?

    # Construct password reset email from a template embedding the link, and send
    # using a custom SMTP server.

snippets/auth/index.py
```diff
@@ -587,6 +587,58 @@ def import_without_password():
...
+        android_install_app=True,
+        android_minimum_version='12',
+        dynamic_link_domain='coolapp.page.link',
+    )
+    # [END init_action_code_settings]
+    return action_code_settings
+
+def password_reset_link():
+    action_code_settings = init_action_code_settings()
+    display_name = 'Example User'
```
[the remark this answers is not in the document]
hiranya911: Removed display_name

snippets/auth/index.py
hiranya911: Done

Issue #277, opened 2019-03-28
FR: Support for sending notification to multiple devices


adihat:
Currently there is support to send notifications to specific devices as documented [here][1].

There is support for sending notifications to multiple devices in Node JS SDK and Java SDK.

Example:
```javascript
const registrationTokens = [
  'YOUR_REGISTRATION_TOKEN_1',
  // …
  'YOUR_REGISTRATION_TOKEN_N',
];

const message = {
  data: {score: '850', time: '2:45'},
  tokens: registrationTokens,
}

admin.messaging().sendMulticast(message)
  .then((response) => {
    console.log(response.successCount + ' messages were sent successfully');
```

This operation uses the `sendAll()` API under the hood and return value is a `BatchResponse`.

It would be good if there was something similar in python sdk:
```python
registration_tokens = [
       'YOUR_REGISTRATION_TOKEN_1',
       // …
      'YOUR_REGISTRATION_TOKEN_N',
]
message = messaging.Message(
    data={
        'score': '850',
        'time': '2:45',
    },
    tokens=registration_tokens,
)
response = messaging.send_multicast(message)
print(f"{response['successCount']} messages were sent successfully")
```

[1]: https://firebase.google.com/docs/cloud-messaging/send-message

aaugustin:
Unless I missed something, this issue was fixed by #283 and can be closed.

It would be nice to have a PyPI release with this feature :-) Thank you :-)

hiranya911:
Release on its way.

Issue #196, opened 2018-08-23
Send message to multiple devices


KoblStone:
I want to send notifications to specific users using tokens. firebase-admin can't send multiple tokens so I'm using:

```
for obj in queryset:
            token = obj.app_token
            message = messaging.Message(android=android_config, data=data, token=token)
            try:
                messaging.send(message=message, dry_run=False, app=default_app)
            except Exception as ex:
                print('Error %s' % ex)
```

But it seems that "FCM" is blocked due to the fact that I send a lot of requests per second. Because after that I can not get a notification for a while.
Can this be solved?

hiranya911:
I believe there's some on going work to support multicast messaging in the new FCM REST API. This will be supported once that is implemented (although I don't know what the exact timeline is).

hiranya911:
In the meantime, can you use topic messaging to implement your use case?

KoblStone:
No. The list is generated dynamically and I will need to create multiple topics

chong-shao:
Update: as mentioned in #315, we are working on a new batch-send API for Admin SDK.

chong-shao:
Further updates will be posted on https://github.com/firebase/firebase-admin-node/issues/315

aaugustin:
I believe this issue was fixed by #283 and can be closed.

chong-shao:
Thanks for the reminder!

PR #283, opened 2019-05-01
Add messaging send_all and send_multicast functions

ZachOrr:
Hey y'all - I'm working on adding the `sendAll` and `sendMulticast` features for #196 based on https://github.com/firebase/firebase-admin-node/pull/453 - I was hoping I could get some assistance with how to represent returns from these functions, since the current `send` endpoint will raise if the response from FCM contains an error, and this doesn't seem like a great solution when sending several requests.

Also, if I could get suggestions on a non-regex way to wrangle batch response content, that would be appreciated. My first thought was seeing if we could dump each individual response content in to a requests Response object, but that doesn't appear to be a thing with Response objects.

### DIFF
--- a/firebase_admin/_messaging_utils.py
+++ b/firebase_admin/_messaging_utils.py
@@ -54,6 +54,33 @@ class Message(object):
         self.condition = condition
 
 
+class MulticastMessage(object):
+    """A message that can be sent to multiple tokens via Firebase Cloud Messaging.
+
+    Contains payload information as well as recipient information. In particular, the message must
+    contain exactly one of token, topic or condition fields.
+
+    Args:
+        tokens: A list of registration token of the device to which the message should be sent.
+        data: A dictionary of data fields (optional). All keys and values in the dictionary must be
+            strings.
+        notification: An instance of ``messaging.Notification`` (optional).
+        android: An instance of ``messaging.AndroidConfig`` (optional).
+        webpush: An instance of ``messaging.WebpushConfig`` (optional).
+        apns: An instance of ``messaging.ApnsConfig`` (optional).
+    """
+    def __init__(self, tokens, data=None, notification=None, android=None, webpush=None, apns=None):
+        _Validators.check_string_list('MulticastMessage.tokens', tokens)
+        if len(tokens) > 100:
+            raise ValueError('MulticastMessage.tokens must contain less than 100 tokens.')
+        self.tokens = tokens
+        self.data = data
+        self.notification = notification
+        self.android = android
+        self.webpush = webpush
+        self.apns = apns
+
+
 class Notification(object):
     """A notification that can be included in a message.
 
@@ -150,7 +177,7 @@ class WebpushConfig(object):
         data: A dictionary of data fields (optional). All keys and values in the dictionary must be
             strings. When specified, overrides any data fields set via ``Message.data``.
         notification: A ``messaging.WebpushNotification`` to be included in the message (optional).
-        fcm_options: A ``messaging.WebpushFcmOptions`` instance to be included in the messsage
+        fcm_options: A ``messaging.WebpushFcmOptions`` instance to be included in the message
             (optional).
 
     .. _Webpush Specification: https://tools.ietf.org/html/rfc8030#section-5
--- a/firebase_admin/_messaging_utils.py
+++ b/firebase_admin/_messaging_utils.py
@@ -72,7 +72,7 @@ class MulticastMessage(object):
     def __init__(self, tokens, data=None, notification=None, android=None, webpush=None, apns=None):
         _Validators.check_string_list('MulticastMessage.tokens', tokens)
         if len(tokens) > 100:
-            raise ValueError('MulticastMessage.tokens must contain less than 100 tokens.')
+            raise ValueError('MulticastMessage.tokens must not contain more than 100 tokens.')
         self.tokens = tokens
         self.data = data
         self.notification = notification
--- a/firebase_admin/messaging.py
+++ b/firebase_admin/messaging.py
@@ -100,7 +100,7 @@ def send(message, dry_run=False, app=None):
     return _get_messaging_service(app).send(message, dry_run)
 
 def send_all(messages, dry_run=False, app=None):
-    """Batch sends the given messages via Firebase Cloud Messaging (FCM).
+    """Sends the given list of messages via Firebase Cloud Messaging as a single batch.
 
     If the ``dry_run`` mode is enabled, the message will not be actually delivered to the
     recipients. Instead FCM performs all the usual validations, and emulates the send operation.
@@ -258,10 +258,7 @@ class BatchResponse(object):
 
     def __init__(self, responses):
         self._responses = responses
-        self._success_count = 0
-        for response in responses:
-            if response.success:
-                self._success_count += 1
+        self._success_count = len([resp for resp in responses if resp.success])
 
     @property
     def responses(self):
@@ -455,7 +452,7 @@ class _MessagingService(object):
 
     def _postproc(self, resp, body):
         if resp.status == 200:
-            return json.loads(body)
+            return json.loads(body.decode())
         else:
             raise Exception('unexpected response')
 
@@ -498,7 +495,7 @@ class _MessagingService(object):
 
         data = {}
         try:
-            parsed_body = json.loads(error.content)
+            parsed_body = json.loads(error.content.decode())
             if isinstance(parsed_body, dict):
                 data = parsed_body
         except ValueError:
--- a/firebase_admin/messaging.py
+++ b/firebase_admin/messaging.py
@@ -120,13 +120,13 @@ def send_all(messages, dry_run=False, app=None):
     return _get_messaging_service(app).send_all(messages, dry_run)
 
 def send_multicast(multicast_message, dry_run=False, app=None):
-    """Sends the given mutlicast message to the mutlicast message tokens via Firebase Cloud Messaging (FCM).
+    """Sends the given mutlicast message to all tokens via Firebase Cloud Messaging (FCM).
 
     If the ``dry_run`` mode is enabled, the message will not be actually delivered to the
     recipients. Instead FCM performs all the usual validations, and emulates the send operation.
 
     Args:
-        message: An instance of ``messaging.MulticastMessage``.
+        multicast_message: An instance of ``messaging.MulticastMessage``.
         dry_run: A boolean indicating whether to run the operation in dry run mode (optional).
         app: An App instance (optional).
 
@@ -139,14 +139,14 @@ def send_multicast(multicast_message, dry_run=False, app=None):
     """
     if not isinstance(multicast_message, MulticastMessage):
         raise ValueError('Message must be an instance of messaging.MulticastMessage class.')
-    messages = map(lambda token: Message(
+    messages = [Message(
         data=multicast_message.data,
         notification=multicast_message.notification,
         android=multicast_message.android,
         webpush=multicast_message.webpush,
         apns=multicast_message.apns,
         token=token
-    ), multicast_message.tokens)
+    ) for token in multicast_message.tokens]
     return _get_messaging_service(app).send_all(messages, dry_run)
 
 def subscribe_to_topic(tokens, topic, app=None):
@@ -254,6 +254,7 @@ class ApiCallError(Exception):
 
 
 class BatchResponse(object):
+    """The response received from a batch request to the FCM API."""
 
     def __init__(self, responses):
         self._responses = responses
@@ -277,6 +278,7 @@ class BatchResponse(object):
 
 
 class SendResponse(object):
+    """The response received from an individual batched request to the FCM API."""
 
     def __init__(self, resp, exception):
         self._exception = exception
@@ -361,7 +363,12 @@ class _MessagingService(object):
         data = self._message_data(message, dry_run)
         try:
             resp = self._client.body(
-                'post', url=self._fcm_url, headers=self._fcm_headers, json=data, timeout=self._timeout)
+                'post',
+                url=self._fcm_url,
+                headers=self._fcm_headers,
+                json=data,
+                timeout=self._timeout
+            )
         except requests.exceptions.RequestException as error:
             if error.response is not None:
                 self._handle_fcm_error(error)
@@ -372,12 +379,13 @@ class _MessagingService(object):
             return resp['name']
 
     def send_all(self, messages, dry_run=False):
+        """Sends the given messages to FCM via the batch API."""
         if not isinstance(messages, list):
             raise ValueError('Messages must be an list of messaging.Message instances.')
 
         responses = []
 
-        def batch_callback(request_id, response, error):
+        def batch_callback(_, response, error):
             exception = None
             if error:
                 exception = self._parse_batch_error(error)
@@ -388,7 +396,13 @@ class _MessagingService(object):
         for message in messages:
             body = json.dumps(self._message_data(message, dry_run))
             req = http.HttpRequest(
-                http=self._transport, postproc=self._postproc, uri=self._fcm_url, method='POST', body=body, headers=self._fcm_headers)
+                http=self._transport,
+                postproc=self._postproc,
+                uri=self._fcm_url,
+                method='POST',
+                body=body,
+                headers=self._fcm_headers
+            )
             batch.add(req)
 
         try:
@@ -455,7 +469,8 @@ class _MessagingService(object):
         except ValueError:
             pass
 
-        raise _MessagingService._parse_fcm_error(data, error.response.content, error.response.status_code, error)
+        raise _MessagingService._parse_fcm_error(
+            data, error.response.content, error.response.status_code, error)
 
     def _handle_iid_error(self, error):
         """Handles errors received from the Instance ID API."""
@@ -476,6 +491,7 @@ class _MessagingService(object):
         raise ApiCallError(code, msg, error)
 
     def _parse_batch_error(self, error):
+        """Parses a googleapiclient.http.HttpError content in to an ApiCallError."""
         if error.content is None:
             msg = 'Failed to call messaging API: {0}'.format(error)
             return ApiCallError(self.INTERNAL_ERROR, msg, error)
@@ -504,6 +520,7 @@ class _MessagingService(object):
 
         msg = error_dict.get('message')
         if not msg:
-            msg = 'Unexpected HTTP response with status: {0}; body: {1}'.format(status_code, content.decode())
+            msg = 'Unexpected HTTP response with status: {0}; body: {1}'.format(
+                status_code, content.decode())
 
         return ApiCallError(code, msg, error)
--- a/firebase_admin/messaging.py
+++ b/firebase_admin/messaging.py
@@ -453,10 +453,17 @@ class _MessagingService(object):
         return data
 
     def _postproc(self, resp, body):
-        if resp.status == 200:
-            return json.loads(body.decode())
-        else:
-            raise Exception('unexpected response')
+        if resp.status is not 200:
+            data = {}
+            try:
+                parsed_body = json.loads(body.decode())
+                if isinstance(parsed_body, dict):
+                    data = parsed_body
+            except ValueError:
+                pass
+            code, msg = _MessagingService._parse_fcm_error(data, body, resp.status)
+            raise ApiCallError(code, msg)
+        return json.loads(body.decode())
 
     def _handle_fcm_error(self, error):
         """Handles errors received from the FCM API."""
@@ -468,8 +475,9 @@ class _MessagingService(object):
         except ValueError:
             pass
 
-        raise _MessagingService._parse_fcm_error(
-            data, error.response.content, error.response.status_code, error)
+        code, msg = _MessagingService._parse_fcm_error(
+            data, error.response.content, error.response.status_code)
+        raise ApiCallError(code, msg, error)
 
     def _handle_iid_error(self, error):
         """Handles errors received from the Instance ID API."""
@@ -502,10 +510,12 @@ class _MessagingService(object):
                 data = parsed_body
         except ValueError:
             pass
-        return _MessagingService._parse_fcm_error(data, error.content, error.resp.status, error)
+
+        code, msg = _MessagingService._parse_fcm_error(data, error.content, error.resp.status)
+        return ApiCallError(code, msg, error)
 
     @classmethod
-    def _parse_fcm_error(cls, data, content, status_code, error):
+    def _parse_fcm_error(cls, data, content, status_code):
         """Parses an error response from the FCM API to a ApiCallError."""
         error_dict = data.get('error', {})
         server_code = None
@@ -521,5 +531,4 @@ class _MessagingService(object):
         if not msg:
             msg = 'Unexpected HTTP response with status: {0}; body: {1}'.format(
                 status_code, content.decode())
-
-        return ApiCallError(code, msg, error)
+        return code, msg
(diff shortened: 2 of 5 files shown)


hiranya911:
Thanks @ZachOrr  for working on this. We actually have an approved API design for this. It goes something like this:

```
firebase_admin.messaging
  send_all(messages, dry_run=False, app=None): BatchResponse
  send_multicast(multicast_message, dry_run=False, app=None): BatchResponse

  class MulticastMessage
    tokens: list<str>
    Everything else on Message class except token, topic and condition

  class BatchResponse
    responses: list<SendResponse>
    success_count: int
    failure_count: int

  class SendResponse
    success: boolean
    message_id: str
    exception: ApiCallError
```

You might want to look into using the Google API client for making the batch requests: https://developers.google.com/api-client-library/python/guide/batch

firebase_admin/messaging.py
```diff
@@ -233,10 +353,14 @@ def __init__(self, app):
                 'Project ID is required to access Cloud Messaging service. Either set the '
                 'projectId option, or use service account credentials. Alternatively, set the '
                 'GOOGLE_CLOUD_PROJECT environment variable.')
-        self._fcm_url = _MessagingService.FCM_URL.format(project_id)
+        self._fcm_service = build('fcm', 'v1', credentials=app.credential.get_credential())
```
ZachOrr: I can't seem to get this to auth by using `credentials`. I've only been able to get it to work with a `developerKey`. I've yet to poke around the `google-api-python-client` library to see what exactly is happening. Any ideas?

firebase_admin/_messaging_utils.py
```diff
@@ -54,6 +54,26 @@ def __init__(self, data=None, notification=None, android=None, webpush=None, apn
...
+    Args:
+        tokens: A list of registration token of the device to which the message should be sent (optional).
+        data: A dictionary of data fields (optional). All keys and values in the dictionary must be
+            strings.
+        notification: An instance of ``messaging.Notification`` (optional).
+        android: An instance of ``messaging.AndroidConfig`` (optional).
+        webpush: An instance of ``messaging.WebpushConfig`` (optional).
+        apns: An instance of ``messaging.ApnsConfig`` (optional).
+    """
+    def __init__(self, tokens=[], data=None, notification=None, android=None, webpush=None, apns=None):
```
ZachOrr: Should `tokens` have a default value, or some validation?

firebase_admin/messaging.py
```diff
@@ -192,10 +250,72 @@ def __init__(self, code, message, detail=None):
         self.detail = detail
 
 
+class BatchResponse(object):
+
+    def __init__(self, responses):
+        if not isinstance(responses, list):
```
ZachOrr: I'm unsure how much validation is the right level of validation. Things seem different in different places in the project. Let me know and I can add/remove as necessary.

firebase_admin/messaging.py
```diff
@@ -245,25 +359,33 @@ def encode_message(cls, message):
...
+            responses.append(send_response)
+            if len(responses) == message_count:
+                send_all_complete.set()
+
+        batch = self._fcm_service.new_batch_http_request(callback=send_all_callback)
+        for message in messages:
+            batch.add(self._message_request(message, dry_run))
+        batch.execute()
+
+        send_all_complete.wait()
```
ZachOrr: Let me know if you have a suggestion for a better way to aggregate responses from this callback and return synchronously - this seemed like an okay approach to me, but I'm open to suggestions.

firebase_admin/messaging.py
```diff
@@ -245,25 +359,33 @@ def encode_message(cls, message):
...
-            }
-            resp = self._client.body(
-                'post', url=self._fcm_url, headers=headers, json=data, timeout=self._timeout)
-        except requests.exceptions.RequestException as error:
-            if error.response is not None:
-                self._handle_fcm_error(error)
-            else:
-                msg = 'Failed to call messaging API: {0}'.format(error)
-                raise ApiCallError(self.INTERNAL_ERROR, msg, error)
+            resp = request.execute(num_retries=4)
```
ZachOrr: To mirror the retry behavior in the existing `HttpClient` code

firebase_admin/messaging.py
```diff
@@ -245,25 +359,33 @@ def encode_message(cls, message):
...
+        def send_all_callback(request_id, response, exception):
+            send_response = SendResponse(response, exception)
+            responses.append(send_response)
+            if len(responses) == message_count:
+                send_all_complete.set()
+
+        batch = self._fcm_service.new_batch_http_request(callback=send_all_callback)
+        for message in messages:
+            batch.add(self._message_request(message, dry_run))
+        batch.execute()
```
ZachOrr: Batches don't seem to be able to be retried when executing - are we okay with that?

firebase_admin/messaging.py
ZachOrr: The error I'm getting is `googleapiclient.errors.HttpError: <HttpError 403 when requesting https://fcm.googleapis.com/$discovery/rest?version=v1 returned "The request is missing a valid API key.">`

hiranya911: I played around with the googleapiclient a bit. I'll post my findings in a separate reply.

hiranya911:
Hi @ZachOrr. From what I can tell `googleapiclient` and `googe.auth` do not play well with each other in Python. This is unfortunate, as they really work well in other languages like Java and C#. In any case, I managed to send a batch request using the following code:

```
import json

import googleapiclient
from googleapiclient import http
from googleapiclient import _auth
import google.auth

# Init Google Credentials (this part is currently handled by firebase_admin.credentials)
_scopes = [
    'https://www.googleapis.com/auth/cloud-platform',
    'https://www.googleapis.com/auth/datastore',
    'https://www.googleapis.com/auth/devstorage.read_write',
    'https://www.googleapis.com/auth/firebase',
    'https://www.googleapis.com/auth/identitytoolkit',
    'https://www.googleapis.com/auth/userinfo.email'
]
creds, _ = google.auth.default(scopes=_scopes)

# It sucks that we have to use this internal _auth module. Wish there was a better way
transport = _auth.authorized_http(creds)

responses = []

def callback(req_id, resp, exception):
  print(req_id, resp, exception)
  responses.append(resp['name'])

def postproc(resp, body):
  """This gets called for each part. It's a low-level function where we can parse the HTTP response."""
  if resp.status == 200:
    return json.loads(body)
  else:
    raise Exception('unexpected response')

batch = http.BatchHttpRequest(callback, 'https://fcm.googleapis.com/batch')

body = json.dumps({'message': {'topic': 'foo'}})
req = http.HttpRequest(
  http=transport, postproc=postproc,
  uri='https://fcm.googleapis.com/v1/projects/my-project-id/messages:send',
  method='POST',
  body=body)

# Queue 3 requests in the batch
batch.add(req)
batch.add(req)
batch.add(req)

batch.execute()
print(responses) # Prints 3 message IDs
```

ZachOrr:
Awesome! Thanks for the help. It seems like using `app.credential.get_credential()` in `_auth.authorized_http` works fine - can we do that?

hiranya911:
@ZachOrr yes, lets do that for now. I'll see if we can find a better alternative for that part.

ZachOrr:
Added tests and added errors raises from batch request http errors.

ZachOrr:
Not entirely sure what the fix is to get that `json.loads` working on some versions of Python. Any suggestions?

firebase_admin/_messaging_utils.py
```diff
@@ -54,6 +54,33 @@ def __init__(self, data=None, notification=None, android=None, webpush=None, apn
...
+            strings.
+        notification: An instance of ``messaging.Notification`` (optional).
+        android: An instance of ``messaging.AndroidConfig`` (optional).
+        webpush: An instance of ``messaging.WebpushConfig`` (optional).
+        apns: An instance of ``messaging.ApnsConfig`` (optional).
+    """
+    def __init__(self, tokens, data=None, notification=None, android=None, webpush=None, apns=None):
+        _Validators.check_string_list('MulticastMessage.tokens', tokens)
+        if len(tokens) > 100:
+            raise ValueError('MulticastMessage.tokens must contain less than 100 tokens.')
```
hiranya911: Lets change this to `... must not contain more than 100 tokens.`

firebase_admin/_messaging_utils.py
```diff
@@ -150,7 +177,7 @@ class WebpushConfig(object):
         data: A dictionary of data fields (optional). All keys and values in the dictionary must be
             strings. When specified, overrides any data fields set via ``Message.data``.
         notification: A ``messaging.WebpushNotification`` to be included in the message (optional).
-        fcm_options: A ``messaging.WebpushFcmOptions`` instance to be included in the messsage
+        fcm_options: A ``messaging.WebpushFcmOptions`` instance to be included in the message
```
hiranya911: Good catch 👍

firebase_admin/messaging.py
```diff
@@ -88,6 +99,56 @@ def send(message, dry_run=False, app=None):
     """
     return _get_messaging_service(app).send(message, dry_run)
 
+def send_all(messages, dry_run=False, app=None):
+    """Batch sends the given messages via Firebase Cloud Messaging (FCM).
```
hiranya911: Lets change to `Sends the given list of messages via Firebase Cloud Messaging as a single batch.`

firebase_admin/messaging.py
```diff
@@ -192,10 +253,60 @@ def __init__(self, code, message, detail=None):
         self.detail = detail
 
 
+class BatchResponse(object):
+    """The response received from a batch request to the FCM API."""
+
+    def __init__(self, responses):
+        self._responses = responses
+        self._success_count = 0
```
hiranya911: `self._success_count = len(filter(lambda resp: resp.success, responses))`

firebase_admin/messaging.py
```diff
@@ -299,6 +447,18 @@ def make_topic_management_request(self, tokens, topic, operation):
...
+        data = {'message': _MessagingService.encode_message(message)}
+        if dry_run:
+            data['validate_only'] = True
+        return data
+
+    def _postproc(self, resp, body):
+        if resp.status == 200:
+            return json.loads(body)
+        else:
+            raise Exception('unexpected response')
```
hiranya911: I used this in my hack. But we need to raise something more detailed here. I think we need to parse the error response body, and construct a `messaging.ApiCallError`.

tests/test_messaging.py
```diff
@@ -38,6 +40,23 @@ def check_encoding(msg, expected=None):
         assert encoded == expected
 
 
+class TestMulticastMessage(object):
+
+    @pytest.mark.parametrize('tokens', NON_LIST_ARGS)
```
hiranya911: Lets also add a test case where we check 100+ tokens are not allowed.

tests/test_messaging.py
```diff
@@ -1316,6 +1335,384 @@ def test_send_fcm_error_code(self, status):
...
+
+    def test_no_project_id(self):
+        def evaluate():
+            app = firebase_admin.initialize_app(testutils.MockCredential(), name='no_project_id')
+            with pytest.raises(ValueError):
+                messaging.send_all([messaging.Message(topic='foo')], app=app)
+        testutils.run_without_project_id(evaluate)
+
+    @pytest.mark.parametrize('msg', NON_LIST_ARGS)
+    def test_invalid_send_all(self, msg):
```
hiranya911: Another test case for 100+ messages

tests/test_messaging.py
```diff
@@ -1316,6 +1335,384 @@ def test_send_fcm_error_code(self, status):
...
+        })
+        _ = self._instrument_batch_messaging_service(status=status, payload=payload)
+        msg = messaging.Message(topic='foo')
+        with pytest.raises(messaging.ApiCallError) as excinfo:
+            messaging.send_all([msg])
+        assert str(excinfo.value) == 'test error'
+        assert str(excinfo.value.code) == 'registration-token-not-registered'
+
+    @pytest.mark.parametrize('status', HTTP_ERRORS)
+    def test_send_all_batch_fcm_error_code(self, status):
```
hiranya911: I think this is a good start. But I'd recommend making a couple of changes:
1. Change all tests to have at least 2 messages in a batch.
2. Include some tests to check what happens when some messages in a batch are delivered successfully while some fail.

firebase_admin/messaging.py
ZachOrr: Is there a situation where FCM will return a non-200 or non-error response code? I'm trying to figure out what to do here. A `messaging.ApiCallError` takes an error, which we really don't have in this case.

tests/test_messaging.py
ZachOrr: Should we add validation here in `send_all` or fallback on the API for validation?

ZachOrr: I added validation in `send_all` for now, since we already have validation making sure `messages` is a list. We can remove though if falling back to the API limits is better.

hiranya911:
Resolves #277

tests/test_messaging.py
hiranya911: Validating here is the right thing to do. We can save the user an RPC call by catching it here.

firebase_admin/messaging.py
hiranya911: Technically it can be any HTTP response. But in practice FCM mostly returns 200 or 4xx/5xx error codes. We can treat anything that is not 200 as an error. 

We need to somehow call the error parsing logic in `_handle_fcm_error()` on this, and get an `ApiCallError` out of it. May be refactor the parsing logic into a new helper function:

```
if resp.status != 200:
  code, message = _parse_error_response(response)
  raise ApiCallError(code, message) # last argument is optional  
```

ZachOrr:
Alright - I think this is ready for a re-review.

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        assert all([r.success for r in batch_response.responses])
+        assert not any([r.exception for r in batch_response.responses])
+
+    def test_send_all_non_error_non_200(self):
+        success_payload = json.dumps({'name': 'message-id'})
+        error_payload = json.dumps({})
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
+        msg = messaging.Message(topic='foo')
+        with pytest.raises(messaging.ApiCallError) as excinfo:
```
hiranya911: This shouldn't raise an error. It should return a BatchResponse with success_count = 1 and failure_count = 1.

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        error_payload = json.dumps({
+            'error': {
+                'status': 'INVALID_ARGUMENT',
+                'message': 'test error'
+            }
+        })
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
+        msg = messaging.Message(topic='foo')
+        with pytest.raises(messaging.ApiCallError) as excinfo:
```
hiranya911: Shouldn't raise.

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        error_payload = json.dumps({
+            'error': {
+                'status': 'NOT_FOUND',
+                'message': 'test error'
+            }
+        })
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
+        msg = messaging.Message(topic='foo')
+        with pytest.raises(messaging.ApiCallError) as excinfo:
```
hiranya911: Shouldn't raise.

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+                        '@type': 'type.googleapis.com/google.firebase.fcm.v1.FcmError',
+                        'errorCode': 'UNREGISTERED',
+                    },
+                ],
+            }
+        })
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
+        msg = messaging.Message(topic='foo')
+        with pytest.raises(messaging.ApiCallError) as excinfo:
```
hiranya911: Shouldn't raise

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        error_payload = json.dumps({
+            'error': {
+                'status': 'INVALID_ARGUMENT',
+                'message': 'test error'
+            }
+        })
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (status, error_payload)]))
+        msg = messaging.Message(topic='foo')
+        batch_response = messaging.send_all([msg, msg])
```
hiranya911: Now I'm confused. Why wouldn't this throw? It is the right behavior, but how is this any different from the test cases above?

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        assert len(batch_response.responses) == 2
+        assert [r.message_id for r in batch_response.responses] == ['message-id', 'message-id']
+        assert all([r.success for r in batch_response.responses])
+        assert not any([r.exception for r in batch_response.responses])
+
+    def test_send_all_non_error_non_200(self):
+        success_payload = json.dumps({'name': 'message-id'})
+        error_payload = json.dumps({})
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
```
hiranya911: Lets use a bit more realistic error code than 202. That's not something you will ever get from FCM. I suppose 500 would be the easiest.

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (status, error_payload)]))
+        msg = messaging.Message(topic='foo')
+        batch_response = messaging.send_all([msg, msg])
+        assert batch_response.success_count is 1
+        assert batch_response.failure_count is 1
+        assert len(batch_response.responses) == 2
+        success_response = batch_response.responses[0]
+        assert success_response.message_id == 'message-id'
+        assert success_response.success
```
hiranya911: `assert success_response.success is True`

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        assert success_response.exception is None
+        error_response = batch_response.responses[1]
+        assert error_response.message_id is None
+        assert not error_response.success
```
hiranya911: `assert  error_response.success is False`

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        assert all([r.success for r in batch_response.responses])
+        assert not any([r.exception for r in batch_response.responses])
+
+    def test_send_multicast_non_error_non_200(self):
+        success_payload = json.dumps({'name': 'message-id'})
+        error_payload = json.dumps({})
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
+        msg = messaging.MulticastMessage(tokens=['foo', 'foo'])
+        with pytest.raises(messaging.ApiCallError) as excinfo:
```
hiranya911: Should not raise

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        error_payload = json.dumps({
+            'error': {
+                'status': 'NOT_FOUND',
+                'message': 'test error'
+            }
+        })
+        _ = self._instrument_batch_messaging_service(
+            payload=self._batch_payload([(200, success_payload), (202, error_payload)]))
+        msg = messaging.MulticastMessage(tokens=['foo', 'foo'])
+        with pytest.raises(messaging.ApiCallError) as excinfo:
```
hiranya911: Should not raise

tests/test_messaging.py
```diff
@@ -1316,6 +1342,546 @@ def test_send_fcm_error_code(self, status):
...
+        assert batch_response.failure_count is 1
+        assert len(batch_response.responses) == 2
+        success_response = batch_response.responses[0]
+        assert success_response.message_id == 'message-id'
+        assert success_response.success
+        assert success_response.exception is None
+        error_response = batch_response.responses[1]
+        assert error_response.message_id is None
+        assert not error_response.success
+        assert error_response.exception
```
hiranya911: `assert error_response.exception is not None`

firebase_admin/messaging.py
```diff
@@ -264,6 +375,42 @@ def send(self, message, dry_run=False):
...
+        if not isinstance(messages, list):
+            raise ValueError('Messages must be an list of messaging.Message instances.')
+        if len(messages) > 100:
+            raise ValueError('send_all messages must not contain more than 100 messages.')
+
+        responses = []
+
+        def batch_callback(_, response, error):
+            exception = None
+            if error:
```
hiranya911: This doesn't look right. I'd except `error` here to be the `ApiCallError` raised in `postproc`.

firebase_admin/messaging.py
```diff
@@ -299,6 +446,26 @@ def make_topic_management_request(self, tokens, topic, operation):
...
+        if resp.status is not 200:
+            data = {}
+            try:
+                parsed_body = json.loads(body.decode())
+                if isinstance(parsed_body, dict):
+                    data = parsed_body
+            except ValueError:
+                pass
+            code, msg = _MessagingService._parse_fcm_error(data, body, resp.status)
+            raise ApiCallError(code, msg)
```
hiranya911: I think I understand why this is not working properly. I found the following code in the `googleapiclient`:

```
     try:
        if resp.status >= 300:
          raise HttpError(resp, content, uri=request.uri)
        response = request.postproc(resp, content)
      except HttpError as e:
        exception = e
```

So we need to raise a `googleapiclient.errors.HttpError` from our `postproc`. Then we can get the content from it, and turn it into an `ApiCallError` in our callback.

ZachOrr: This is my confusion - if we get a >= 300 from whatever our API call returns, we don't hit our `postproc` function. We catch the `googleapiclient.errors.HttpError` https://github.com/firebase/firebase-admin-python/pull/283/files#diff-3101917f5d55cf93592ea47ca59a7015R409

Our `postproc` appears to only be called when we get < 300 status code. Which was my question earlier about "when would we get a non-200 and non-error response code from FCM". It feels like we can just throw a generic error in our `postproc` if it's non-200, right? Other cases should already be handled?

hiranya911: FCM mainly responds with 200 and 4xx/5xx responses. In case of 4xx/5xx the above code will raise an `HTTPError`, which we can catch in our callback, and convert into an `ApiCallError`. I believe you're already doing this. 

So that brings us to `postproc`. You're right that `postproc` will only execute for 2xx responses. AFAIK FCM only returns 200 in this range. So we can relax our logic a little bit as follows:

```
def _postproc(self, resp, body):
        """Handle response from batch API request."""
        # This only gets called for 2xx responses.
        return json.loads(body.decode())
```

ZachOrr:
I think I have everything addressed so far.

hiranya911:
Thanks @ZachOrr. Looks good. Please fix the lint error detected by CI, and then I can approve/merge this.

You can add `delete resp` to the method to deal with unused args.

ZachOrr:
🙌 Got the lint passing. Thanks @hiranya911

hiranya911:
Thanks @ZachOrr. I'm going to merge this now. I will try and get it released next week. We also need to add 1-2 integration test cases to further verify this. I can work on it, or you're also welcome to provide a PR for that.

PR #297, opened 2019-06-13
Migrating FCM Send APIs to the New Exceptions

hiranya911:
* Introduced new FCM-specific error types in the `messaging` module.
* Updated the shared error handling code in `_utils` to handle  platform errors.
* Implemented logic for handling errors raised by `requests` and `googleapiclient` packages (latter used by FCM batch operations).
* Updated FCM send APIs to raise new exception types.
* Updated affected tests and docs.

### DIFF
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -97,24 +97,23 @@ def handle_platform_error_from_requests(error, handle_func=None):
     response = error.response
     content = response.content.decode()
     status_code = response.status_code
-    error_dict, code, message = _parse_platform_error(content, status_code)
+    error_dict, message = _parse_platform_error(content, status_code)
     exc = None
     if handle_func:
-        exc = handle_func(error_dict, message, error, error.response)
+        exc = handle_func(error, message, error_dict)
 
-    return exc if exc else handle_requests_error(error, message, code)
+    return exc if exc else handle_requests_error(error, message, error_dict)
 
 
-def handle_requests_error(error, message=None, code=None):
+def handle_requests_error(error, message=None, error_dict=None):
     """Constructs a ``FirebaseError`` from the given requests error.
 
     Args:
         error: An error raised by the requests module while making an HTTP call.
         message: A message to be included in the resulting ``FirebaseError`` (optional). If not
             specified the string representation of the ``error`` argument is used as the message.
-        code: A GCP error code that will be used to determine the resulting error type (optional).
-            If not specified the HTTP status code on the error response is used to determine a
-            suitable error code.
+        error_dict: A parsed error payload to extract error code from (optional). If not specified
+            parses the response payload available in the ``error``.
 
     Returns:
         FirebaseError: A ``FirebaseError`` that can be raised to the user code.
@@ -132,6 +131,10 @@ def handle_requests_error(error, message=None, code=None):
             message='Unknown error while making a remote service call: {0}'.format(error),
             cause=error)
 
+    if error_dict is None:
+        error_dict = {}
+
+    code = error_dict.get('status')
     if not code:
         code = _http_status_to_error_code(error.response.status_code)
     if not message:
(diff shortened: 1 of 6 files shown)


firebase_admin/_utils.py
```diff
@@ -14,13 +14,19 @@
...
+import httplib2
 import requests
+import six
 
 import firebase_admin
 from firebase_admin import exceptions
 
 
-_STATUS_TO_EXCEPTION_TYPE = {
+_ERROR_CODE_TO_EXCEPTION_TYPE = {
```
bklimt: It seems unwise to mix http status codes with grpc status codes in one map.

And you could always do `_ERROR_CODE_TO_EXCEPTION_TYPE[_HTTP_STATUS_TO_ERROR_CODE[X]]` anyway.

firebase_admin/_utils.py
```diff
@@ -45,19 +79,50 @@ def _get_initialized_app(app):
...
 
-def handle_requests_error(error, message=None, status=None):
+
+def handle_platform_error_from_requests(error, handle_func=None):
+    """Constructs a ``FirebaseError`` from the given requests error.
+
+    This can be used to handle errors returned by Google Cloud Platform (GCP) APIs.
+
+    Args:
+        error: An error raised by the reqests module while making an HTTP call to a GCP API.
```
bklimt: s/reqests/requests/

firebase_admin/_utils.py
```diff
@@ -45,19 +79,50 @@ def _get_initialized_app(app):
...
+        handle_func: A function that can be used to handle platform errors in a custom way. When
+            specified, this function will be called with four arguments -- parsed error response,
+            error message, source exception from requests and the HTTP response object.
```
bklimt: Personally, I'd prefer an oxford comma in there. :-) But more seriously, please make this a bulleted list so that it's easier to scan what the four arguments are.

firebase_admin/_utils.py
```diff
@@ -45,19 +79,50 @@ def _get_initialized_app(app):
...
+
+    response = error.response
+    content = response.content.decode()
+    status_code = response.status_code
+    error_dict, code, message = _parse_platform_error(content, status_code)
+    exc = None
+    if handle_func:
+        exc = handle_func(error_dict, message, error, error.response)
+
+    return exc if exc else handle_requests_error(error, message, code)
```
bklimt: If might be a little cleaner than all this if you can make `handle_requests_error` have the same format as `handle_func`. The advantages would be:
1. You could have the `handle_func` default to `handle_requests_error` and not have any of this branching code, and
2. It would be easier to document `handle_func` by pointing to `handle_requests_error` as an example implementation.

firebase_admin/_utils.py
```diff
@@ -45,19 +79,50 @@ def _get_initialized_app(app):
...
+def handle_requests_error(error, message=None, code=None):
     """Constructs a ``FirebaseError`` from the given requests error.
 
     Args:
         error: An error raised by the reqests module while making an HTTP call.
         message: A message to be included in the resulting ``FirebaseError`` (optional). If not
             specified the string representation of the ``error`` argument is used as the message.
-        status: An HTTP status code that will be used to determine the resulting error type
-            (optional). If not specified the HTTP status code on the error response is used.
+        code: An HTTP status code or GCP error code that will be used to determine the resulting
```
bklimt: Yeah, it still seems sketchy to me to mix types here. Would it be possible to make this one or the other?

It occurs to me that `handle_platform_error_from_requests` extracts `code`, which is then ignored if a `handle_func` is set. Maybe if `handle_requests_error` is a `handle_func`, then the code doesn't need to be passed in, and `handle_requests_error` could call `_parse_platform_error` itself?

I'm not sure if that's a good idea, but it's worth thinking about.

firebase_admin/_utils.py
```diff
@@ -75,9 +140,119 @@ def handle_requests_error(error, message=None, status=None):
...
+def _http_response_from_googleapiclient_error(error):
+    """Creates a requests HTTP Response object from the given googleapiclient error."""
+    resp = requests.models.Response()
+    resp.raw = six.BytesIO(error.content)
+    resp.status_code = error.resp.status
+    return resp
+
+
+def _lookup_error_type(code):
+    """Maps an error code to an exception type."""
```
bklimt: Please document what `error code` means here.

firebase_admin/_utils.py
hiranya911: Done. Removed the `HTTP status -> Exception type` mappings from this dict, and refactored the code accordingly.

firebase_admin/_utils.py
hiranya911: Done

firebase_admin/_utils.py
hiranya911: Done

firebase_admin/_utils.py
hiranya911: `handle_requests_error` is supposed to be service-agnostic. I expect to use it for non-OP errors as well (we already do so in the `instance_id` module, which is a non-OP endpoint). So I extracted the OP-specific parsing to a new helper method named `_handle_func_requests` and made its signature consistent with `handle_func`.

But I do still need some branching since we allow `handle_func` to return `None`.

firebase_admin/_utils.py
hiranya911: I've sort of done both, but not quite the way described in your comment. PTAL.

hiranya911:
Thanks for the review @bklimt. This is ready for another look.

PR #306, opened 2019-07-05
Raising FirebaseError from create_session_cookie() API

hiranya911:
* Introduced `InvalidIdTokenError` and `UnexpectedResponseError` exception types.
* Implemented logic for parsing error responses sent by the Auth service.
* Updated the `create_session_cookie()` API to raise new exception types.

### DIFF
--- a/firebase_admin/_auth_utils.py
+++ b/firebase_admin/_auth_utils.py
@@ -255,7 +255,7 @@ def _parse_error_body(response):
 
 
 def _build_error_message(code, exc_type, custom_message):
-    preamble = exc_type.default_message if (
+    default_message = exc_type.default_message if (
         exc_type and hasattr(exc_type, 'default_message')) else 'Error while calling Auth service'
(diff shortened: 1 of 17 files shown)


firebase_admin/_auth_utils.py
```diff
@@ -188,3 +191,71 @@ def validate_action_type(action_type):
...
+            code = code[:separator]
+
+    return code, custom_message
+
+
+def _build_error_message(code, exc_type, custom_message):
+    preamble = exc_type.default_message if (
+        exc_type and hasattr(exc_type, 'default_message')) else 'Error while calling Auth service'
+    ext = ' {0}'.format(custom_message) if custom_message else ''
+    return '{0} ({1}).{2}'.format(preamble, code, ext)
```
bklimt: It's kind of confusing that it's called a `default_message` but then gets used as a `preamble`.

hiranya911: Renamed to `default_message` for consistency.

PR #309, opened 2019-07-18
Introducing UserNotFoundError type

hiranya911:
Error handling updated in `get_user()`, `get_user_by_email()` and `get_user_by_phone_number()` APIs.

### DIFF
--- a/firebase_admin/_auth_utils.py
+++ b/firebase_admin/_auth_utils.py
@@ -210,7 +210,7 @@ class UnexpectedResponseError(exceptions.UnknownError):
 
 
 class UserNotFoundError(exceptions.NotFoundError):
-    """Failed to find a user with the specified details."""
+    """No user record found for the specified identifier."""
 
     default_message = 'No user record found for the given identifier'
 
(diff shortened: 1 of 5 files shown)


firebase_admin/_auth_utils.py
```diff
@@ -243,7 +253,7 @@ def _parse_error_body(response):
         pass
 
     # Auth error response format: {"error": {"message": "AUTH_ERROR_CODE: Optional text"}}
-    code = error_dict.get('message')
+    code = error_dict.get('message') if isinstance(error_dict, dict) else None
```
bklimt: I could be wrong, but this doesn't seem very pythonic to me checking `isinstance`. Is it ever valid for `error_dict` to not be 'None' and also not support the `get` method?

hiranya911: As long as the server response format doesn't change we should be fine. I'm being a little too defensive here. In general, it's not clear what the Auth response format would be in the event of a major outage or similar issue.

PR #314, opened 2019-07-30
Project management API migrated to new error types

hiranya911:
As part of the error handling overhaul of the Admin SDK, we have to use the new exception types defined in `firebase_admin.exceptions` in the `project_management` module as well. This PR:

1. Removes the existing `ApiCallError` type and the custom error handling logic.
2. Delegates error handling to the shared `_utils` module which knows how to handle OnePlatform errors.
3. Updates the affected unit and integration tests.

go/firebase-error-handling
go/firebase-error-handling-py

### DIFF
--- a/firebase_admin/project_management.py
+++ b/firebase_admin/project_management.py
@@ -25,6 +25,7 @@ import requests
 import six
 
 import firebase_admin
+from firebase_admin import exceptions
 from firebase_admin import _http_client
 from firebase_admin import _utils
 
@@ -139,21 +140,6 @@ def _check_not_none(obj, field_name):
     return obj
 
 
-class ApiCallError(Exception):
-    """An error encountered while interacting with the Firebase Project Management Service."""
-
-    def __init__(self, message, error):
-        Exception.__init__(self, message)
-        self.detail = error
-
-
-class _PollingError(Exception):
-    """An error encountered during the polling of an app's creation status."""
-
-    def __init__(self, message):
-        Exception.__init__(self, message)
-
-
 class AndroidApp(object):
     """A reference to an Android app within a Firebase project.
 
@@ -185,7 +171,7 @@ class AndroidApp(object):
             AndroidAppMetadata: An ``AndroidAppMetadata`` instance.
 
         Raises:
-            ApiCallError: If an error occurs while communicating with the Firebase Project
+            FirebaseError: If an error occurs while communicating with the Firebase Project
                 Management Service.
         """
         return self._service.get_android_app_metadata(self._app_id)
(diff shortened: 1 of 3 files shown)


firebase_admin/project_management.py
```diff
@@ -640,16 +607,17 @@ def _poll_app_creation(self, operation_name):
...
                 response = poll_response.get('response')
                 if response:
                     return response
                 else:
-                    raise _PollingError(
-                        'Polling finished, but the operation terminated in an error.')
-        raise _PollingError('Polling deadline exceeded.')
+                    raise exceptions.UnknownError(
+                        'Polling finished, but the operation terminated in an error.',
+                        http_response=http_response)
```
weixifan: Also for your consideration: we can actually use `poll_response.get('error')` as the http_response argument -- if it is non-empty and non-null -- here instead of the full `http_response`.

firebase_admin/project_management.py
```diff
@@ -678,28 +646,20 @@ def add_sha_certificate(self, app_id, certificate_to_add):
...
         name = _check_not_none(certificate_to_delete, 'certificate_to_delete').name
         path = '/v1beta1/{0}'.format(name)
-        self._make_request('delete', path, name, 'SHA ID')
+        self._make_request('delete', path)
+
+    def _make_request(self, method, url, json=None):
+        body, _ = self._body_and_response(method, url, json)
+        return body
 
-    def _make_request(self, method, url, resource_identifier, resource_identifier_label, json=None):
```
weixifan: This is a somewhat minor point: Does `_utils.handle_platform_error_from_requests` raise an error that includes the request? If not, then we might be losing some helpful information in the error.

firebase_admin/project_management.py
hiranya911: http_response should be the full `requests.Response` object returned by the `requests` library -- including headers, body and request information. So this is handled correctly.

firebase_admin/project_management.py
hiranya911: One can get to the outgoing `requests.Request` object as follows:

```
try:
  method_call()
except exceptions.FirebaseError as error:
  req = error.http_response.request
  print(req.method, req.url, req.body)
```

The raised error also includes the original `requests.error.RequestError`. Among all this, and the exception stacktrace, developers should have enough details to implement fine-grained error handling.

weixifan: Ah, sounds good.

firebase_admin/project_management.py
weixifan: Thanks for the explanation. Sounds good.

PR #318, opened 2019-08-05
Migrated the db module to the new exception types

hiranya911:
1. Using new exception types in the `db` module.
2. Removing the old `db.ApiCallError` type.
3. Renamed `TransactionError` to `TransactionAbortedError` and extended from `exceptions.AbortedError` for clarity.

Question: RTDB returns 401 Unauthenticated errors when it really should be sending 403 Forbidden. The error message clearly reads "Permission denied" (see integration tests). Should we handle this case, and turn errors into `PermissionDeniedError` instead? 

This completes the migration to the new error handling model for the Python SDK.

### DIFF
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -457,11 +457,9 @@ class TestReference(object):
             del data
             raise ValueError('test error')
 
-        with pytest.raises(db.TransactionAbortedError) as excinfo:
+        with pytest.raises(ValueError) as excinfo:
             ref.transaction(transaction_update)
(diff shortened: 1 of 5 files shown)


tests/test_db.py
```diff
@@ -64,6 +66,34 @@ class _Object(object):
     pass
 
 
+class _RefOperations(object):
```
bklimt: Please add a comment for this class explaining what it's for.

You might also want to move it down closer to where it's used?

hiranya911: Done

PR #321, opened 2019-08-13
FCM multicast snippets

hiranya911:
Snippets to appear on https://firebase.google.com/docs/cloud-messaging/send-message#send_messages_to_multiple_devices

### DIFF
--- a/snippets/messaging/cloud_messaging.py
+++ b/snippets/messaging/cloud_messaging.py
@@ -220,3 +220,72 @@ def unsubscribe_from_topic():
     # for the contents of response.
     print(response.success_count, 'tokens were unsubscribed successfully')
     # [END unsubscribe]
+
+
+def send_all():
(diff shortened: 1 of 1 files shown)


hiranya911:
@egilmorez can we update the site with these snippets?

snippets/messaging/cloud_messaging.py
```diff
@@ -220,3 +220,72 @@ def unsubscribe_from_topic():
...
     # [END unsubscribe]
+
+
+def send_all():
+    registration_token = 'YOUR_REGISTRATION_TOKEN'
+    # [START send_all]
+    # Create a list containing up to 100 messages.
+    messages = [
+        messaging.Message(
+            notification=messaging.Notification('Price drop', '5% off all electronics'),
```
bklimt: Those % signs in a string literal don't cause any problems, I guess?

hiranya911: Not a problem in this context.

PR #323, opened 2019-08-14
Adding beginning of _MLKitService

ifielker:


firebase_admin/mlkit.py
```diff
@@ -0,0 +1,5 @@
+class _MLKitService(object):
```
hiranya911: Please add the license header.

PR #331, opened 2019-08-20
Firebase ML Kit List Models API implementation

ifielker:
Includes ListModelsPage and ModelIterator

### DIFF
--- a/firebase_admin/mlkit.py
+++ b/firebase_admin/mlkit.py
@@ -27,7 +27,7 @@ from firebase_admin import _utils
 
 
 _MLKIT_ATTRIBUTE = '_mlkit'
-MAX_PAGE_SIZE = 100
+_MAX_PAGE_SIZE = 100
 
 
 def _get_mlkit_service(app):
@@ -145,10 +145,10 @@ class ListModelsPage(object):
         Returns:
             iterator: An iterator of Model instances.
         """
-        return ModelIterator(self)
+        return _ModelIterator(self)
 
 
-class ModelIterator(object):
+class _ModelIterator(object):
     """An iterator that allows iterating over models, one at a time.
 
     This implementation loads a page of models into memory, and iterates on them.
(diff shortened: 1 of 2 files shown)


firebase_admin/mlkit.py
```diff
@@ -27,6 +27,7 @@
 
 
 _MLKIT_ATTRIBUTE = '_mlkit'
+MAX_PAGE_SIZE = 100
```
hiranya911: Prefix with `_` to indicate that it's not part of the public API.

firebase_admin/mlkit.py
```diff
@@ -69,7 +76,107 @@ def __eq__(self, other):
...
+        project starting from this page. The iterator will never buffer more than
+        one page of models in memory at a time.
+
+        Returns:
+            iterator: An iterator of Model instances.
+        """
+        return ModelIterator(self)
+
+
+class ModelIterator(object):
```
hiranya911: Rename to `_ModelIterator`. This doesn't need to be in the public API.

tests/test_mlkit.py
```diff
@@ -33,27 +35,84 @@
...
+MODEL_NAME_3 = 'projects/{0}/models/{1}'.format(PROJECT_ID, MODEL_ID_3)
+DISPLAY_NAME_3 = 'displayName3'
+MODEL_JSON_3 = {
+    'name': MODEL_NAME_3,
+    'displayName': DISPLAY_NAME_3
+}
+MODEL_3 = mlkit.Model(MODEL_JSON_3)
+
 _EMPTY_RESPONSE = json.dumps({})
+_DEFAULT_GET_RESPONSE = json.dumps(MODEL_JSON_1)
```
hiranya911: Nit: I see some constants prefixed with `_` and some not. Should we be consistent within the same test file?

firebase_admin/mlkit.py

firebase_admin/mlkit.py

tests/test_mlkit.py
ifielker: done

PR #337, opened 2019-09-03
Firebase ML Kit Create Model API implementation

ifielker:
Firebase ML Kit Create Model API implementation
* also long running operation handling

### DIFF
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -106,6 +106,27 @@ def handle_platform_error_from_requests(error, handle_func=None):
     return exc if exc else _handle_func_requests(error, message, error_dict)
 
 
+def handle_operation_error(error):
+    """Constructs a ``FirebaseError`` from the given operation error.
+
+    Args:
+        error: An error returned by a long running operation.
+
+    Returns:
+        FirebaseError: A ``FirebaseError`` that can be raised to the user code.
+    """
+    if not isinstance(error, dict):
+        return exceptions.UnknownError(
+            message='Unknown error while making a remote service call: {0}'.format(error),
+            cause=error)
+
+    status_code = error.get('code')
+    message = error.get('message')
+    error_code = _http_status_to_error_code(status_code)
+    err_type = _error_code_to_exception_type(error_code)
+    return err_type(message=message)
+
+
 def _handle_func_requests(error, message, error_dict):
     """Constructs a ``FirebaseError`` from the given GCP error.
 
--- a/firebase_admin/mlkit.py
+++ b/firebase_admin/mlkit.py
@@ -150,13 +150,9 @@ class Model(object):
         return model
 
     def _update_from_dict(self, data):
-        data_copy = dict(data)
-        tflite_format = None
-        tflite_format_data = data_copy.pop('tfliteModel', None)
-        if tflite_format_data:
-            tflite_format = TFLiteFormat.from_dict(tflite_format_data)
-        self.model_format = tflite_format
-        self._data = data_copy
+        copy = Model.from_dict(data)
+        self.model_format = copy.model_format
+        self._data = copy._data # pylint: disable=protected-access
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
@@ -233,13 +229,15 @@ class Model(object):
                     len(self._data.get('activeOperations')) > 0)
 
     def wait_for_unlocked(self, max_time_seconds=None):
-        if self.locked:
-            mlkit_service = _get_mlkit_service(self._app)
-            op_name = self._data.get('activeOperations')[0].get('name')
-            model_dict = mlkit_service.handle_operation(
-                mlkit_service.get_operation(op_name),
-                max_time_seconds=max_time_seconds)
-            self._update_from_dict(model_dict)
+        if not self.locked:
+            return
+
+        mlkit_service = _get_mlkit_service(self._app)
+        op_name = self._data.get('activeOperations')[0].get('name')
+        model_dict = mlkit_service.handle_operation(
+            mlkit_service.get_operation(op_name),
+            max_time_seconds=max_time_seconds)
+        self._update_from_dict(model_dict)
 
     @property
     def model_format(self):
@@ -561,6 +559,19 @@ class _MLKitService(object):
         except requests.exceptions.RequestException as error:
             raise _utils.handle_platform_error_from_requests(error)
 
+    def _exponential_backoff(self, max_polling_attempts, current_attempt, stop_time):
+        """Sleeps for the appropriate amount of time. Or thows deadline exceeded."""
+        if max_polling_attempts is not None and current_attempt >= max_polling_attempts:
+            raise exceptions.DeadlineExceededError('Polling max attempts exceeded.')
+        delay_factor = pow(
+            _MLKitService.POLL_EXPONENTIAL_BACKOFF_FACTOR, current_attempt)
+        wait_time_seconds = delay_factor * _MLKitService.POLL_BASE_WAIT_TIME_SECONDS
+        after_sleep_time = (datetime.datetime.now() +
+                            datetime.timedelta(seconds=wait_time_seconds))
+        if stop_time is not None and after_sleep_time > stop_time:
+            raise exceptions.DeadlineExceededError('Polling max time exceeded.')
+        time.sleep(wait_time_seconds)
+
     def handle_operation(self, operation, max_polling_attempts=None, max_time_seconds=None,
                          always_return_model=False):
         """Handles long running operations.
@@ -580,6 +591,7 @@ class _MLKitService(object):
         Raises:
             TypeError: if the operation is not a dictionary.
             ValueError: If the operation is malformed.
+            err: If the operation exceeds polling attempts or stop_time
         """
         if not isinstance(operation, dict):
             raise TypeError('Operation must be a dictionary.')
@@ -590,34 +602,26 @@ class _MLKitService(object):
         start_time = datetime.datetime.now()
         stop_time = (None if max_time_seconds is None else
                      start_time + datetime.timedelta(seconds=max_time_seconds))
-        while True:
-            if operation.get('done'):
-                if operation.get('response'):
-                    return operation.get('response')
-                elif operation.get('error'):
-                    raise _utils.handle_operation_error(operation.get('error'))
-                else:
-                    # A 'done' operation must have either a response or an error.
-                    raise ValueError('Operation is malformed.')
-            else:
-                # We just got this operation. Wait before getting another
-                # so we don't exceed the GetOperation maximum request rate.
-                if max_polling_attempts is not None and current_attempt >= max_polling_attempts:
-                    if always_return_model:
-                        return get_model(model_id).as_dict()
-                    raise exceptions.DeadlineExceededError('Polling max attempts exceeded.')
-                delay_factor = pow(
-                    _MLKitService.POLL_EXPONENTIAL_BACKOFF_FACTOR, current_attempt)
-                wait_time_seconds = delay_factor * _MLKitService.POLL_BASE_WAIT_TIME_SECONDS
-                after_sleep_time = (datetime.datetime.now() +
-                                    datetime.timedelta(seconds=wait_time_seconds))
-                if stop_time is not None and after_sleep_time > stop_time:
-                    if always_return_model:
-                        return get_model(model_id).as_dict()
-                    raise exceptions.DeadlineExceededError('Polling max time exceeded.')
-                time.sleep(wait_time_seconds)
-                operation = self.get_operation(op_name)
-                current_attempt += 1
+        while not operation.get('done'):
+            # We just got this operation. Wait before getting another
+            # so we don't exceed the GetOperation maximum request rate.
+            try:
+                self._exponential_backoff(max_polling_attempts, current_attempt, stop_time)
+            except exceptions.DeadlineExceededError as err:
+                if always_return_model:
+                    return get_model(model_id).as_dict()
+                raise err
+            operation = self.get_operation(op_name)
+            current_attempt += 1
+
+        if operation.get('response'):
+            return operation.get('response')
+        elif operation.get('error'):
+            raise _utils.handle_operation_error(operation.get('error'))
+        else:
+            # A 'done' operation must have either a response or an error.
+            raise ValueError('Operation is malformed.')
+
 
 
     def create_model(self, model):
--- a/firebase_admin/mlkit.py
+++ b/firebase_admin/mlkit.py
@@ -28,6 +28,7 @@ import six
 
 from firebase_admin import _http_client
 from firebase_admin import _utils
+from firebase_admin import exceptions
 
 
 _MLKIT_ATTRIBUTE = '_mlkit'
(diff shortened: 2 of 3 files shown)


tests/test_mlkit.py
```diff
@@ -185,29 +230,47 @@
...
+def instrument_mlkit_service(app=None, status=200, uri=None, payload=None):
     if not app:
         app = firebase_admin.get_app()
     mlkit_service = mlkit._get_mlkit_service(app)
     recorder = []
-    mlkit_service._client.session.mount(
-        'https://mlkit.googleapis.com',
-        testutils.MockAdapter(payload, status, recorder)
-    )
+    if uri is None or uri is 'projects':
```
hiranya911: May be change `uri` to a boolean `operations=False` parameter.

Or add an else clause below and raise an exception.

tests/test_mlkit.py
```diff
@@ -353,7 +416,131 @@ def test_model_source_validation_errors(self, model_source):
...
+    @staticmethod
+    def _url(project_id):
+        return BASE_URL + 'projects/{0}/models'.format(project_id)
+
+    @staticmethod
+    def _op_url(project_id, model_id):
+        return BASE_URL + \
+            'operations/project/{0}/model/{1}/operation/123'.format(project_id, model_id)
+
+    def test_create_model_immediate_done(self):
```
hiranya911: You can drop _create_model_ part from these test case names. They are already nested under `TestCreateModel`. Use short names like `test_immediate_done`, `test_with_polling`, and `test_operation_error`.

firebase_admin/_utils.py
```diff
@@ -106,6 +106,27 @@ def handle_platform_error_from_requests(error, handle_func=None):
...
+
+    Returns:
+        FirebaseError: A ``FirebaseError`` that can be raised to the user code.
+    """
+    if not isinstance(error, dict):
+        return exceptions.UnknownError(
+            message='Unknown error while making a remote service call: {0}'.format(error),
+            cause=error)
+
+    status_code = error.get('code')
```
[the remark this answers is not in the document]
ifielker: it has a numeric code and a string message plus optional Any field for details. No status field.

firebase_admin/mlkit.py
```diff
@@ -53,6 +57,11 @@ def _get_mlkit_service(app):
     return _utils.get_app_service(app, _MLKIT_ATTRIBUTE, _MLKitService)
 
 
+def create_model(model, app=None):
```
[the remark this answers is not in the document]

firebase_admin/mlkit.py
```diff
@@ -459,6 +482,45 @@ def __init__(self, app):
...
+            base_url=_MLKitService.OPERATION_URL)
+
+    def get_operation(self, op_name):
+        _validate_operation_name(op_name)
+        try:
+            return self._operation_client.body('get', url=op_name)
+        except requests.exceptions.RequestException as error:
+            raise _utils.handle_platform_error_from_requests(error)
+
+    def handle_operation(self, operation):
```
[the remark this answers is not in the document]

tests/test_mlkit.py
```diff
@@ -185,29 +230,47 @@
 invalid_string_or_none_args = [0, -1, 4.2, 0x10, False, list(), dict()]
 
 
+# For validation type errors
 def check_error(err, err_type, msg=None):
-    assert isinstance(err, err_type)
+    err_value = err.value
```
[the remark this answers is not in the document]
ifielker: done.

tests/test_mlkit.py
ifielker: done

tests/test_mlkit.py
ifielker: done

tests/test_mlkit.py
```diff
@@ -299,7 +363,7 @@ def test_model_format_setters(self):
     def test_model_display_name_validation_errors(self, display_name, exc_type):
         with pytest.raises(exc_type) as err:
             mlkit.Model(display_name=display_name)
-        check_error(err.value, exc_type)
+        check_error(err, exc_type)
```
hiranya911: Lets rename err to excinfo to be consistent with other test modules and the helper functions above.

firebase_admin/mlkit.py
```diff
@@ -459,6 +523,61 @@ def __init__(self, app):
...
+
+        for current_attempt in range(_MLKitService.MAX_POLLING_ATTEMPTS):
+            if operation.get('done'):
+                if operation.get('response'):
+                    return operation.get('response')
+                elif operation.get('error'):
+                    raise _utils.handle_operation_error(operation.get('error'))
+                else:
+                    # A 'done' operation must have either a response or an error.
+                    raise ValueError('Operation is malformed.')
```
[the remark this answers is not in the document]
ifielker: No, this is not expected, the backend should only give us valid Operations, but it has happened before (during console development).

tests/test_mlkit.py
ifielker: done

firebase_admin/mlkit.py
```diff
@@ -448,6 +507,11 @@ class _MLKitService(object):
     """Firebase MLKit service."""
 
     PROJECT_URL = 'https://mlkit.googleapis.com/v1beta1/projects/{0}/'
+    OPERATION_URL = 'https://mlkit.googleapis.com/v1beta1/'
+    OPERATION_POLL_DELAY_SECONDS = 30
+    MAX_POLLING_ATTEMPTS = 10
```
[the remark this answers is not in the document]
ifielker: Discussed offline. (& agreed upon changes done)

firebase_admin/mlkit.py
```diff
@@ -173,6 +232,15 @@ def locked(self):
         return bool(self._data.get('activeOperations') and
                     len(self._data.get('activeOperations')) > 0)
 
+    def wait_for_unlocked(self, max_time_seconds=None):
+        if self.locked:
```
hiranya911: Return early if `not self.locked`

firebase_admin/mlkit.py
```diff
@@ -459,6 +550,85 @@ def __init__(self, app):
...
+        if not isinstance(operation, dict):
+            raise TypeError('Operation must be a dictionary.')
+        op_name = operation.get('name')
+        _, model_id = _validate_and_parse_operation_name(op_name)
+
+        current_attempt = 0
+        start_time = datetime.datetime.now()
+        stop_time = (None if max_time_seconds is None else
+                     start_time + datetime.timedelta(seconds=max_time_seconds))
+        while True:
```
hiranya911: `while not operation.get('done'):` might simplify this loop a bit.

firebase_admin/mlkit.py
```diff
@@ -459,6 +550,85 @@ def __init__(self, app):
...
+                operation = self.get_operation(op_name)
+                current_attempt += 1
+
+
+    def create_model(self, model):
+        _validate_model(model)
+        try:
+            return self.handle_operation(
+                self._client.body('post', url='models', json=model.as_dict()),
+                max_polling_attempts=1,
```
hiranya911: I think this shouldn't poll at all. It should be `max_polling_attempts=0`.

firebase_admin/mlkit.py
```diff
@@ -459,6 +550,85 @@ def __init__(self, app):
...
+                    # A 'done' operation must have either a response or an error.
+                    raise ValueError('Operation is malformed.')
+            else:
+                # We just got this operation. Wait before getting another
+                # so we don't exceed the GetOperation maximum request rate.
+                if max_polling_attempts is not None and current_attempt >= max_polling_attempts:
+                    if always_return_model:
+                        return get_model(model_id).as_dict()
+                    raise exceptions.DeadlineExceededError('Polling max attempts exceeded.')
+                delay_factor = pow(
```
hiranya911: Extract this part (up until sleep) into a helper method.

```
def _exponential_backoff(self, current_attempt, stop_time):
    # calculate delay, validate and sleep
```

firebase_admin/mlkit.py
```diff
@@ -89,16 +138,26 @@ def __init__(self, display_name=None, tags=None, model_format=None):
...
         tflite_format_data = data_copy.pop('tfliteModel', None)
         if tflite_format_data:
             tflite_format = TFLiteFormat.from_dict(tflite_format_data)
         model = Model(model_format=tflite_format)
         model._data = data_copy  # pylint: disable=protected-access
+        model._app = app # pylint: disable=protected-access
         return model
 
+    def _update_from_dict(self, data):
+        data_copy = dict(data)
```
hiranya911:
```
copy = Model.from_dict(data)
self.model_format = copy.model_format
self._data = copy._data
```

tests/test_mlkit.py
```diff
@@ -185,34 +265,84 @@
...
+    err = excinfo.value
     assert isinstance(err, exceptions.FirebaseError)
     assert err.code == code
     assert err.http_response is not None
     assert err.http_response.status_code == status
     assert str(err) == msg
 
 
-def instrument_mlkit_service(app=None, status=200, payload=None):
+def instrument_mlkit_service(status=200, payload=None, operations=False, app=None):
```
hiranya911: You can modify this helper to handle the multiple requests too.

```
if isinstance(status, list):
   adapter = MockMultiRequestAdapter(...)
else:
   adapter = MockAdapter()
```

tests/test_mlkit.py
```diff
@@ -185,34 +265,84 @@
...
     return recorder
 
 
 class TestModel(object):
     """Tests mlkit.Model class."""
+    @classmethod
+    def setup_class(cls):
+        cred = testutils.MockCredential()
+        firebase_admin.initialize_app(cred, {'projectId': PROJECT_ID})
+        mlkit._MLKitService.POLL_BASE_WAIT_TIME_SECONDS = 0.1  # shorter for test
```
hiranya911: Can this be 0?

tests/test_mlkit.py
```diff
@@ -351,9 +481,180 @@ def test_model_source_validation_errors(self, model_source):
...
     def test_gcs_tflite_source_validation_errors(self, uri, exc_type):
-        with pytest.raises(exc_type) as err:
+        with pytest.raises(exc_type) as excinfo:
             mlkit.TFLiteGCSModelSource(gcs_tflite_uri=uri)
-        check_error(err.value, exc_type)
+        check_error(excinfo, exc_type)
+
+    def test_wait_for_unlocked_not_locked(self):
+        model = mlkit.Model(display_name="not_locked")
+        model.wait_for_unlocked()
```
hiranya911: Perhaps this should return something. False to indicate it didn't wait, True to indicate it waited successfully.

tests/test_mlkit.py
```diff
@@ -351,9 +481,180 @@ def test_model_source_validation_errors(self, model_source):
...
+        assert len(recorder) == 1
+        assert recorder[0].method == 'GET'
+        assert recorder[0].url == TestModel._op_url(PROJECT_ID, MODEL_ID_1)
+
+    def test_wait_for_unlocked_timeout(self):
+        instrument_mlkit_service(status=200, operations=True, payload=OPERATION_NOT_DONE_RESPONSE)
+        mlkit._MLKitService.POLL_BASE_WAIT_TIME_SECONDS = 5  # longer for timeout
+        model = mlkit.Model.from_dict(LOCKED_MODEL_JSON_1)
+        with pytest.raises(Exception) as excinfo:
+            model.wait_for_unlocked(max_time_seconds=3)
```
hiranya911: Set this to a much smaller value please (e.g. 0.1). We want the unit tests to run very fast.

Alternatively, can we add a Sleeper type somewhere that can be modified for tests?

```
# mlkit.py

class Sleeper(object):
     def sleep(self, duration_seconds):
         time.sleep(duration_seconds)

_SLEEPER = Sleeper()


# Unit tests
class TestSleeper():
    def sleep(self, delay_seconds):
        pass

mlkit._SLEEPER = TestSleeper()
```

tests/test_mlkit.py
```diff
@@ -351,9 +481,180 @@ def test_model_source_validation_errors(self, model_source):
...
+        assert model == CREATED_MODEL_1
+        assert len(create_recorder) == 1
+        assert create_recorder[0].method == 'POST'
+        assert create_recorder[0].url == TestCreateModel._url(PROJECT_ID)
+        assert len(operation_recorder) == 1
+        assert operation_recorder[0].method == 'GET'
+        assert operation_recorder[0].url == TestCreateModel._op_url(PROJECT_ID, MODEL_ID_1)
+
+    def test_with_get_returns_locked(self):
+        recorder = instrument_mlkit_service_multi(
```
hiranya911: I think `create_model()` should only ever make 1 RPC. WDYT?

firebase_admin/mlkit.py

firebase_admin/mlkit.py

firebase_admin/mlkit.py
ifielker: I can do that, but in practice almost nothing is done on the initial operation return from the create/update functions, but a lot of faster things are done on the first poll. That's why I set it to poll once. 
Why do you prefer no polling?

firebase_admin/mlkit.py

firebase_admin/mlkit.py
ifielker: nice! Thanks

tests/test_mlkit.py
ifielker: done

tests/test_mlkit.py
ifielker: I left it at 0.1 because that means it actually sleeps which is closer to the actual code execution pattern. (I assumed that sleep(0) doesn't actually sleep. Is that wrong?)
If you are worried about total test time, all 104 tests pass in 0.76 seconds on my machine...

tests/test_mlkit.py
ifielker: If you mean "was it already unlocked when I asked it to wait" then users can just check the .locked property of the model. I'm trying to imagine a scenario where I care if it was already unlocked but don't want to check the .locked property. Can you think of any?
If you mean "we tried to wait but the wait time was exceeded" then it throws exceptions.DeadlineExceeded in that case. To me, the point of this function is to have a model that is unlocked at the end of it.

tests/test_mlkit.py
ifielker: All 104 tests run in under 1 second total. 
This max time in this test is designed to time out before it polls a single time because I set the POLL_BASE_WAIT_TIME_SECONDS to 5 seconds and the maximum wait time to 3 seconds, so when it calculates the after_sleep_time it will not poll at all (stop_time = x + 3, after_sleep_time = x + 5).

tests/test_mlkit.py
ifielker: I'm not sure how that would work. It would have to be a CreateModel rpc which returns an operation. But create_model always returns a (possibly locked) model... The operation does not have model information if it is still running... 

I think it should do exactly what it does now: 
call the CreateModel api, and up to 1 GetOperation poll because that will give us the "low hanging fruit" for models to not be locked. If the GetOperation still results in a not-done operation, then we need to GetModel to get the model information (not available in the Operations returned) so we can return the locked model.

firebase_admin/mlkit.py
```diff
@@ -459,6 +558,93 @@ def __init__(self, app):
...
+            current_attempt += 1
+
+        if operation.get('done'):
+            if operation.get('response'):
+                return operation.get('response')
+            elif operation.get('error'):
+                raise _utils.handle_operation_error(operation.get('error'))
+            else:
+                # A 'done' operation must have either a response or an error.
+                raise ValueError('Operation is malformed.')
```
hiranya911: Push this out of the if-else structure so it becomes the default exit clause. Otherwise `done=false and always_return_model=false` case goes unhandled. This probably doesn't occur in practice, but it's good to cover all possible branches. Also may be change the error message:

```
raise ValueError('Invalid operation response: {0}`.format(operation))
```

firebase_admin/mlkit.py
```diff
@@ -459,6 +558,93 @@ def __init__(self, app):
...
+        if not isinstance(operation, dict):
+            raise TypeError('Operation must be a dictionary.')
+        op_name = operation.get('name')
+        _, model_id = _validate_and_parse_operation_name(op_name)
+
+        current_attempt = 0
+        start_time = datetime.datetime.now()
+        stop_time = (None if max_time_seconds is None else
+                     start_time + datetime.timedelta(seconds=max_time_seconds))
+        while polling and not operation.get('done'):
```
hiranya911: Nit: s/polling/wait_for_operation/ for clarity.

firebase_admin/mlkit.py
ifielker: removing always_return_model instead and making it the opposite of polling essentially. So we either have a done operation, a deadline exceeded, or (no polling) and return a locked model.

PR #338, opened 2019-09-05
Fix: corrected ValueError message

bertdida:
Used type of `check_revoked` variable instead of `app` for `ValueError` message.


bertdida:
@googlebot I signed it!

Issue #349, opened 2019-09-17
FR: string representation


Azd325:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Linux
  * Firebase SDK version: 3.0.0
  * Library version: 3.0.0
  * Firebase Product: cloud messages (auth, database, storage, etc)

#### Steps to reproduce:

It would be great to have a string representation of a object to have a quick preview of his contents because this is not really helpful.
```python
> str(message)
'<firebase_admin._messaging_utils.Message object at 0x11bdcda10>'

```

I think it would be great if at least the str function could present the dict variant of e.g the message Object
```python

class Message:
    def __str__(self):
        return self.__dict__

> str(Message)
{'data': None, 'notification': <firebase_admin._messaging_utils.Notification object at 0x11bdad050>, 'android': None, 'webpush': None, 'apns': None, 'fcm_options': None, 'token': None, 'topic': None, 'condition': None}


```

hiranya911:
Sounds like a good idea. There's a custom `MessageEncoder` class that knows how to serialize all these objects. Perhaps we can use that to implement this.

Azd325:
Would be quite nice if there some at least some human readable for the str version. :)

Thanks for the hint with `MessageEncoder`

Azd325:
`MessageEncoder` does the job for the moment.
Is there also a way that is not a function in a protected module?

hiranya911:
@Azd325 I wasn't suggesting you directly use `MessageEncoder`. Rather implement the functionality as a proper public API, based on `MessageEncoder`.

lahirumaramba:
`__str__()` function is now implemented in `Message` class with the changes included in #350

PR #350, opened 2019-09-23
fix(fcm): String representation in Message class

lahirumaramba:
### Discussion

  * Add string representation in `Message` class
  * Use `MessageEncoder` to convert `Message` objects to valid JSON strings

### Testing

   * None

### API Changes

   * None

Resolves: #349 

RELEASE NOTE: `Message` class now implements the `__str__()` contract which can be used to obtain a string representation of a `Message` instance.

### DIFF
--- a/firebase_admin/_messaging_utils.py
+++ b/firebase_admin/_messaging_utils.py
@@ -57,6 +57,9 @@ class Message(object):
         self.topic = topic
         self.condition = condition
 
+    def __str__(self):
+        return json.dumps(MessageEncoder().default(self))
+
 
 class MulticastMessage(object):
     """A message that can be sent to multiple tokens via Firebase Cloud Messaging.
--- a/firebase_admin/_messaging_utils.py
+++ b/firebase_admin/_messaging_utils.py
@@ -58,7 +58,7 @@ class Message(object):
         self.condition = condition
 
     def __str__(self):
-        return json.dumps(MessageEncoder().default(self))
+        return json.dumps(self, cls=MessageEncoder, sort_keys=True)
 
 
 class MulticastMessage(object):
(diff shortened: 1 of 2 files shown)


lahirumaramba:
Sample output:
```
>>> from firebase_admin import messaging
>>> message = messaging.Message(
        data={
            'score': '850',
            'time': '2:45',
        },
        token='token',
    )
>>> print(message)
{"token": "token", "data": {"score": "850", "time": "2:45"}}
>>> str(message)
'{"token": "token", "data": {"score": "850", "time": "2:45"}}'
>>>
```

firebase_admin/_messaging_utils.py
```diff
@@ -57,6 +57,9 @@ def __init__(self, data=None, notification=None, android=None, webpush=None, apn
         self.topic = topic
         self.condition = condition
 
+    def __str__(self):
+        return json.dumps(MessageEncoder().default(self))
```
hiranya911: Would this work?

```
json.dumps(self, cls=MessageEncoder)
```

lahirumaramba: Yes! Thank you.
Will update with test cases soon.

lahirumaramba:
> One suggestion added.
> 
> This also brings up the question whether we should support `__str__` on all public classes in this module (e.g. AndroidConfig, Notification etc). But `Message` is probably enough to begin with. After all, other objects are never used by themselves -- they are always part of a `Message`.
> 
> Also please add a test case.

- Good point! I also thought about adding `__str__` support to all other classes but like you said they will always be part of a `Message`. I think we can start with `Message` for now.

- PR is updated with test cases.

- One other thing to note is that since we use `MessageEncoder` in the implementation, `__str__` inherits the validation checks from `MessageEncoder`... I do not think this is a problem but wanted to mention as this could be different from the expected behavior of a `str()` function.

PR #352, opened 2019-09-27
Remove (base64) 'REDACTED' passwords from user records.

rsgowman:
These values *look* like passwords hashes, but aren't, leading to
potential confusion.

Additionally, added docs to CONTRIBUTING.md detailing how to add the
permission that causes password hashes to be properly returned as well
as adjusting the test failure message should the developer not add that
permission.

b/141189502

### DIFF
--- a/firebase_admin/_user_mgt.py
+++ b/firebase_admin/_user_mgt.py
@@ -26,7 +26,7 @@ from firebase_admin import _user_import
 
 MAX_LIST_USERS_RESULTS = 1000
 MAX_IMPORT_USERS_SIZE = 1000
-B64_REDACTED = base64.b64encode("REDACTED")
+B64_REDACTED = base64.b64encode(b'REDACTED')
 
 
 class Sentinel(object):
--- a/firebase_admin/_user_mgt.py
+++ b/firebase_admin/_user_mgt.py
@@ -14,8 +14,8 @@
 
 """Firebase user management sub module."""
 
+import base64
 import json
-
 import requests
 import six
 from six.moves import urllib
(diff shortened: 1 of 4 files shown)


rsgowman:
Port of https://github.com/firebase/firebase-admin-node/pull/660

firebase_admin/_user_mgt.py
```diff
@@ -26,6 +26,7 @@
 
 MAX_LIST_USERS_RESULTS = 1000
 MAX_IMPORT_USERS_SIZE = 1000
+B64_REDACTED = base64.b64encode("REDACTED")
```
hiranya911: Add `b'REDACTED'` to fix the test failures on Python3

integration/test_auth.py
```diff
@@ -228,8 +228,14 @@ def test_list_users(new_user_list):
             assert isinstance(user, auth.ExportedUserRecord)
             if user.uid in new_user_list:
                 fetched.append(user.uid)
-                assert user.password_hash is not None
-                assert user.password_salt is not None
+                assert user.password_hash is not None, (
+                    "Missing password_hash field. A common cause would be forgetting to add the " +
```
hiranya911: Nit: Extract the repeated part into  constant.

firebase_admin/_user_mgt.py
rsgowman: Oops. (Fixed; thanks.)

integration/test_auth.py
rsgowman: Done.

PR #371, opened 2019-11-28
Modify Operation Handling to not require a name for Done Operations

ifielker:
Some operations which are immediately done do not require or have an operation name (since there will not be any polling of a 'done' operation).

### DIFF
--- a/firebase_admin/mlkit.py
+++ b/firebase_admin/mlkit.py
@@ -810,28 +810,36 @@ class _MLKitService(object):
         """
         if not isinstance(operation, dict):
             raise TypeError('Operation must be a dictionary.')
-        op_name = operation.get('name')
-        _, model_id = _validate_and_parse_operation_name(op_name)
-
-        current_attempt = 0
-        start_time = datetime.datetime.now()
-        stop_time = (None if max_time_seconds is None else
-                     start_time + datetime.timedelta(seconds=max_time_seconds))
-        while wait_for_operation and not operation.get('done'):
-            # We just got this operation. Wait before getting another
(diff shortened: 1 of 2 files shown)


firebase_admin/mlkit.py
```diff
@@ -810,28 +810,36 @@ def handle_operation(self, operation, wait_for_operation=False, max_time_seconds
...
             if operation.get('response'):
                 return operation.get('response')
             elif operation.get('error'):
                 raise _utils.handle_operation_error(operation.get('error'))
-
-        # If the operation is not complete or timed out, return a (locked) model instead
-        return get_model(model_id).as_dict()
+            raise exceptions.UnknownError(message='Internal Error: Malformed Operation.')
+        else:
+            op_name = operation.get('name')
```
hiranya911: Nit: Consider moving this `else` block logic into a helper function. `_poll_until_complete(operation)`  or similar.

ifielker: I think that might be misleading. We only do the polling if wait_for_operation is true. So we could either only call the _poll_until_complete when wait_for_operation is true and have it only be the while loop and then duplicate the done checking below the while loop in both handle_operation and _poll_until_complete or we could name it something else like _poll_until_complete_if_waiting or something? I actually prefer the way it is now, but I'm open to other suggestions.

hiranya911: May be `_poll_until_complete(operation, wait_for_operation)`?

I'm fine with how it is implemented now. Just thought the else block could be a bit smaller for clarity. Your call.

PR #375, opened 2019-12-13
Adding File naming capability to from_saved_model and from_keras_model.

ifielker:
So you can optionally choose a file name for the model being saved to your bucket. This can be useful if you have more than one model and don't want to overwrite.

### DIFF
--- a/firebase_admin/ml.py
+++ b/firebase_admin/ml.py
@@ -51,7 +51,6 @@ _MAX_PAGE_SIZE = 100
 _MODEL_ID_PATTERN = re.compile(r'^[A-Za-z0-9_-]{1,60}$')
 _DISPLAY_NAME_PATTERN = re.compile(r'^[A-Za-z0-9_-]{1,60}$')
 _TAG_PATTERN = re.compile(r'^[A-Za-z0-9_-]{1,60}$')
-_MODEL_FILE_NAME_PATTERN = re.compile(r'^[A-Za-z0-9_]{1,60}.tflite$')
 _GCS_TFLITE_URI_PATTERN = re.compile(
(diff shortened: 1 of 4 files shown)


firebase_admin/ml.py
```diff
@@ -541,16 +543,17 @@ def from_saved_model(cls, saved_model_dir, bucket_name=None, app=None):
         """
         TFLiteGCSModelSource._assert_tf_enabled()
         tflite_model = TFLiteGCSModelSource._tf_convert_from_saved_model(saved_model_dir)
-        open('firebase_ml_model.tflite', 'wb').write(tflite_model)
-        return TFLiteGCSModelSource.from_tflite_model_file(
-            'firebase_ml_model.tflite', bucket_name, app)
+        open(model_file_name, 'wb').write(tflite_model)
```
hiranya911: I missed this earlier. But this is leaking a file descriptor. Can you change this to:

```
with open(model_file_name, 'wb) as model_file:
    model_file.write(tflite_model)
```

firebase_admin/ml.py
```diff
@@ -563,9 +566,8 @@ def from_keras_model(cls, keras_model, bucket_name=None, app=None):
         """
         TFLiteGCSModelSource._assert_tf_enabled()
         tflite_model = TFLiteGCSModelSource._tf_convert_from_keras_model(keras_model)
-        open('firebase_ml_model.tflite', 'wb').write(tflite_model)
-        return TFLiteGCSModelSource.from_tflite_model_file(
-            'firebase_ml_model.tflite', bucket_name, app)
+        open(model_file_name, 'wb').write(tflite_model)
```
hiranya911: Here too.

PR #383, opened 2020-01-13
Add API to link/unlink provider info to/from user account.

nrsim:


integration/test_auth.py
```diff
@@ -288,6 +288,36 @@ def test_update_user(new_user):
...
+    assert user.phone_number == phone
+    assert len(user.provider_data) == 2
+
+    user = auth.update_user(
+        new_user.uid,
+            delete_provider_ids=['phone', 'google.com'])
+    assert user.uid == new_user.uid
+    assert user.phone_number is None
+    assert len(user.provider_data) == 1
+
```
rsgowman: A new requirement is that this API should also work with non-federated providers. (It's not something we're going to actually recommend, since there's more direct ways to do that.) Unfortunately, the backend for this API doesn't support non-federated providers, so we're going to have to fake that a bit.

Concretely, you've used update_user() to link to the 'google.com' provider. Could you also add calls for phone, email, oauth and saml? (Not necessarily in this exact test.) It'll require you to also change the implementation.

If this looks like it'll cause this PR to balloon in size, then let's talk first and see if we can come up with other options.

tests/test_user_mgt.py
```diff
@@ -488,6 +488,52 @@ def test_update_user_delete_fields(self, user_mgt_app):
...
+            'localId' : 'testuser',
+            'linkProviderUserInfo' : {
+                'rawId': 'test',
+                'providerId': 'google.com',
+                'email': 'test@example.com',
+                'displayName': 'Test Name',
+                'photoUrl': 'https://test.com/user.png'
+            }
+        }
+
```
rsgowman: Consider adding "failure" tests.
a) empty delete_providers_ids list
b) invalid provider_id. (I'm not sure we can tell if a provider_id is valid or not... other than checking for None or empty string.)
c) invalid provider_uid. (Same; just None/empty checks.)

These should all cause failures before hitting the network, so are easy to check from a unit test.

PR #386, opened 2020-01-14
Create a GitHub Actions based CI Pipeline

hiranya911:
Moved CI tasks to GitHub Actions. Kept the Travis CI job for testing on Py3.4, as that's not supported in GitHub Actions. Hopefully we can get rid of it completely once https://github.com/googleapis/google-cloud-python/issues/10136 is addressed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: Continuous Integration
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [3.4, 3.5, 3.6, 3.7, pypy3]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python }}
+      uses: actions/setup-python@v1
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Test with pytest
+      if: success() || failure()
+      run: pytest
+    - name: Set up Node.js 10
+      uses: actions/setup-node@v1
+      with:
+        node-version: 10.x
+    - name: Run integration tests against emulator
+      run: |
+        npm install -g firebase-tools
```
yuchenshi: I think we should use https://firebase.tools instead so that we can drop the setup-node action and build faster, but I think this is fine for now as well. Also, have we considered caching `~/.cache/firebase` which contains the emulator and (later) the standalone CLI?

hiranya911: Been experimenting with this, but firebase.tools absolutely refuses to work in the GitHub Actions environment (files a bug for the CLI for this). There are bunch of workarounds I can use to get this to work -- which involves pulling the CLI binary for Linux explicitly. But I'd rather do all that in a separate PR.

PR #388, opened 2020-01-15
Add API to look up users by provider user ID.

nrsim:


integration/test_auth.py
```diff
@@ -177,6 +177,27 @@ def new_user_with_params():
...
+        uid=random_id,
+        email=email,
+        phone_number=phone,
+        display_name='Random User',
+        photo_url='https://example.com/photo.png',
+        user_metadata=auth.UserMetadata(100, 150),
+        password_hash=b'password', password_salt=b'NaCl', custom_claims={'admin': True},
+        email_verified=True,
+        disabled=False,
+        provider_data=[auth.UserProvider(uid='test', provider_id='google.com')])
```
rsgowman: Same comment as before: uid=test might cause this to fail if the test doesn't get properly cleaned up.

integration/test_auth.py
```diff
@@ -177,6 +177,27 @@ def new_user_with_params():
...
+def imported_user_with_params():
+    random_id, email = _random_id()
+    phone = _random_phone()
+    user = auth.ImportUserRecord(
+        uid=random_id,
+        email=email,
+        phone_number=phone,
+        display_name='Random User',
+        photo_url='https://example.com/photo.png',
+        user_metadata=auth.UserMetadata(100, 150),
```
rsgowman: I couldn't figure out what 100,150 were without looking up UserMetadata. Named parameters would make this more obvious, i.e. 

`auth.UserMetadata(creation_timestamp=100, last_sign_in_timestamp=150)`

Optional.

PR #390, opened 2020-01-23
Firebase ML Modify Operation Handling Code to match rpc codes not html codes

ifielker:
Modify Operation handling code to match rpc codes not html codes.

### DIFF
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -58,7 +58,7 @@ _HTTP_STATUS_TO_ERROR_CODE = {
     503: exceptions.UNAVAILABLE,
 }
 
-
+# See https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
 _RPC_CODE_TO_ERROR_CODE = {
     1: exceptions.CANCELLED,
     2: exceptions.UNKNOWN,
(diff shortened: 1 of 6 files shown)


firebase_admin/_utils.py
```diff
@@ -120,9 +139,9 @@ def handle_operation_error(error):
             message='Unknown error while making a remote service call: {0}'.format(error),
             cause=error)
 
-    status_code = error.get('code')
+    rpc_code = error.get('code')
```
hiranya911: Doesn't this response contain a `status` field? That's how we map errors sent by other OnePlatform APIs. See `_handle_func_requests` function in this module.

https://cloud.google.com/apis/design/errors#http_mapping

ifielker: It does for non-operations responses. But for Operation errors it does not. Just message and code (details optional and not applicable in this case)

firebase_admin/_utils.py
```diff
@@ -59,6 +59,25 @@
 }
 
 
+_RPC_CODE_TO_ERROR_CODE = {
```
hiranya911: Nit: May be mention https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto in a comment so we know where these mappings come from.

ifielker: done

PR #391, opened 2020-01-23
Mlkit fix date handling2

ifielker:
Fixes create/update date handling to match what comes back from the actual backend.

### DIFF
--- a/firebase_admin/ml.py
+++ b/firebase_admin/ml.py
@@ -20,7 +20,6 @@ deleting, publishing and unpublishing Firebase ML models.
 
 
 import datetime
-import numbers
 import re
 import time
 import requests
@@ -246,20 +245,12 @@ class Model(object):
     @property
     def create_time(self):
         """The time the model was created."""
-        seconds = self._data.get('createTime', {}).get('seconds')
-        if not isinstance(seconds, numbers.Number):
-            return None
-
-        return datetime.datetime.fromtimestamp(float(seconds))
+        return self._data.get('createTime', None)
 
     @property
     def update_time(self):
         """The time the model was last updated."""
-        seconds = self._data.get('updateTime', {}).get('seconds')
-        if not isinstance(seconds, numbers.Number):
-            return None
-
-        return datetime.datetime.fromtimestamp(float(seconds))
+        return self._data.get('updateTime', None)
 
     @property
     def validation_error(self):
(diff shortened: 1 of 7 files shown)


firebase_admin/ml.py
```diff
@@ -246,20 +245,12 @@ def display_name(self, display_name):
     @property
     def create_time(self):
         """The time the model was created."""
-        seconds = self._data.get('createTime', {}).get('seconds')
-        if not isinstance(seconds, numbers.Number):
-            return None
-
-        return datetime.datetime.fromtimestamp(float(seconds))
+        return self._data.get('createTime', None)
```
hiranya911: We generally expose timestamps as milliseconds since epoch values in Python. We might want to retain that behavior here as well. String timestamps are kind of awkward in Python (since it requires the user to handle the timezone and the format of the string).

https://github.com/firebase/firebase-admin-python/blob/00cf1d34fdd8ca3531ff3556688c2e785db17920/firebase_admin/_user_mgt.py#L51-L66

What did we agree to in the API review for this?

ifielker: Millis. I changed it to do the conversion. Thanks.

firebase_admin/ml.py
```diff
@@ -243,23 +242,25 @@ def display_name(self, display_name):
         self._data['displayName'] = _validate_display_name(display_name)
         return self
 
+    @staticmethod
+    def _convert_to_millis(date_string):
+        if date_string is None:
```
hiranya911: Nit: `if not date_string:`

tests/test_ml.py
```diff
@@ -130,8 +120,8 @@
 CREATED_UPDATED_MODEL_JSON_1 = {
     'name': MODEL_NAME_1,
     'displayName': DISPLAY_NAME_1,
-    'createTime': CREATE_TIME_JSON,
-    'updateTime': UPDATE_TIME_JSON,
+    'createTime': CREATE_TIME,
```
hiranya911: Can these 2 values ever be not set in response? At least the parsing logic assumes so. Might want to add a test case for it.

ifielker: They will always be set in the response.

firebase_admin/ml.py
ifielker: done

PR #392, opened 2020-01-23
Firebase Ml Fix upload file naming

ifielker:
TFLite upload to GCS bucket works great from CoLab as it was previously. That's because all the tflite files I wanted to upload were in the root of the Colab runtime file list. That might not be true in scripting circumstances (like integration tests for example). Now when you upload file "/path/to/my/model.tflite" it gets saved as Firebase/ML/Models/model.tflite instead of Firebase/ML/Models/path/to/your/model.tflite.

### DIFF
--- a/firebase_admin/ml.py
+++ b/firebase_admin/ml.py
@@ -430,8 +430,18 @@ class _CloudStorageClient(object):
     def upload(bucket_name, model_file_name, app):
         """Upload a model file to the specified Storage bucket."""
         _CloudStorageClient._assert_gcs_enabled()
+
+        # Calculate the destination file_name (remove path if present)
+        file_name = model_file_name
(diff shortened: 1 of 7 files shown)


firebase_admin/ml.py
```diff
@@ -440,8 +440,18 @@ def _parse_gcs_tflite_uri(uri):
     def upload(bucket_name, model_file_name, app):
         """Upload a model file to the specified Storage bucket."""
         _CloudStorageClient._assert_gcs_enabled()
+
+        # Calculate the destination file_name (remove path if present)
+        file_name = model_file_name
```
hiranya911: This whole thing can be condensed to:

```
file_name = os.path.basename(model_file_name)
```

That should work in Windows environments too (that's probably not a requirement, but just in case)

ifielker: Nice! done.

PR #395, opened 2020-01-29
Using a default HTTP timeout in db module

hiranya911:
The `google-auth` package now sets a default timeout of 120 seconds for all outgoing requests (see https://github.com/googleapis/google-auth-library-python/pull/435). This is causing a couple of test failures in our codebase: https://github.com/firebase/firebase-admin-python/actions/runs/32412728

But more importantly this change leads to unpredictable default behavior in applications that use `firebase_admin`. Specifically, developers who are on older versions of `google-auth` won't get any timeout (`timeout=None`), while developers on new versions will get a 2 minutes timeout (`timeout=120`).

This PR addresses the above discrepancy by setting a default timeout in our own code. This has the following advantages:
1. Developers using `firebase_admin` will see consistent behavior regardless of their `google-auth` version.
2. We get more control over the default timeout, and are no longer affected by the changes in `google-auth`.

Note that `httpTimeout` option is currently only respected by a subset of our modules. Other modules (e.g. auth), will get their default timeout from `google-auth` for now. I'm using the same default value of 120 seconds to mitigate this, but we should properly address this for all modules in a subsequent PR.

### DIFF
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -59,6 +59,9 @@ _HTTP_STATUS_TO_ERROR_CODE = {
 }
 
 
+DEFAULT_HTTP_TIMEOUT_SECONDS = 120
+
+
 def _get_initialized_app(app):
     """Returns a reference to an initialized App instance."""
     if app is None:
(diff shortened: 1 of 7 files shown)


hiranya911:
Closing in favor of #397

PR #396, opened 2020-01-29
Fix broken test (which relied on default timeout being None).

rsgowman:
The upstream behaviour has changed such that None is no longer the
default:

https://github.com/googleapis/google-auth-library-python/pull/435


rsgowman:
I don't know if there are other implications to this value changing...?

rsgowman:
Oh, never mind; I just noticed https://github.com/firebase/firebase-admin-python/pull/395

PR #397, opened 2020-01-29
fix: Setting a default timeout on all HTTP connections

hiranya911:
The `google-auth` package now sets a default timeout of 120 seconds for all outgoing requests (see https://github.com/googleapis/google-auth-library-python/pull/435). This is causing a couple of test failures in our codebase: https://github.com/firebase/firebase-admin-python/actions/runs/32412728

But more importantly this change leads to unpredictable default behavior in applications that use `firebase_admin`. Specifically, developers who are on older versions of `google-auth` won't get any timeout (`timeout=None`), while developers on new versions will get a 2 minutes timeout (`timeout=120`).

This PR addresses the above discrepancy by setting a default timeout in our own code. This has the following advantages:
1. Developers using `firebase_admin` will see consistent behavior regardless of their `google-auth` version.
2. We get more control over the default timeout, and are no longer affected by the changes in `google-auth`.

Note that `httpTimeout` option is currently only respected by a subset of our modules. And there's some code duplication in how various modules read the `httpTimeout` option. We can address this in a separate PR.

RELEASE NOTE: Admin SDK now sets a default timeout of 120 seconds on all outgoing HTTP requests.

### DIFF
--- a/firebase_admin/__init__.py
+++ b/firebase_admin/__init__.py
@@ -49,8 +49,8 @@ def initialize_app(credential=None, options=None, name=_DEFAULT_APP_NAME):
           Google Application Default Credentials are used.
       options: A dictionary of configuration options (optional). Supported options include
           ``databaseURL``, ``storageBucket``, ``projectId``, ``databaseAuthVariableOverride``,
-          ``serviceAccountId`` and ``httpTimeout``. If ``httpTimeout`` is not set, HTTP
-          connections initiated by client modules such as ``db`` will not time out.
+          ``serviceAccountId`` and ``httpTimeout``. If ``httpTimeout`` is not set, the SDK
+          uses a default timeout of 120 seconds.
       name: Name of the app (optional).
     Returns:
       App: A newly initialized instance of App.
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -775,7 +775,7 @@ class _DatabaseService:
             self._auth_override = json.dumps(auth_override, separators=(',', ':'))
         else:
             self._auth_override = None
-        self._timeout = app.options.get('httpTimeout')
+        self._timeout = app.options.get('httpTimeout', _http_client.DEFAULT_TIMEOUT_SECONDS)
         self._clients = {}
 
         emulator_host = os.environ.get(_EMULATOR_HOST_ENV_VAR)
@@ -905,9 +905,8 @@ class _Client(_http_client.JsonHttpClient):
           params: Dict of query parameters to add to all outgoing requests.
         """
         _http_client.JsonHttpClient.__init__(
-            self, credential=credential, base_url=base_url, headers={'User-Agent': _USER_AGENT})
-        self.credential = credential
-        self.timeout = timeout
+            self, credential=credential, base_url=base_url, timeout=timeout,
+            headers={'User-Agent': _USER_AGENT})
         self.params = params if params else {}
 
     def request(self, method, url, **kwargs):
@@ -937,8 +936,6 @@ class _Client(_http_client.JsonHttpClient):
                 query = extra_params
         kwargs['params'] = query
 
-        if self.timeout:
-            kwargs['timeout'] = self.timeout
         try:
             return super(_Client, self).request(method, url, **kwargs)
         except requests.exceptions.RequestException as error:
--- a/firebase_admin/db.py
+++ b/firebase_admin/db.py
@@ -900,13 +900,13 @@ class _Client(_http_client.JsonHttpClient):
           credential: A Google credential that can be used to authenticate requests.
           base_url: A URL prefix to be added to all outgoing requests. This is typically the
               Firebase Realtime Database URL.
-          timeout: HTTP request timeout in seconds. If not set connections will never
+          timeout: HTTP request timeout in seconds. If set to None connections will never
               timeout, which is the default behavior of the underlying requests library.
           params: Dict of query parameters to add to all outgoing requests.
         """
-        _http_client.JsonHttpClient.__init__(
-            self, credential=credential, base_url=base_url, timeout=timeout,
-            headers={'User-Agent': _USER_AGENT})
+        super().__init__(
+            credential=credential, base_url=base_url,
+            timeout=timeout, headers={'User-Agent': _USER_AGENT})
         self.params = params if params else {}
 
     def request(self, method, url, **kwargs):
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -774,7 +774,6 @@ class TestDatabaseInitialization:
         default_ref = db.reference()
         other_ref = db.reference(url='https://other.firebaseio.com')
         for ref in [default_ref, other_ref]:
-            assert ref._client.timeout == _http_client.DEFAULT_TIMEOUT_SECONDS
             if override == {}:
                 assert 'auth_variable_override' not in ref._client.params
             else:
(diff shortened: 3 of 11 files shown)


tests/test_messaging.py
```diff
@@ -1641,7 +1642,8 @@ def test_send(self):
         assert recorder[0].url == self._get_url('explicit-project-id')
         assert recorder[0].headers['X-GOOG-API-FORMAT-VERSION'] == '2'
         assert recorder[0].headers['X-FIREBASE-CLIENT'] == self._CLIENT_VERSION
-        assert recorder[0]._extra_kwargs['timeout'] is None
+        assert recorder[0]._extra_kwargs['timeout'] == pytest.approx(
```
rsgowman: (again, though if you remove these assertions you should probably add a `test_timeout` test.)

tests/test_project_management.py
```diff
@@ -522,6 +523,8 @@ def _assert_request_is_correct(
         assert request.url == expected_url
         client_version = 'Python/Admin/{0}'.format(firebase_admin.__version__)
         assert request.headers['X-Client-Version'] == client_version
+        assert request._extra_kwargs['timeout'] == pytest.approx(
```
rsgowman: (again)

tests/test_user_mgt.py
```diff
@@ -102,12 +103,18 @@ def _check_user_record(user, expected_uid='testuser'):
 
 class TestAuthServiceInitialization:
 
+    def test_default_timeout(self, user_mgt_app):
+        auth_service = auth._get_auth_service(user_mgt_app)
+        user_manager = auth_service.user_manager
+        assert user_manager._client.timeout == _http_client.DEFAULT_TIMEOUT_SECONDS
```
rsgowman: Hm... It seems like all of our components will have an approximately identical test like this one. We could remove them all and just add this into the _http_client tests. Or we could add a parameterized test that verifies that each component sets the timeout appropriately. (No action required; I'm just thinking out loud.)

firebase_admin/__init__.py
```diff
@@ -49,8 +49,8 @@ def initialize_app(credential=None, options=None, name=_DEFAULT_APP_NAME):
           Google Application Default Credentials are used.
       options: A dictionary of configuration options (optional). Supported options include
           ``databaseURL``, ``storageBucket``, ``projectId``, ``databaseAuthVariableOverride``,
-          ``serviceAccountId`` and ``httpTimeout``. If ``httpTimeout`` is not set, HTTP
-          connections initiated by client modules such as ``db`` will not time out.
+          ``serviceAccountId`` and ``httpTimeout``. If ``httpTimeout`` is not set, the SDK
+          uses a default timeout of 120 seconds.
```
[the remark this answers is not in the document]
hiranya911: `google-auth` shipped the same change as minor version bump. Given our next version is going to be a major version bump this should be ok. And yes, I also prefer to consider this a bug fix than anything else.

tests/test_user_mgt.py
hiranya911: Yes, and I think the existing test cases in _http_client already provides us sufficient coverage. In each of the other modules it's probably enough to check that the Client.timeout is correctly set. We can do that clean up in a future iteration.

firebase_admin/db.py
```diff
@@ -905,9 +905,8 @@ def __init__(self, credential, base_url, timeout, params=None):
           params: Dict of query parameters to add to all outgoing requests.
         """
         _http_client.JsonHttpClient.__init__(
-            self, credential=credential, base_url=base_url, headers={'User-Agent': _USER_AGENT})
-        self.credential = credential
-        self.timeout = timeout
+            self, credential=credential, base_url=base_url, timeout=timeout,
```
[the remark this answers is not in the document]
hiranya911: Done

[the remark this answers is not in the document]
hiranya911: Done

tests/test_db.py
```diff
@@ -736,10 +737,11 @@ def test_valid_db_url(self, url):
         ref._client.session.mount(url, adapter)
         assert ref._client.base_url == 'https://test.firebaseio.com'
         assert 'auth_variable_override' not in ref._client.params
-        assert ref._client.timeout is None
+        assert ref._client.timeout == _http_client.DEFAULT_TIMEOUT_SECONDS
```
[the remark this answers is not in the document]
hiranya911: Moved timeout checks into their own test cases

tests/test_http_client.py
```diff
@@ -73,6 +83,27 @@ def test_credential():
...
+    assert client.base_url == ''
+    assert client.timeout == timeout
+    recorder = _instrument(client, 'body')
+    resp = client.request('get', _TEST_URL)
+    assert resp.status_code == 200
+    assert resp.text == 'body'
+    assert len(recorder) == 1
+    assert recorder[0].method == 'GET'
+    assert recorder[0].url == _TEST_URL
+    if timeout:
```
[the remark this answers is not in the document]
hiranya911: Done and done

tests/test_http_client.py
```diff
@@ -73,6 +83,27 @@ def test_credential():
     assert recorder[0].method == 'GET'
     assert recorder[0].url == _TEST_URL
     assert recorder[0].headers['Authorization'] == 'Bearer mock-token'
+    assert recorder[0]._extra_kwargs['timeout'] == pytest.approx(
+        _http_client.DEFAULT_TIMEOUT_SECONDS, 0.001)
+
+@pytest.mark.parametrize('timeout', [7, None])
+def test_timeout(timeout):
+    client = _http_client.HttpClient(timeout=timeout)
+    assert client.session is not None
```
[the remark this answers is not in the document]
hiranya911: Done

tests/test_http_client.py
```diff
@@ -28,43 +28,53 @@ def test_http_client_default_session():
     client = _http_client.HttpClient()
     assert client.session is not None
     assert client.base_url == ''
+    assert client.timeout == _http_client.DEFAULT_TIMEOUT_SECONDS
```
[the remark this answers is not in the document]
hiranya911: Done

tests/test_messaging.py
hiranya911: Moved to TestTimeout class

tests/test_project_management.py
hiranya911: Done

firebase_admin/db.py
```diff
@@ -900,13 +900,13 @@ def __init__(self, credential, base_url, timeout, params=None):
           credential: A Google credential that can be used to authenticate requests.
           base_url: A URL prefix to be added to all outgoing requests. This is typically the
               Firebase Realtime Database URL.
-          timeout: HTTP request timeout in seconds. If not set connections will never
+          timeout: HTTP request timeout in seconds. If set to None connections will never
```
rsgowman: nit: the 'which is the default behaviour ...' part is no longer true.

tests/test_http_client.py
```diff
@@ -83,26 +73,27 @@ def test_credential():
...
+    recorder = _instrument(client, 'body')
+    client.request('get', _TEST_URL)
+    assert len(recorder) == 1
     assert recorder[0]._extra_kwargs['timeout'] == pytest.approx(
         _http_client.DEFAULT_TIMEOUT_SECONDS, 0.001)
 
-@pytest.mark.parametrize('timeout', [7, None])
-def test_timeout(timeout):
+@pytest.mark.parametrize('timeout', [7, 0, None])
+def test_custom_timeout(timeout):
```
rsgowman: This function looks mostly identical to test_default_timeout. You could optionally refactor to something like this:

```python
def _helper_test_timeout(**kwargs):
    client = _http_client.HttpClient()
    timeout = kwargs.get('timeout', _http_client.DEFAULT_TIMEOUT_SECONDS)
    assert client.timeout == timeout
    ...
    if timeout is None
        assert recorder[0]._extra_kwargs['timeout'] is None
    else:
        assert ... == pytest.approx(timeout, 0.001)

def test_default_timeout():
    _helper_test_timeout()

@parameterize(...)
def test_custom_timeout(timeout)
    _helper_test_timeout(timeout=timeout)
```

(Or you could parameterize the helper directly.)

tests/test_project_management.py
```diff
@@ -523,14 +523,33 @@ def _assert_request_is_correct(
...
+        options = {
+            'httpTimeout': timeout,
+            'projectId': 'test-project-id'
+        }
+        app = firebase_admin.initialize_app(testutils.MockCredential(), options, 'timeout-app')
+        try:
+            project_management_service = project_management._get_project_management_service(app)
+            assert project_management_service._client.timeout == timeout
+        finally:
+            firebase_admin.delete_app(app)
```
rsgowman: BaseProjectManagementTest seems to call testutils.cleanup_apps() in it's teardown() method. If you're willing to rely on that, you could drop the try, finally, and delete_app() lines. Presumably you'd nave to not hard-code the app name ("timeout-app") but instead use something based on the test parameter or something like that instead. Optional.

I have no objection to leaving it in either, but it's odd to see the app cleaned up in this test case, but not the one above.

firebase_admin/db.py
hiranya911:
```
If set to None connections will never
timeout, which is the default behavior of the underlying requests library.
```

That is still true. The `requests` library behavior hasn't yet changed. The default `timeout=None` behavior still comes from there.

tests/test_http_client.py
hiranya911: Refactored with parameterization.

tests/test_project_management.py
hiranya911: Done

firebase_admin/db.py
rsgowman: Oops; yeah I missed the change from 'if unset' to 'if set to none' (despite suggesting exactly that earlier. 😳)

PR #400, opened 2020-01-30
feat(auth): Add bulk get/delete methods

rsgowman:
This PR allows callers to retrieve a list of users by unique identifier (uid, email, phone, federated provider uid) as well as to delete a list of users.

RELEASE NOTE: Added `get_users()` and `delete_users()` APIs for retrieving and deleting user accounts in bulk.


integration/test_auth.py
```diff
@@ -17,6 +17,7 @@
 import datetime
 import random
 import time
+from typing import List
```
rsgowman: I used a few type annotations in the test. They're 3.5, so I *think* they're fair game? I've not tried to use them throughout or anything, and I've avoided the non-test files completely. (Basically, I just added a few so that I could get tab completion. :)

integration/test_auth.py
```diff
@@ -219,6 +237,78 @@ def test_get_user(new_user_with_params):
...
+class TestGetUsers:
+    @staticmethod
+    def _map_user_record_to_uid_email_phones(user_record):
+        return {
+            'uid': user_record.uid,
+            'email': user_record.email,
+            'phone_number': user_record.phone_number
+        }
+
+    def test_returns_users_by_various_identifier_types_in_a_single_call(self, new_user_record_list):
```
hiranya911: Use shorter names to be consistent with the rest of this file?

`test_multiple_uid_types():`

integration/test_auth.py
```diff
@@ -219,6 +237,78 @@ def test_get_user(new_user_with_params):
...
+    _sign_in_with_password(new_user_with_params.email, 'secret', api_key)
+    new_user_with_params = auth.get_user(new_user_with_params.uid)
+
+    # Ensure the last refresh time occurred at approximately 'now'. (With a
+    # tolerance of up to 1 minute; we ideally want to ensure that any timezone
+    # considerations are handled properly, so as long as we're within an hour,
+    # we're in good shape.)
+    millis_per_second = 1000
+    seconds_per_minute = 60
+    millis_per_minute = millis_per_second * seconds_per_minute
```
hiranya911: Can probably inline seconds_per_minute here.

tests/test_user_mgt.py
```diff
@@ -301,6 +301,47 @@ def test_get_user_by_phone_http_error(self, user_mgt_app):
         assert excinfo.value.cause is not None
 
 
+class TestGetUsers:
+
+    def test_should_reject_when_given_more_than_100_identifiers(self, user_mgt_app):
```
hiranya911: Use short names like the rest of this file.

tests/test_user_mgt.py
```diff
@@ -301,6 +301,47 @@ def test_get_user_by_phone_http_error(self, user_mgt_app):
...
+        identifiers = [
+            auth.UidIdentifier('valid_id1'),
+            auth.UidIdentifier('valid_id2'),
+            auth.UidIdentifier('invalid id; too long. ' + '.'*128),
+            auth.UidIdentifier('valid_id4'),
+            auth.UidIdentifier('valid_id5')]
+
+        with pytest.raises(ValueError):
+            auth.get_users(identifiers, app=user_mgt_app)
+
```
hiranya911: Add a success case test.

tests/test_user_mgt.py
```diff
@@ -612,6 +653,47 @@ def test_delete_user_unexpected_response(self, user_mgt_app):
         assert isinstance(excinfo.value, exceptions.UnknownError)
 
 
+class TestDeleteUsers:
+
```
hiranya911: Success case test

firebase_admin/_user_mgt.py
```diff
@@ -19,8 +19,10 @@
 from urllib import parse
 
 import requests
+import iso8601
```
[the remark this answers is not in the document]
rsgowman: I went (mostly) with 1.

Looking at the spec a bit closer, it's actually restricted to RFC3339 (effectively a subset of iso8601), which is a bit easier. I suspect my implementation may be incomplete wrt rfc3339, but I think it's at least real close to the spec as specified by timestamp.proto.

Still, it feels awful to (re)implement this kind of functionality. :(

requirements.txt
```diff
@@ -8,3 +8,4 @@ google-api-core[grpc] >= 1.14.0, < 2.0.0dev; platform.python_implementation != '
 google-api-python-client >= 1.7.8
 google-cloud-firestore >= 1.4.0; platform.python_implementation != 'PyPy'
 google-cloud-storage >= 1.18.0
+iso8601 >= 0.1.12
```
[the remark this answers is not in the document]
rsgowman: (see other comment / done)

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,89 @@
...
+        pass
+
+    # Note: %z parses timezone offsets, but requires the timezone offset *not*
+    # include a separating ':'. As of python 3.7, this was relaxed.
+    # TODO(rsgowman): Once python3.7 becomes our floor, we can drop the regex
+    # replacement.
+    datestr_modified = re.sub(r'(\d\d):(\d\d)$', r'\1\2', datestr_modified)
+
+    try:
+        print("trying with micros")
```
hiranya911: Drop the print statements.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +351,85 @@ def iterate_all(self):
...
+    @property
+    def errors(self):
+        """A list of ``auth.BatchDeleteErrorInfo`` instances describing the
+        errors that were encountered during the deletion. Length of this list
+        is equal to `failure_count`.
+        """
+        return self._errors
+
+
+class BatchDeleteErrorInfo:
```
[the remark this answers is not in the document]
hiranya911: Sounds good.

As for using `@property` our convention is to use it for immutable properties (i.e. setters not allowed). So it's usage here is correct.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +378,97 @@ def iterate_all(self):
...
+    @property
+    def index(self):
+        return self._index
+
+    @property
+    def reason(self):
+        return self._reason
+
+    @property
+    def local_id(self):
```
hiranya911: It seems this type wasn't really mentioned in the API proposal. Should we use `uid` here instead of `local_id`? We haven't exposed `local_id` terminology anywhere else.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +378,97 @@ def iterate_all(self):
...
+        """Constructs a BatchDeleteAccountsResponse instance, corresponseing to
+        the json representing the BatchDeleteAccountsResponse proto.
+
+        Args:
+            errors: List of dictionaries, with each dictionary representing a
+                BatchDeleteErrorInfo instance as returned by the server. None
+                implies an empty list.
+        """
+        if errors:
+            self.errors = [BatchDeleteErrorInfo(err) for err in errors]
```
hiranya911:
```
self.errors = [BatchDeleteErrorInfo(err) for err in errors] if errors else []
```

firebase_admin/auth.py
```diff
@@ -84,11 +93,13 @@
 ActionCodeSettings = _user_mgt.ActionCodeSettings
 CertificateFetchError = _token_gen.CertificateFetchError
 DELETE_ATTRIBUTE = _user_mgt.DELETE_ATTRIBUTE
+DeleteUsersResult = _user_mgt.DeleteUsersResult
```
hiranya911: Also BatchDeleteErrorInfo

firebase_admin/_rfc3339.py
rsgowman: Oops.

integration/test_auth.py
```diff
@@ -219,6 +237,78 @@ def test_get_user(new_user_with_params):
...
+            'email': user_record.email,
+            'phone_number': user_record.phone_number
+        }
+
+    def test_returns_users_by_various_identifier_types_in_a_single_call(self, new_user_record_list):
+        users = auth.get_users([
+            auth.UidIdentifier(new_user_record_list[0].uid),
+            auth.EmailIdentifier(new_user_record_list[1].email),
+            auth.PhoneIdentifier(new_user_record_list[2].phone_number)])
+        actual = sorted(
```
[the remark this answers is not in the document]
hiranya911: Oh yes. Lambdas in `key=` is totally fine.

hiranya911:
Adding @egilmorez to review the documentation bits in `auth.py`

firebase_admin/_user_mgt.py
rsgowman: uid is better than local_id. Done.

> API proposal.

Hm, right. It's listed as ErrorInfo there, which per docs on ErrorInfo, makes it inconvenient to do without a breaking change. But it's *only* the docs; we could instead change them to read "Represents an error encountered while importing an ImportUserRecord *or deleting multiple users*". We'd then also have to eliminate uid/local_id. Which wouldn't be terrible; user's could recover it from the index property combined with their input list.

firebase_admin/_user_mgt.py
rsgowman: Done.

firebase_admin/auth.py
rsgowman: Done. (Also BatchDeleteAccountsResponse)

hiranya911: `BatchDeleteAccountsResponse` does not belong in the public API if I understood correct. `BatchDeleteErrorInfo` is exposed via `DeleteUsersResult.errors` so only that's needed.

firebase_admin/_user_mgt.py
rsgowman: Ok, I've switched to ErrorInfo.

firebase_admin/_user_import.py
```diff
@@ -472,11 +472,16 @@ def standard_scrypt(cls, memory_cost, parallelization, block_size, derived_key_l
 
 
 class ErrorInfo:
-    """Represents an error encountered while importing an ``ImportUserRecord``."""
+    """Represents an error encountered while performing a batch operation such
+    as importing users or deleting multiple user accounts.
+    """
+    # TODO(rsgowman): This class used to be specific to importing users (hence
```
hiranya911: _auth_utils could be a good place. But feel free to defer that to future PR.

firebase_admin/_user_import.py
```diff
@@ -472,11 +472,16 @@ def standard_scrypt(cls, memory_cost, parallelization, block_size, derived_key_l
...
+    # it's home in _user_import.py). It's now also used by bulk deletion of
+    # users. Move this to a more common location.
 
     def __init__(self, error):
-        self._index = error['index']
-        self._reason = error['message']
+        self._index = error.get('index', 0)
```
hiranya911: Defaulting to index 0 sounds risky? Does the backend actually ever report an error without index? Seems safer to error out in this case than trying to sweep it under the rug.

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,87 @@
...
+
+def parse_to_epoch(datestr):
+    """Parses an RFC3339 date string and return the number of seconds since the
+    epoch (as a float).
+
+    In particular, this method is meant to parse the strings returned by the
+    JSON mapping of protobuf google.protobuf.timestamp.Timestamp instances:
+    https://github.com/protocolbuffers/protobuf/blob/4cf5bfee9546101d98754d23ff378ff718ba8438/src/google/protobuf/timestamp.proto#L99
+
+    This method has microsecond precision; i.e. nanoseconds will be truncated.
```
egilmorez: I think "i.e." can be omitted without losing the sense.

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Args:
+        datestr: A string in RFC3339 format.
+    Returns:
+        Float: The number of seconds since the unix epoch.
```
egilmorez: I see this as "Unix epoch" on wikipedia, without all caps, but with a leading cap.

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,87 @@
...
+    Raises:
+        ValueError: Raised if the datestr is not a valid RFC3339 date string.
```
egilmorez: Suggest backticks for code font/literal.

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,87 @@
...
+
+def _parse_to_datetime(datestr):
+    """Parse an RFC3339 date string and return a python datetime instance.
+
+    Args:
+        datestr: A string in RFC3339 format.
+    Returns:
+        datetime: The corresponding datetime (with timezone information).
+    Raises:
+        ValueError: Raised if the datestr is not a valid RFC3339 date string.
```
egilmorez: Suggest backticks for code font/literal.

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,87 @@
...
+    return _parse_to_datetime(datestr).timestamp()
+
+
+def _parse_to_datetime(datestr):
+    """Parse an RFC3339 date string and return a python datetime instance.
+
+    Args:
+        datestr: A string in RFC3339 format.
+    Returns:
+        datetime: The corresponding datetime (with timezone information).
```
egilmorez: Suggest backticks for code font/literal.

firebase_admin/_user_mgt.py
```diff
@@ -65,6 +71,16 @@ def last_sign_in_timestamp(self):
         """
         return self._last_sign_in_timestamp
 
+    @property
+    def last_refresh_timestamp(self):
+        """The time at which the user was last active (ID token refreshed).
+
+        Returns:
+          integer: Milliseconds since epoch timestamp, or None if the user was
```
egilmorez: Does it literally return the string "None?"  The cap makes me thinks so . . . and if so this should be backticked probably.

firebase_admin/_rfc3339.py
```diff
@@ -0,0 +1,87 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Parse RFC3339 date strings"""
+
+from datetime import datetime, timezone
+import re
+
+def parse_to_epoch(datestr):
+    """Parses an RFC3339 date string and return the number of seconds since the
```
egilmorez: Suggest "returns"

firebase_admin/_user_identifier.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class EmailIdentifier(UserIdentifier):
+    """Used for looking up an account by email.
+
+    See ``auth.get_user()``.
+    """
+
+    def __init__(self, email):
+        """Constructs a new EmailIdentifier.
```
egilmorez: Suggest backticks for code font/literal.

firebase_admin/_user_identifier.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class ProviderIdentifier(UserIdentifier):
+    """Used for looking up an account by provider.
+
+    See ``auth.get_user()``.
+    """
+
+    def __init__(self, provider_id, provider_uid):
+        """Constructs a new ProviderIdentifier.
```
egilmorez: Suggest backticks for code font/literal.  Maybe even `ProviderIdentifier` object?

firebase_admin/_user_mgt.py
```diff
@@ -280,6 +301,33 @@ def password_salt(self):
...
+        Args:
+            users: List of `UserRecord` instances.
+            not_found: List of `UserIdentifier` instances.
+        """
+        self._users = users
+        self._not_found = not_found
+
+    @property
+    def users(self):
+        """Set of UserRecords, corresponding to the set of users that were
```
egilmorez: Should this be backticked ``UserRecord`` objects?

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
         return _UserIterator(self)
 
 
+class DeleteUsersResult:
+    """Represents the result of the ``auth.delete_users()`` API."""
+
+    def __init__(self, result, total):
+        """Constructs a DeleteUsersResult.
```
egilmorez: Suggest backticks for literal.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
...
+
+        Args:
+          result: BatchDeleteAccountsResponse: The proto response, wrapped in a
+              BatchDeleteAccountsResponse instance.
```
egilmorez: Suggest backticks for literal.

Is the result after the colon above code-fonted as well?

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
...
+        self._success_count = total - len(errors)
+        self._failure_count = len(errors)
+        self._errors = errors
+
+    @property
+    def success_count(self):
+        """Returns the number of users that were deleted successfully (possibly
+        zero).
+
+        Users that did not exist prior to calling delete_users() will be
```
egilmorez: Suggest backticks for literal, and "are considered to be successfully deleted."

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
...
+    def errors(self):
+        """A list of ``auth.ErrorInfo`` instances describing the errors that
+        were encountered during the deletion. Length of this list is equal to
+        `failure_count`.
+        """
+        return self._errors
+
+
+class BatchDeleteAccountsResponse:
+    """Represents the results of a delete_users() call."""
```
egilmorez: Suggest backticks for literal.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
...
+
+    def __init__(self, errors=None):
+        """Constructs a BatchDeleteAccountsResponse instance, corresponseing to
+        the json representing the BatchDeleteAccountsResponse proto.
```
egilmorez: Suggest "JSON"

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
...
+        `failure_count`.
+        """
+        return self._errors
+
+
+class BatchDeleteAccountsResponse:
+    """Represents the results of a delete_users() call."""
+
+    def __init__(self, errors=None):
+        """Constructs a BatchDeleteAccountsResponse instance, corresponseing to
```
egilmorez: Suggest backticks for literal, here and just below.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +379,63 @@ def iterate_all(self):
...
+        the json representing the BatchDeleteAccountsResponse proto.
+
+        Args:
+            errors: List of dictionaries, with each dictionary representing a
+                ErrorInfo instance as returned by the server. None implies an
```
egilmorez: Suggest backticks for literal, and "an ``ErrorInfo`` instance..."

firebase_admin/_user_mgt.py
```diff
@@ -483,6 +588,62 @@ def get_user(self, **kwargs):
...
+    def get_users(self, identifiers):
+        """Looks up multiple users by their identifiers (uid, email, etc.)
+
+        Args:
+            identifiers: UserIdentifier[]: The identifiers indicating the user
+                to be looked up. Must have <= 100 entries.
+
+        Returns:
+            list[dict[string, string]]: List of dicts representing the json
+            UserInfo responses from the server.
```
egilmorez: Suggest backticks for literal.

firebase_admin/auth.py
```diff
@@ -309,6 +326,60 @@ def get_user_by_phone_number(phone_number, app=None):
...
 
 
+def get_users(identifiers, app=None):
+    """Gets the user data corresponding to the specified identifiers.
+
+    There are no ordering guarantees; in particular, the nth entry in the
+    result list is not guaranteed to correspond to the nth entry in the input
+    parameters list.
+
+    Only a maximum of 100 identifiers may be supplied. If more than 100
```
egilmorez: Suggest "A maximum . . . "

firebase_admin/auth.py
```diff
@@ -309,6 +326,60 @@ def get_user_by_phone_number(phone_number, app=None):
...
+    identifiers are supplied, this method will immediately raise a ValueError.
```
egilmorez: Suggest "method raises a ``ValueError``.

firebase_admin/auth.py
```diff
@@ -446,6 +517,36 @@ def delete_user(uid, app=None):
...
+def delete_users(uids, app=None):
+    """Deletes the users specified by the given identifiers.
+
+    Deleting a non-existing user won't generate an error. (i.e. this method is
+    idempotent.) Non-existing users will be considered to be successfully
+    deleted, and will therefore be counted in the
+    DeleteUserResult.success_count value.
+
+    Only a maximum of 1000 identifiers may be supplied. If more than 1000
+    identifiers are supplied, this method will immediately raise a ValueError.
```
egilmorez: Suggest "method raises a ``ValueError``.

firebase_admin/auth.py
```diff
@@ -446,6 +517,36 @@ def delete_user(uid, app=None):
     user_manager.delete_user(uid)
 
 
+def delete_users(uids, app=None):
+    """Deletes the users specified by the given identifiers.
+
+    Deleting a non-existing user won't generate an error. (i.e. this method is
+    idempotent.) Non-existing users will be considered to be successfully
```
egilmorez: Suggest "are considered"

firebase_admin/_user_import.py
rsgowman: Yeah, good point. Done.

firebase_admin/_rfc3339.py
rsgowman: Done.

firebase_admin/_rfc3339.py
rsgowman: Done.

firebase_admin/_rfc3339.py
rsgowman: Done.

firebase_admin/_rfc3339.py
rsgowman: Done.

firebase_admin/_rfc3339.py
rsgowman: Done.

firebase_admin/_rfc3339.py
rsgowman: Done.

firebase_admin/_user_identifier.py
rsgowman: Done.

firebase_admin/_user_identifier.py
rsgowman: Done/done (throughout).

firebase_admin/_user_mgt.py
rsgowman: No, it returns the python `None` value (roughly corresponding to null in other languages). But backticks seem appropriate regardless (done).

firebase_admin/_user_mgt.py
rsgowman: Done (here and immediately below).

firebase_admin/_user_mgt.py
rsgowman: Done.

firebase_admin/_user_mgt.py
rsgowman: > backticks for literal

Done.

> after the colon

Hm. I'm not sure that should even be there. I've removed it and rephrased it slightly.

firebase_admin/_user_mgt.py
rsgowman: Done/done.

firebase_admin/_user_mgt.py
rsgowman: Done.

firebase_admin/_user_mgt.py
rsgowman: Done (also fixed spelling typo)

firebase_admin/_user_mgt.py
rsgowman: Done.

firebase_admin/_user_mgt.py
rsgowman: Done/done.

firebase_admin/_user_mgt.py
rsgowman: Done. (Also json->JSON)

firebase_admin/auth.py
rsgowman: Done.

firebase_admin/auth.py
rsgowman: Done.

firebase_admin/auth.py
rsgowman: Done.

firebase_admin/auth.py
rsgowman: Done.

firebase_admin/_user_mgt.py
```diff
@@ -331,6 +381,63 @@ def iterate_all(self):
...
+    @property
+    def failure_count(self):
+        """Returns the number of users that failed to be deleted (possibly
+        zero).
+        """
+        return self._failure_count
+
+    @property
+    def errors(self):
+        """A list of ``auth.ErrorInfo`` instances describing the errors that
```
RelaxedPear: Just checking, is this supposed to be in double back ticks?

firebase_admin/auth.py
```diff
@@ -446,6 +517,36 @@ def delete_user(uid, app=None):
     user_manager.delete_user(uid)
 
 
+def delete_users(uids, app=None):
+    """Deletes the users specified by the given identifiers.
+
+    Deleting a non-existing user won't generate an error. (i.e. this method is
+    idempotent.) Non-existing users are considered to be successfully
+    deleted, and will therefore be counted in the
+    `DeleteUserResult.success_count` value.
```
RelaxedPear: Suggest:

"Deleting a non-existing user does not generate an error (the method is idempotent). Non-existing users are considered to be successfully deleted and are therefore included in the `DeleteUserResult.success_count` value.

firebase_admin/_user_mgt.py
rsgowman: No idea! It seems both double and single backticks are used throughout. Single backticks are more common though, so I switched to that.

firebase_admin/auth.py
rsgowman: Done.

firebase_admin/_user_identifier.py
```diff
@@ -0,0 +1,80 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Classes to uniquely identify a user."""
+
+class UserIdentifier:
+    """Identifies a user to be looked up."""
+
```
hiranya911, line 19: Nit: Should we at least have a `pass` statement here just for completeness sake?

firebase_admin/_user_identifier.py
```diff
@@ -0,0 +1,80 @@
...
+    See ``auth.get_user()``.
+    """
+
+    def __init__(self, uid):
+        """Constructs a new `UidIdentifier` object.
+
+        Args:
+            uid: A user ID string.
+        """
+        self.uid = uid
```
hiranya911, line 33: Do we need to validate these arguments? (Not empty, not None, string type etc)

firebase_admin/_auth_client.py
```diff
@@ -182,6 +183,58 @@ def get_user_by_phone_number(self, phone_number):
...
+                return next((
+                    True
+                    for user_info in user_record.provider_data
+                    if identifier.provider_id == user_info.provider_id
+                    and identifier.provider_uid == user_info.uid
+                ), False)
+            raise TypeError("Unexpected type: {}".format(type(identifier)))
+
+        def _is_user_found(identifier, user_records):
+            return next(
```
hiranya911, line 228:
```
return any(_matches(identifier, user_record) for user_record in user_records)
```

This is cleaner and also cleverly stops calls to _matches as soon as the first match is found.

firebase_admin/_auth_client.py
```diff
@@ -182,6 +183,58 @@ def get_user_by_phone_number(self, phone_number):
...
+                (True for user_record in user_records if _matches(identifier, user_record)),
+                False)
+
+        users = [_user_mgt.UserRecord(user) for user in response]
```
hiranya911, line 232: Should we push this logic inside GetUsersResult?

firebase_admin/_user_mgt.py
```diff
@@ -492,6 +599,58 @@ def get_user(self, **kwargs):
...
+        """
+        if not identifiers:
+            return []
+        if len(identifiers) > 100:
+            raise ValueError('`identifiers` parameter must have <= 100 entries.')
+
+        payload = {}
+        for identifier in identifiers:
+            if isinstance(identifier, _user_identifier.UidIdentifier):
+                _auth_utils.validate_uid(identifier.uid, required=True)
```
hiranya911, line 627: What if we push this validation and the subsequent marshaling to the corresponding classes?

Each identifier class can expose something like:

```
def marshal(self, payload):
   payload['localId'] = payload.get('localId', []).append(self.uid)
```

firebase_admin/_user_mgt.py
```diff
@@ -492,6 +599,58 @@ def get_user(self, **kwargs):
...
+                payload['localId'] = payload.get('localId', []) + [identifier.uid]
```
hiranya911, line 628: .append()

firebase_admin/_user_identifier.py
rsgowman: It can't hurt, though implies we'll be doing some of the validation twice. (I don't think that's a problem.) Done.

Note that we don't validate arguments to (eg) `auth.get_user_by_email()`, instead, just deferring to the implementation function (user_mgt.get_user). This is a slightly different scenario, since for UserIdentifier, the user can create that ahead of time and possibly use it in some other context.

Another improvement (that I don't want to make in this PR) is to use type annotations. All of our supported python versions now support type annotations and they'd give you some of this for free (and at "compile" time too.)

firebase_admin/_auth_client.py
rsgowman: Done.

firebase_admin/_auth_client.py
rsgowman: I've been treating GUR as a "dumb" struct and therefore kept most logic out of it. (I like that it doesn't need to know/care what the response from the network looks like.) But I don't feel strongly. (Something has to adapt the network response.)

That said, all the other ports behave this way; if we change it here, we should probably change it everywhere else too. If you feel strongly, then let's do exactly that. Else, I'd prefer to just leave this be.

firebase_admin/_user_mgt.py
rsgowman: I did this (just now) but didn't really like the results. It makes this function shorter, but it alters the `UserIdentifier` types such that they need to know how the json payload is constructed. (For instance, should it be `ProviderIdentifier`s job to know that `provider_uid` maps to `rawId` on the wire?) It also spreads building the payload across 2 files and 5 classes, instead of having it in just one spot.

One possible improvement we could make here is to eliminate the validation (or at least, refactor it somewhat). As per one of your other comments, I've already moved some validation to the ctors, but that won't stop users from changing the values after construction, but before calling get_users(), so we should keep the validation here. But if we altered (eg) UidIdentifier.uid to be UidIdentifier._uid and added a getter (essentially making these objects immutable) I think that could allow us to skip validation here entirely, which would allow this block to focus solely on creating the payload. I've done that. PTAL.

firebase_admin/_user_mgt.py
rsgowman: append doesn't work (directly) here. The problem is that if localId doesn't exist in the dict, then it returns an empty list. This list would then be appended with the uid, but since list.append() doesn't return itself, we end up setting payload['localId'] to None.

OTOH, using a defaultdict would allow append to work nicely. Done.

firebase_admin/_user_identifier.py
```diff
@@ -30,7 +32,12 @@ def __init__(self, uid):
         Args:
             uid: A user ID string.
         """
-        self.uid = uid
+        _auth_utils.validate_uid(uid, required=True)
+        self._uid = uid
```
hiranya911, line 36: You can collapse these into a single line:

```
self._uid = _auth_utils.validate_uid(...)
```

Same below.

firebase_admin/_user_mgt.py
hiranya911: Thanks. I like where this ended up 👍

firebase_admin/_user_identifier.py
rsgowman: Done.

firebase_admin/_user_mgt.py
hiranya911: I like how validation turned out (and the immutability of the properties). I don't feel strongly about the marshalling part. When I first suggested it, I felt that having each class being in charge of marshalling was a good thing. If we add more implementations of the UserIdentifier interface in the future, that's when this will matter more. I'm ok with leaving this as is for now.

PR #403, opened 2020-02-03
fix(fcm): Convert event_time to UTC

lahirumaramba:
- Check whether the `datetime` object is naive or timezone aware
- If a naive `datetime` object is provided then set its timezone to the local timezone
- Convert the `event_time` to UTC Zulu format

RELEASE NOTE: `AndroidNotification` class now correctly formats the `event_time` field sent to the Cloud Messaging service.

### DIFF
--- a/firebase_admin/_messaging_encoder.py
+++ b/firebase_admin/_messaging_encoder.py
@@ -324,6 +324,8 @@ class MessageEncoder(json.JSONEncoder):
 
         event_time = result.get('event_time')
         if event_time:
+            # if the datetime instance is not naive (tzinfo is present), convert to UTC
+            # otherwise (tzinfo is None) assume the datetime instance is already in UTC
             if event_time.tzinfo is not None:
                 event_time = event_time.astimezone(datetime.timezone.utc)
             result['event_time'] = event_time.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
--- a/firebase_admin/_messaging_encoder.py
+++ b/firebase_admin/_messaging_encoder.py
@@ -324,8 +324,9 @@ class MessageEncoder(json.JSONEncoder):
 
         event_time = result.get('event_time')
         if event_time:
-            utc_event_time = event_time.astimezone(datetime.timezone.utc)
-            result['event_time'] = utc_event_time.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+            if event_time.tzinfo is not None:
+                event_time = event_time.astimezone(datetime.timezone.utc)
+            result['event_time'] = event_time.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
 
         priority = result.get('notification_priority')
         if priority:
--- a/firebase_admin/_messaging_encoder.py
+++ b/firebase_admin/_messaging_encoder.py
@@ -19,6 +19,7 @@ import json
 import math
 import numbers
 import re
+import pytz
 
 import firebase_admin._messaging_utils as _messaging_utils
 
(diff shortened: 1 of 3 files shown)


firebase_admin/_messaging_encoder.py
```diff
@@ -324,7 +325,12 @@ def encode_android_notification(cls, notification):
 
         event_time = result.get('event_time')
         if event_time:
-            result['event_time'] = str(event_time.isoformat()) + 'Z'
+            if event_time.tzinfo is None:
```
hiranya911: I don't think you have to do any of this. At least on my machine, following works as expected.

```
>>> import datetime
>>> d = datetime.datetime.now()
>>> d
datetime.datetime(2020, 2, 3, 13, 39, 32, 133863)
>>> d.astimezone(datetime.timezone.utc)
datetime.datetime(2020, 2, 3, 21, 39, 32, 133863, tzinfo=datetime.timezone.utc)
```

The first is in my local timezone (13:39 PST). The second is in UTC (21:39 UTC). Isn't that all we want?

tests/test_messaging.py
```diff
@@ -547,7 +547,10 @@ def test_android_notification(self):
                     click_action='ca', title_loc_key='tlk', body_loc_key='blk',
                     title_loc_args=['t1', 't2'], body_loc_args=['b1', 'b2'], channel_id='c',
                     ticker='ticker', sticky=True,
-                    event_timestamp=datetime.datetime(2019, 10, 20, 15, 12, 23, 123),
+                    event_timestamp=datetime.datetime(
```
hiranya911: Need a test case that triggers the no-timezone case as well.

firebase_admin/_messaging_encoder.py
lahirumaramba: Thanks! You're right. I missed that `astimezone` uses the local time zone if the provided datetime object is naive. Will update the code.

tests/test_messaging.py
lahirumaramba: With the new changes do we still need a test case for a no-timezone? No time-zone cases will depend on the local timezone of the test servers

hiranya911: I think one separate test case just focusing on `event_time` would be useful. Something like:

```
now = datetime.datetime.now()
expected = now.astimezone(utc).strftime(....)
# serialize test message with 'now' and compare against expected.
```

lahirumaramba:
It turns out that in Python 3.5 `astimezone()` cannot be used with naive `datetime`s. Therefore, we will assume that all naive `datetime`s are already to be in the UTC timezone. The API docs are updated to reflect this change.

@hiranya911 @egilmorez

firebase_admin/_messaging_utils.py
```diff
@@ -98,7 +98,8 @@ class AndroidNotification:
             the user clicks it (optional).
         event_timestamp: For notifications that inform users about events with an absolute time
             reference, sets the time that the event in the notification occurred as a
-            ``datetime.datetime`` instance. Notifications in the panel are sorted by this time
+            ``datetime.datetime`` instance. If the ``datetime.datetime`` instance is naive it will
+            be assumed to be in the UTC timezone. Notifications in the panel are sorted by this time
```
lahirumaramba: FYI @egilmorez this is the API docs change. Thanks!

firebase_admin/_messaging_encoder.py
```diff
@@ -324,7 +324,9 @@ def encode_android_notification(cls, notification):
 
         event_time = result.get('event_time')
         if event_time:
-            result['event_time'] = str(event_time.isoformat()) + 'Z'
+            if event_time.tzinfo is not None:
+                event_time = event_time.astimezone(datetime.timezone.utc)
+            result['event_time'] = event_time.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
```
hiranya911: Please add a comment here explaining the logic and the rationale here.

lahirumaramba: Updated! Thanks!

firebase_admin/_messaging_utils.py
egilmorez: Not critical, butI wonder if there is a more terse way to say "it will be assumed to be in the" . . . would "naive, it defaults to" be correct?  Maybe "naive, it is set to" ?

If those aren't accurate, this is fine as/is.

PR #410, opened 2020-02-11
chore: Implementing Pypi publish and Tweet steps

hiranya911:
Implementing the last 2 steps of the release workflow. Used pre-existing Actions for both. This precludes the need for having Python installed in the runtime.

PS: I haven't configured the necessary secrets in GitHub yet. Will do that later once we are ready to try this out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Release
+name: Release Candidate
 
 on:
   # Only run the workflow when a PR is updated or when a developer explicitly requests
@@ -49,7 +49,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install wheel
+        pip install setuptools wheel
 
     - name: Run unit tests
       run: pytest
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -131,11 +120,22 @@ jobs:
...
       if: success() &&
         contains(github.event.pull_request.labels.*.name, 'release:tweet')
-      run: echo Posting Tweet
+      uses: ethomson/send-tweet-action@v1
+      with:
+        status: >
+          ${{ steps.preflight.outputs.version }} of @Firebase Admin Python SDK is avaialble.
+          https://github.com/firebase/firebase-admin-python/releases/tag/${{ steps.preflight.outputs.version }}
+        consumer-key: ${{ secrets.FIREBASE_TWITTER_CONSUMER_KEY }}
+        consumer-secret: ${{ secrets.FIREBASE_TWITTER_CONSUMER_SECRET }}
```
lahirumaramba: Not critical, but I was wondering whether we should remove the `FIREBASE_` prefix from these secrets.
`TWITTER_CONSUMER_SECRET` instead `FIREBASE_TWITTER_CONSUMER_SECRET`

hiranya911: Done

PR #416, opened 2020-02-24
chore: Cleaning up scripts used in the old release process

hiranya911:
* Removed the bash scripts used in the old process.
* Setting CI flow to get triggered by `pull_request` events instead of `push`. This is required to ensure that we run CI on pull requests sent by external contributors. This does mean we no longer run CI on every push, which should be fine given that we review all changes through a PR.


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: Continuous Integration
 
-on: push
+on: pull_request
```
lahirumaramba: Do we need both `push` and `pull_request` events to trigger CI if we make changes to a PR based on reviews?

hiranya911: pull_request events are triggered on `open`, `synchronize` and `reopen` sub events. `synchronize` sub event covers the case we need.

PR #423, opened 2020-03-20
Modifying operation handling to support backend changes

ifielker:
Modifying operation handling to support backend changes

### DIFF
--- a/firebase_admin/ml.py
+++ b/firebase_admin/ml.py
@@ -841,7 +841,8 @@ class _MLService:
 
         op_name = _validate_operation_name(operation.get('name'))
         metadata = operation.get('metadata')
-        if metadata is None or '@type' not in metadata or 'ModelOperationMetadata' not in metadata.get('@type'):
+        if (metadata is None or '@type' not in metadata or
+                'ModelOperationMetadata' not in metadata.get('@type')):
             raise TypeError('Unknown type of operation metadata.')
         _, model_id = _validate_and_parse_name(metadata.get('name'))
         current_attempt = 0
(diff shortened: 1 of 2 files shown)


firebase_admin/ml.py
```diff
@@ -841,8 +839,12 @@ def handle_operation(self, operation, wait_for_operation=False, max_time_seconds
                 raise _utils.handle_operation_error(operation.get('error'))
             raise exceptions.UnknownError(message='Internal Error: Malformed Operation.')
 
-        op_name = operation.get('name')
-        _, model_id = _validate_and_parse_operation_name(op_name)
+        op_name = _validate_operation_name(operation.get('name'))
```
hiranya911: Nit: Push this further down, as it's not immediately used.

firebase_admin/ml.py
```diff
@@ -841,8 +839,12 @@ def handle_operation(self, operation, wait_for_operation=False, max_time_seconds
...
+        metadata = operation.get('metadata')
+        if (metadata is None or '@type' not in metadata or
+                'ModelOperationMetadata' not in metadata.get('@type')):
+            raise TypeError('Unknown type of operation metadata.')
```
hiranya911:
```suggestion
        metadata = operation.get('metadata', {})
        metadata_type = metadata.get('@type', '')
        if not metadata_type.endswith('.ModelOperationMetadata'):
            raise TypeError('Unknown type of operation metadata.')
```

firebase_admin/ml.py
ifielker: It's an early out. If the operation name is not valid, you won't be able to get_operation below so you might as well not bother with unpacking metadata or setting start and stop times etc. Also, pushing it down would make the testing a bit harder as we'd have to include valid metadata with the op_name validation tests.

firebase_admin/ml.py
ifielker: Nice! Done.

PR #424, opened 2020-03-20
Added create_tenant() and update_tenant() APIs

hiranya911:


firebase_admin/tenant_mgt.py
```diff
@@ -141,6 +194,56 @@ def get_tenant(self, tenant_id):
...
+        except requests.exceptions.RequestException as error:
+            raise _auth_utils.handle_auth_backend_error(error)
+        else:
+            return Tenant(body)
+
+    def update_tenant(
+            self, tenant_id, display_name=None, allow_password_sign_up=None,
+            enable_email_link_sign_in=None):
+        """Updates the specified tenant with the given parameters."""
+        if not isinstance(tenant_id, str) or not tenant_id:
```
lahirumaramba: I think `isinstance(tenant_id, str)` covers `not tenant_id` cases, right?
Also, did we choose not to use `_auth_utils.validate_string` here to provide a more specific error message?

lahirumaramba:
@hiranya911 should we get the docs reviewed by someone or would you rather do that in the final merge PR?

firebase_admin/tenant_mgt.py
hiranya911: That doesn't cover the case where `tenant_id == ''`.

It's the same reason why we can't use `validate_string()` here.

hiranya911:
I'll get the docs reviewed towards the end, when everything's in place.

PR #442, opened 2020-04-06
feat(auth): Added OIDCProviderConfig type and get/delete APIs

hiranya911:
* Added `OIDCProviderConfig` class.
* Added get/delete functions.


firebase_admin/_auth_providers.py
```diff
@@ -149,6 +165,15 @@ def __init__(self, http_client, project_id, tenant_id=None):
...
             self.base_url += '/tenants/{0}'.format(tenant_id)
 
+    def get_oidc_provider_config(self, provider_id):
+        _validate_oidc_provider_id(provider_id)
+        body = self._make_request('get', '/oauthIdpConfigs/{0}'.format(provider_id))
+        return OIDCProviderConfig(body)
+
+    def delete_oidc_provider_config(self, provider_id):
+        _validate_oidc_provider_id(provider_id)
+        self._make_request('delete', '/oauthIdpConfigs/{0}'.format(provider_id))
```
lahirumaramba, line 175: Should we return the status 'success/failed' here?

firebase_admin/_auth_providers.py
```diff
@@ -253,6 +278,16 @@ def _make_request(self, method, path, **kwargs):
             raise _auth_utils.handle_auth_backend_error(error)
 
 
+def _validate_oidc_provider_id(provider_id):
+    if not isinstance(provider_id, str):
+        raise ValueError(
+            'Invalid OIDC provider ID: {0}. Provider ID must be a non-empty string.'.format(
```
lahirumaramba, line 284: This will not check for empty strings though, will it?

firebase_admin/_auth_providers.py
hiranya911: Return of these delete methods are `void`.

firebase_admin/_auth_providers.py
hiranya911: The immediate next check (`if not provider_id.startswith('oidc.'):`) will cover empty strings.

PR #445, opened 2020-04-14
Mlkit add headers

ifielker:
Add SDK version headers to the requests

### DIFF
--- a/firebase_admin/ml.py
+++ b/firebase_admin/ml.py
@@ -27,6 +27,7 @@ from urllib import parse
 
 import requests
 
+import firebase_admin
 from firebase_admin import _http_client
 from firebase_admin import _utils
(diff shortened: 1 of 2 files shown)


firebase_admin/ml.py
```diff
@@ -785,9 +786,15 @@ def __init__(self, app):
         self._project_url = _MLService.PROJECT_URL.format(self._project_id)
         self._client = _http_client.JsonHttpClient(
             credential=app.credential.get_credential(),
+            headers={
```
hiranya911, line 789: Nit: Avoid duplication. Initialize the headers into a variable, and pass it to the constructors.

ifielker: done.

PR #449, opened 2020-04-16
Adding tensorflow and keras so we don't skip tests

ifielker:
Adding tensorflow and keras so we don't skip tests


.github/workflows/release.yml
```diff
@@ -54,6 +54,8 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         pip install setuptools wheel
+        pip install tensorflow
```
hiranya911, line 57: Nit: May be combine into one line? Your call.

ifielker: I think it's fine

PR #455, opened 2020-04-23
fix(auth): Snippets for multi-tenancy and IdP management APIs

hiranya911:
Snippets to appear in:
* https://cloud.google.com/identity-platform/docs/managing-providers-programmatically
* https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants

### DIFF
--- a/snippets/auth/index.py
+++ b/snippets/auth/index.py
@@ -25,6 +25,7 @@ import firebase_admin
 from firebase_admin import credentials
 from firebase_admin import auth
 from firebase_admin import exceptions
+from firebase_admin import tenant_mgt
 
 sys.path.append("lib")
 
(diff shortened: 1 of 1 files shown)


snippets/auth/index.py
```diff
@@ -634,6 +635,418 @@ def send_custom_email(email, link):
...
+            '-----BEGIN CERTIFICATE-----\nCERT2...\n-----END CERTIFICATE-----',
+        ],
+        rp_entity_id='P_ENTITY_ID',
+        callback_url='https://project-id.firebaseapp.com/__/auth/handler')
+
+    print('Created new SAML provider:', saml.provider_id)
+	# [END create_saml_provider]
+
+def update_saml_provider_config():
+	# [START update_saml_provider]
```
lahirumaramba, line 657: Do we put the `[START]` `[END]` tags inside the function or outside?
I have seen them outside in some code samples

hiranya911: We normally put them inside, since the custom function signatures do not add any useful information to the snippet.

PR #456, opened 2020-04-28
fix(auth): Fixed some API reference mistakes

hiranya911:
* Adding some missing signatures to the `__all__` list.
* Correcting an incorrect alias.

There's a [lint error](https://github.com/firebase/firebase-admin-python/actions/runs/89621645) that is caused by a regression in the latest version of Astroid (see https://github.com/PyCQA/astroid/issues/775). I'm specifying an old version of Astroid until that gets resolved.

### DIFF
--- a/firebase_admin/ml.py
+++ b/firebase_admin/ml.py
@@ -148,7 +148,7 @@ def list_models(list_filter=None, page_size=None, page_token=None, app=None):
     """Lists models from Firebase ML.
 
     Args:
-        list_filter: a list filter string such as "tags:'tag_1'". None will return all models.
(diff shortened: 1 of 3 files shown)


firebase_admin/ml.py
```diff
@@ -148,7 +148,7 @@ def list_models(list_filter=None, page_size=None, page_token=None, app=None):
     """Lists models from Firebase ML.
 
     Args:
-        list_filter: a list filter string such as "tags:'tag_1'". None will return all models.
+        list_filter: a list filter string such as ``tags:'tag_1'``. None will return all models.
```
ifielker, line 151: Thanks!

PR #461, opened 2020-05-12
Followup to #400 to ensure all new types are exported

rsgowman:


firebase_admin/auth.py
```diff
@@ -363,7 +366,7 @@ def get_users(identifiers, app=None):
 
     Returns:
         GetUsersResult: A ``GetUsersResult`` instance corresponding to the
-            specified identifiers.
+        specified identifiers.
```
hiranya911, line 368: Thanks. I think line 362 needs to be "A list of ``UserIdentifier`` instances...."

rsgowman: Done. (It *used* to be Identifier, but then we changed it. I must have missed updating it here.)

PR #464, opened 2020-05-27
Add auth bulk get/delete snippets

samtstern:
Based on:
https://github.com/firebase/snippets-node/pull/86

### DIFF
--- a/snippets/auth/index.py
+++ b/snippets/auth/index.py
@@ -207,7 +207,7 @@ def bulk_get_users():
     for user in result.users:
         print(user)
 
-    print ('Unable to find users corresponding to these identifiers:')
+    print('Unable to find users corresponding to these identifiers:')
     for uid in result.notfound:
         print(uid)
     # [END bulk_get_users]
--- a/snippets/auth/index.py
+++ b/snippets/auth/index.py
@@ -205,10 +205,10 @@ def bulk_get_users():
 
     print('Successfully fetched user data:')
     for user in result.users:
-        print(user)
+        print(user.uid)
 
     print('Unable to find users corresponding to these identifiers:')
-    for uid in result.notfound:
+    for uid in result.not_found:
         print(uid)
     # [END bulk_get_users]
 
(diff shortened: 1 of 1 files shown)


snippets/auth/index.py
```diff
@@ -192,6 +192,26 @@ def get_user_by_email():
...
+    result = auth.get_users([
+        auth.UidIdentifier('uid1'),
+        auth.EmailIdentifier('user2@example.com'),
+        auth.PhoneIdentifier(+15555550003),
+        auth.ProviderIdentifier('google.com', 'google_uid4')
+    ])
+
+    print('Successfully fetched user data:')
+    for user in result.users:
+        print(user)
```
rsgowman, line 208: Same problem as in java... we haven't defined `__str__()` on our objects, so the results of these print statements aren't great. You could use the same solution, i.e.: `print(user.uid)`

snippets/auth/index.py
```diff
@@ -192,6 +192,26 @@ def get_user_by_email():
...
+
+    print('Unable to find users corresponding to these identifiers:')
+    for uid in result.notfound:
```
rsgowman, line 211: s/notfound/not_found/

snippets/auth/index.py
```diff
@@ -192,6 +192,26 @@ def get_user_by_email():
...
+        print(uid)
```
rsgowman, line 212: This one's unfortunate. There's no `__str__()`, or anything else that can be printed. You could compare to the input identifiers, but that's too wordy for a snippet. This might be the best we can do (and will automatically become better if we add `__str__()` in the future). No action required.

snippets/auth/index.py
```diff
@@ -242,6 +262,18 @@ def delete_user(uid):
...
+def bulk_delete_users():
+    # [START bulk_delete_users]
+    from firebase_admin import auth
+
+    result = auth.delete_users(["uid1", "uid2", "uid3"])
+
+    print('Successfully deleted {0} users'.format(result.success_count))
+    print('Failed to delete {0} users'.format(result.failure_count))
+    for err in result.errors:
+        print(err)
```
rsgowman, line 274: How about: `print('Index: {0}; Reason: {1}'.format(result.index, result.reason))`

Issue #472, opened 2020-06-26
FR: Manage projects and security rules


radyz:
I recently came across a project where needed it the following functionality and had to implement it via the rest api:

- Manage projects: https://firebase.google.com/docs/projects/api/reference/rest/v1beta1/projects
- Manage operations: https://firebase.google.com/docs/projects/api/reference/rest/v1beta1/operations/get
- Manage security rules: https://firebase.google.com/docs/reference/admin/node/admin.securityRules

I'll be glad to integrate those into the sdk

hiranya911:
Managing projects is already supported: https://firebase.google.com/docs/reference/admin/python/firebase_admin.project_management

We do not wish to expose operations management directly. But we already use it internally for certain use cases.

Security rules management is currently available in our Node.js SDK. If you or somebody wishes to port that to Python that would be great :)

PR #474, opened 2020-06-27
Add managements methods for web apps

radyz:
Related to https://github.com/firebase/firebase-admin-python/issues/472

- [x] Create web app
- [x] List web apps
- [x] Get web app


hiranya911:
Thanks @radyz for the PR. This will probably require an internal API review as it adds to the existing API surface. Please sit tight.

PR #477, opened 2020-07-13
Added integration test for AutoML

ifielker:
Added conditional integration test for AutoML.


integration/test_ml.py
```diff
@@ -133,17 +139,20 @@ def check_model(model, args):
     assert model.etag is not None
 
 
-def check_model_format(model, has_model_format=False, validation_error=None):
+def check_model_format(model, has_model_format=False, validation_error=None, is_automl=False):
```
hiranya911, line 142: I'd advise having 2 methods:

```
check_tflite_model(model, has_format=False, validation_error=None)
check_automl_model(model, has_format=False)
```

integration/test_ml.py
```diff
@@ -371,3 +381,50 @@ def test_from_saved_model(saved_model_dir):
...
+def automl_model():
+    assert _AUTOML_ENABLED
+
+    # It takes > 20 minutes to train a model, so we expect a predefined AutoMl
+    # model named 'py_sdk_integ_test1' to exist in the project, or we skip
+    # the test.
+    automl_client = automl_v1.AutoMlClient()
+    project_id = firebase_admin.get_app().project_id
+    parent = automl_client.location_path(project_id, 'us-central1')
+    models = automl_client.list_models(parent, filter_="display_name=py_sdk_integ_test1")
```
hiranya911, line 401: May be call it `admin_sdk_integ_test1`. I suppose you will use this in Node.js too.

integration/test_ml.py
```diff
@@ -34,6 +35,11 @@
 except ImportError:
     _TF_ENABLED = False
 
+try:
+    from google.cloud import automl_v1
```
hiranya911, line 39: Once this is merged to master, let's also update the `.github/workflows/release.yml` to install this package, and add the Auto ML resource to the project Kobayashi Maru.

integration/test_ml.py
ifielker: done

integration/test_ml.py
ifielker: I split it up, but a bit differently - automl is also TFLite format just with a different source, so that might be misleading.
Also, tflite would have contained both models with a tflite_gcs format and models with no format.
Since we will likely implement more than TFLite models, I split it into 3: models with no format, models with tflite format from gcs source, and models with tflite format from automl source. We can then add additional format/source combinations easily in future.

integration/test_ml.py
ifielker: sounds good

PR #486, opened 2020-08-27
Fix doc spelling

therefromhere:


therefromhere:
Note - lint fail should be resolved by #485

PR #510, opened 2020-12-03
fix: Accept Path-like objects in credential factory functions

### LINKED ISSUE #501: pathlib and Django v3.1 Compatibility
* Operating System version: Windows 10 home single language
  * Firebase SDK version: _____
  * Library version: 4.4.0
  * Firebase Product: auth

Recently with Django 3.1 release, the previous os.path library was removed and pathlib is introduced for maintaining the file paths. The class type returned by os.path is str while pathlib returns its own pathlib.Path class type.

This raises ValueError even if the path is correct using pathlib and if we do it using the os.path library it works fine.

ValueError: Invalid certificate argument. Certificate argument must be a file path, or a dict containing th
e parsed file contents.

So, to reproduce this I figured out two ways:

1. Explicit typecasting of the path argument.
```
credentials.Certificate(str(path_to_json))
```

2. Some tweaks into the credentials.py file to accept pathlib object as well.

```
if isinstance(cert, str) or isinstance(cert, pathlib.Path):
            with open(cert) as json_file:
                json_data = json.load(json_file)
        elif isinstance(cert, dict):
            json_data = cert
        else:
            raise ValueError(
                'Invalid certificate argument: "{0}". Certificate argument must be a file path, '
                'or a dict containing the parsed file contents.'.format(cert))
```

hiranya911:
Credential factory functions like `credentials.cert()` currently only accept `str` typed file paths. But Python 3.4 introduced the `pathlib` module and with that a wide range of path-like types. This PR makes it possible to pass these path-like objects directly to our credential factory functions.

```py
import pathlib

path = pathlib.Path('home', 'user', 'firebase.json')
cred = credentials.Cert(path)
```

Basically we are going to treat anything that is accepted by `pathlib.Path` as a valid file path. This constructor has good built-in argument validation, and throws `TypeError` for things that not path-like.

Resolves #501

### DIFF
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -53,6 +53,8 @@ RESERVED_CLAIMS = set([
 ])
 METADATA_SERVICE_URL = ('http://metadata.google.internal/computeMetadata/v1/instance/'
                         'service-accounts/default/email')
+ALGORITHM_RS256 = 'RS256'
+ALGORITHM_NONE = 'none'
 
 # Emulator fake account
 AUTH_EMULATOR_EMAIL = 'firebase-auth-emulator@example.com'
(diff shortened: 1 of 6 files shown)


hiranya911:
As suspected encountering issues with Python 3.5. Instead of implementing an ugly hack, I think a better course of action is to drop Python 3.5 support (which EoL'ed a couple of months ago), and delay this fix until then.

lahirumaramba (approved):
Great! Thanks, Hiranya!

PR #526, opened 2021-01-05
Remove delayed response message for holidays

lahirumaramba:
- Remove delayed response message for holidays


hiranya911 (approved):
Thanks. LGTM!

PR #531, opened 2021-02-05
feat(auth): Add auth emulator support via the FIREBASE_AUTH_EMULATOR_HOST environment variable.

### LINKED ISSUE #503: FR: Support for authentication emulator
* Operating System version: Ubuntu 20.04
  * Firebase SDK version: 8.1.1
  * Library version: 4.4.0
  * Firebase Product: auth

It would be excellent if the Firebase Admin Python SDK could connect to the local authentication emulator. Similar issues exists for [the Go version](https://github.com/firebase/firebase-admin-go/issues/409) and for the [Java version](https://github.com/firebase/firebase-admin-java/issues/493).

muru:
### Discussion

Closes #503, by adding support for the authentication emulator via the `FIREBASE_AUTH_EMULATOR_HOST` environment variable.

Additionally, I added a change that *shouldn't* break anything, but does make the SDK work better with the emulator,  fixing the equivalent of https://github.com/firebase/firebase-admin-node/issues/1084. I ran into this while testing these changes in a project, and I feel it should be part of "supporting" the emulator, but I am open to putting it in a separate PR if it doesn't belong.

### Testing

- [x] Existing tests pass
- [x] Modified the existing tests for user management, token generation and providers to run for both emulated and non-emulated cases.

### API Changes

None, except insofar that the `FIREBASE_AUTH_EMULATOR_HOST` environment variable becomes part of the API.

### DIFF
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -128,7 +128,7 @@ class Client:
                 raise _auth_utils.TenantIdMismatchError(
                     'Invalid tenant ID: {0}'.format(token_tenant_id))
 
-        if not self.emulated and check_revoked:
+        if check_revoked:
             self._check_jwt_revoked(verified_claims, _token_gen.RevokedIdTokenError, 'ID token')
         return verified_claims
 
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -24,6 +24,7 @@ from firebase_admin import _token_gen
 from firebase_admin import _user_identifier
 from firebase_admin import _user_import
 from firebase_admin import _user_mgt
+from firebase_admin import _utils
 
 
 class Client:
@@ -36,18 +37,37 @@ class Client:
             2. set the project ID explicitly via Firebase App options, or
             3. set the project ID via the GOOGLE_CLOUD_PROJECT environment variable.""")
 
-        credential = app.credential.get_credential()
+        credential = None
         version_header = 'Python/Admin/{0}'.format(firebase_admin.__version__)
         timeout = app.options.get('httpTimeout', _http_client.DEFAULT_TIMEOUT_SECONDS)
+        # Non-default endpoint URLs for emulator support are set in this dict later.
+        endpoint_urls = {}
+        self.emulated = False
+
+        # If an emulator is present, check that the given value matches the expected format and set
+        # endpoint URLs to use the emulator. Additionally, use a fake credential.
+        emulator_host = _auth_utils.get_emulator_host()
+        if emulator_host:
+            base_url = 'http://{0}/identitytoolkit.googleapis.com'.format(emulator_host)
+            endpoint_urls['v1'] = base_url + '/v1'
+            endpoint_urls['v2beta1'] = base_url + '/v2beta1'
+            credential = _utils.EmulatorAdminCredentials()
+            self.emulated = True
+        else:
+            # Use credentials if provided
+            credential = app.credential.get_credential()
+
         http_client = _http_client.JsonHttpClient(
             credential=credential, headers={'X-Client-Version': version_header}, timeout=timeout)
 
         self._tenant_id = tenant_id
-        self._token_generator = _token_gen.TokenGenerator(app, http_client)
+        self._token_generator = _token_gen.TokenGenerator(
+            app, http_client, url_override=endpoint_urls.get('v1'))
         self._token_verifier = _token_gen.TokenVerifier(app)
-        self._user_manager = _user_mgt.UserManager(http_client, app.project_id, tenant_id)
+        self._user_manager = _user_mgt.UserManager(
+            http_client, app.project_id, tenant_id, url_override=endpoint_urls.get('v1'))
         self._provider_manager = _auth_providers.ProviderConfigClient(
-            http_client, app.project_id, tenant_id)
+            http_client, app.project_id, tenant_id, url_override=endpoint_urls.get('v2beta1'))
 
     @property
     def tenant_id(self):
@@ -108,7 +128,7 @@ class Client:
                 raise _auth_utils.TenantIdMismatchError(
                     'Invalid tenant ID: {0}'.format(token_tenant_id))
 
-        if check_revoked:
+        if not self.emulated and check_revoked:
             self._check_jwt_revoked(verified_claims, _token_gen.RevokedIdTokenError, 'ID token')
         return verified_claims
 
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -84,11 +84,12 @@ class TokenGenerator:
 
     ID_TOOLKIT_URL = 'https://identitytoolkit.googleapis.com/v1'
 
-    def __init__(self, app, http_client):
+    def __init__(self, app, http_client, url_override=None):
         self.app = app
         self.http_client = http_client
         self.request = transport.requests.Request()
-        self.base_url = '{0}/projects/{1}'.format(self.ID_TOOLKIT_URL, app.project_id)
+        url_prefix = url_override or self.ID_TOOLKIT_URL
+        self.base_url = '{0}/projects/{1}'.format(url_prefix, app.project_id)
         self._signing_provider = None
 
     def _init_signing_provider(self):
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -53,6 +53,19 @@ RESERVED_CLAIMS = set([
 METADATA_SERVICE_URL = ('http://metadata.google.internal/computeMetadata/v1/instance/'
                         'service-accounts/default/email')
 
+# Emulator fake account
+AUTH_EMULATOR_EMAIL = 'firebase-auth-emulator@example.com'
+
+
+class _EmulatedSigner(google.auth.crypt.Signer):
+    key_id = None
+
+    def __init__(self):
+        pass
+
+    def sign(self, message):
+        return b''
+
 
 class _SigningProvider:
     """Stores a reference to a google.auth.crypto.Signer."""
@@ -78,6 +91,10 @@ class _SigningProvider:
         signer = iam.Signer(request, google_cred, service_account)
         return _SigningProvider(signer, service_account)
 
+    @classmethod
+    def for_emulator(cls):
+        return _SigningProvider(_EmulatedSigner(), AUTH_EMULATOR_EMAIL)
+
 
 class TokenGenerator:
     """Generates custom tokens and session cookies."""
@@ -94,6 +111,8 @@ class TokenGenerator:
 
     def _init_signing_provider(self):
         """Initializes a signing provider by following the go/firebase-admin-sign protocol."""
+        if _auth_utils.is_emulated():
+            return _SigningProvider.for_emulator()
         # If the SDK was initialized with a service account, use it to sign bytes.
         google_cred = self.app.credential.get_credential()
         if isinstance(google_cred, google.oauth2.service_account.Credentials):
@@ -286,12 +305,14 @@ class _JWTVerifier:
         verify_id_token_msg = (
             'See {0} for details on how to retrieve {1}.'.format(self.url, self.short_name))
 
+        emulated = _auth_utils.is_emulated()
+
         error_message = None
         if audience == FIREBASE_AUDIENCE:
             error_message = (
                 '{0} expects {1}, but was given a custom '
                 'token.'.format(self.operation, self.articled_short_name))
-        elif not header.get('kid'):
+        elif not emulated and not header.get('kid'):
             if header.get('alg') == 'HS256' and payload.get(
                     'v') == 0 and 'uid' in payload.get('d', {}):
                 error_message = (
@@ -299,7 +320,7 @@ class _JWTVerifier:
                     'token.'.format(self.operation, self.articled_short_name))
             else:
                 error_message = 'Firebase {0} has no "kid" claim.'.format(self.short_name)
-        elif header.get('alg') != 'RS256':
+        elif not emulated and header.get('alg') != 'RS256':
             error_message = (
                 'Firebase {0} has incorrect algorithm. Expected "RS256" but got '
                 '"{1}". {2}'.format(self.short_name, header.get('alg'), verify_id_token_msg))
@@ -330,11 +351,14 @@ class _JWTVerifier:
             raise self._invalid_token_error(error_message)
 
         try:
-            verified_claims = google.oauth2.id_token.verify_token(
-                token,
-                request=request,
-                audience=self.project_id,
-                certs_url=self.cert_url)
+            if emulated:
+                verified_claims = payload
+            else:
+                verified_claims = google.oauth2.id_token.verify_token(
+                    token,
+                    request=request,
+                    audience=self.project_id,
+                    certs_url=self.cert_url)
             verified_claims['uid'] = verified_claims['sub']
             return verified_claims
         except google.auth.exceptions.TransportError as error:
--- a/tests/test_token_gen.py
+++ b/tests/test_token_gen.py
@@ -148,7 +148,7 @@ def auth_app(request):
     This can be used in any scenario where the private key is required. Use user_mgt_app
     for everything else.
     """
-    monkeypatch = pytest.MonkeyPatch()
+    monkeypatch = testutils.new_monkeypatch()
     if request.param['emulated']:
         monkeypatch.setenv(EMULATOR_HOST_ENV_VAR, AUTH_EMULATOR_HOST)
         monkeypatch.setitem(TOKEN_MGT_URLS, 'ID_TOOLKIT', EMULATED_ID_TOOLKIT_URL)
@@ -159,7 +159,7 @@ def auth_app(request):
 
 @pytest.fixture(scope='module', params=[{'emulated': False}, {'emulated': True}])
 def user_mgt_app(request):
-    monkeypatch = pytest.MonkeyPatch()
+    monkeypatch = testutils.new_monkeypatch()
     if request.param['emulated']:
         monkeypatch.setenv(EMULATOR_HOST_ENV_VAR, AUTH_EMULATOR_HOST)
         monkeypatch.setitem(TOKEN_MGT_URLS, 'ID_TOOLKIT', EMULATED_ID_TOOLKIT_URL)
--- a/tests/test_token_gen.py
+++ b/tests/test_token_gen.py
@@ -55,6 +55,14 @@ INVALID_JWT_ARGS = {
     'NonEmptyDictToken': {'a': 1},
 }
 
+ID_TOOLKIT_URL = 'https://identitytoolkit.googleapis.com/v1'
+EMULATOR_HOST_ENV_VAR = 'FIREBASE_AUTH_EMULATOR_HOST'
+AUTH_EMULATOR_HOST = 'localhost:9099'
+EMULATED_ID_TOOLKIT_URL = 'http://{}/identitytoolkit.googleapis.com/v1'.format(AUTH_EMULATOR_HOST)
+TOKEN_MGT_URLS = {
+    'ID_TOOLKIT': ID_TOOLKIT_URL,
+}
+
 # Fixture for mocking a HTTP server
 httpserver = plugin.httpserver
 
@@ -121,7 +129,7 @@ def _instrument_user_manager(app, status, payload):
     user_manager = client._user_manager
     recorder = []
     user_manager.http_client.session.mount(
-        _token_gen.TokenGenerator.ID_TOOLKIT_URL,
+        TOKEN_MGT_URLS['ID_TOOLKIT'],
         testutils.MockAdapter(payload, status, recorder))
     return user_manager, recorder
 
@@ -133,23 +141,33 @@ def _overwrite_iam_request(app, request):
     client = auth._get_client(app)
     client._token_generator.request = request
 
-@pytest.fixture(scope='module')
-def auth_app():
+@pytest.fixture(scope='module', params=[{'emulated': False}, {'emulated': True}])
+def auth_app(request):
     """Returns an App initialized with a mock service account credential.
 
     This can be used in any scenario where the private key is required. Use user_mgt_app
     for everything else.
     """
+    monkeypatch = pytest.MonkeyPatch()
+    if request.param['emulated']:
+        monkeypatch.setenv(EMULATOR_HOST_ENV_VAR, AUTH_EMULATOR_HOST)
+        monkeypatch.setitem(TOKEN_MGT_URLS, 'ID_TOOLKIT', EMULATED_ID_TOOLKIT_URL)
     app = firebase_admin.initialize_app(MOCK_CREDENTIAL, name='tokenGen')
     yield app
     firebase_admin.delete_app(app)
+    monkeypatch.undo()
 
-@pytest.fixture(scope='module')
-def user_mgt_app():
+@pytest.fixture(scope='module', params=[{'emulated': False}, {'emulated': True}])
+def user_mgt_app(request):
+    monkeypatch = pytest.MonkeyPatch()
+    if request.param['emulated']:
+        monkeypatch.setenv(EMULATOR_HOST_ENV_VAR, AUTH_EMULATOR_HOST)
+        monkeypatch.setitem(TOKEN_MGT_URLS, 'ID_TOOLKIT', EMULATED_ID_TOOLKIT_URL)
     app = firebase_admin.initialize_app(testutils.MockCredential(), name='userMgt',
                                         options={'projectId': 'mock-project-id'})
     yield app
     firebase_admin.delete_app(app)
+    monkeypatch.undo()
 
 @pytest.fixture
 def env_var_app(request):
--- a/tests/test_token_gen.py
+++ b/tests/test_token_gen.py
@@ -76,13 +76,18 @@ def _merge_jwt_claims(defaults, overrides):
 
 def verify_custom_token(custom_token, expected_claims, tenant_id=None):
     assert isinstance(custom_token, bytes)
-    token = google.oauth2.id_token.verify_token(
-        custom_token,
-        testutils.MockRequest(200, MOCK_PUBLIC_CERTS),
-        _token_gen.FIREBASE_AUDIENCE)
+    expected_email = MOCK_SERVICE_ACCOUNT_EMAIL
+    if _is_emulated():
+        expected_email = _token_gen.AUTH_EMULATOR_EMAIL
+        token = jwt.decode(custom_token, verify=False)
+    else:
+        token = google.oauth2.id_token.verify_token(
+            custom_token,
+            testutils.MockRequest(200, MOCK_PUBLIC_CERTS),
+            _token_gen.FIREBASE_AUDIENCE)
     assert token['uid'] == MOCK_UID
-    assert token['iss'] == MOCK_SERVICE_ACCOUNT_EMAIL
-    assert token['sub'] == MOCK_SERVICE_ACCOUNT_EMAIL
+    assert token['iss'] == expected_email
+    assert token['sub'] == expected_email
     if tenant_id is None:
         assert 'tenant_id' not in token
     else:
@@ -141,7 +146,15 @@ def _overwrite_iam_request(app, request):
     client = auth._get_client(app)
     client._token_generator.request = request
 
-@pytest.fixture(scope='module', params=[{'emulated': False}, {'emulated': True}])
+
+def _is_emulated():
+    emulator_host = os.getenv(EMULATOR_HOST_ENV_VAR, '')
+    return emulator_host and '//' not in emulator_host
+
+
+# These fixtures are set to the default function scope as the emulator environment variable bleeds
+# over when in module scope.
+@pytest.fixture(params=[{'emulated': False}, {'emulated': True}])
 def auth_app(request):
     """Returns an App initialized with a mock service account credential.
 
@@ -157,7 +170,7 @@ def auth_app(request):
     firebase_admin.delete_app(app)
     monkeypatch.undo()
 
-@pytest.fixture(scope='module', params=[{'emulated': False}, {'emulated': True}])
+@pytest.fixture(params=[{'emulated': False}, {'emulated': True}])
 def user_mgt_app(request):
     monkeypatch = testutils.new_monkeypatch()
     if request.param['emulated']:
@@ -230,6 +243,12 @@ class TestCreateCustomToken:
             auth.create_custom_token(user, claims, app=auth_app)
 
     def test_noncert_credential(self, user_mgt_app):
+        if _is_emulated():
+            # Should work fine with the emulator, so do a condensed version of
+            # test_sign_with_iam below.
+            custom_token = auth.create_custom_token(MOCK_UID, app=user_mgt_app).decode()
+            self._verify_signer(custom_token, _token_gen.AUTH_EMULATOR_EMAIL)
+            return
         with pytest.raises(ValueError):
             auth.create_custom_token(MOCK_UID, app=user_mgt_app)
 
@@ -304,7 +323,7 @@ class TestCreateCustomToken:
     def _verify_signer(self, token, signer):
         segments = token.split('.')
         assert len(segments) == 3
-        body = json.loads(base64.b64decode(segments[1]).decode())
+        body = jwt.decode(token, verify=False)
         assert body['iss'] == signer
         assert body['sub'] == signer
 
@@ -406,13 +425,23 @@ class TestVerifyIdToken:
         'BadFormatToken': 'foobar'
     }
 
-    @pytest.mark.parametrize('id_token', valid_tokens.values(), ids=list(valid_tokens))
-    def test_valid_token(self, user_mgt_app, id_token):
-        _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
-        claims = auth.verify_id_token(id_token, app=user_mgt_app)
+    tokens_accepted_in_emulator = [
+        'NoKid',
+        'WrongKid',
+        'FutureToken',
+        'ExpiredToken'
+    ]
+
+    def _assert_valid_token(self, id_token, app):
+        claims = auth.verify_id_token(id_token, app=app)
         assert claims['admin'] is True
         assert claims['uid'] == claims['sub']
         assert claims['firebase']['sign_in_provider'] == 'provider'
+
+    @pytest.mark.parametrize('id_token', valid_tokens.values(), ids=list(valid_tokens))
+    def test_valid_token(self, user_mgt_app, id_token):
+        _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
+        self._assert_valid_token(id_token, app=user_mgt_app)
 
     def test_valid_token_with_tenant(self, user_mgt_app):
         _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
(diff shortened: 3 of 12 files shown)


yuchenshi (commented):
The approach looks right to me, although we should also handle *verifying* ID tokens and session cookies.

yuchenshi:
cc @brianrodri who may be interested in this.

muru:
@yuchenshi  At the time I started work on this I thought the Google people were going to rework how those are handled. Now that the firebase CLI is supposed to work OK with session cookies, I'll try to include it in integration tests (which I'd held off on) and see what else needs to be fixed here.

yuchenshi:
> @yuchenshi At the time I started work on this I thought the Google people were going to rework how those are handled. Now that the firebase CLI is supposed to work OK with session cookies, I'll try to include it in integration tests (which I'd held off on) and see what else needs to be fixed here.

I really appreciate it. https://github.com/firebase/firebase-admin-node/pull/1148 has some integration tests that you can draw inspiration from and it shows the desired behavior for verifying ID tokens.

firebase_admin/_token_gen.py
```diff
@@ -198,7 +200,7 @@ def create_session_cookie(self, id_token, expires_in):
         url = '{0}:createSessionCookie'.format(self.base_url)
         payload = {
             'idToken': id_token,
-            'validDuration': expires_in,
+            'validDuration': str(expires_in),
```
hiranya911, line 203: I think this is no longer necessary? @yuchenshi should be able to confirm.

tests/testutils.py
```diff
@@ -58,6 +60,15 @@ def run_without_project_id(func):
                 os.environ[env_var] = gcloud_project
 
 
+def new_monkeypatch():
+    try:
+        monkeypatch = pytest.MonkeyPatch()
```
hiranya911, line 65: Local variable not needed. Can just return here and below.

hiranya911 (changes requested):
Thanks for the PR @muru. Looks mostly good. I've pointed out a few things that can be improved.

firebase_admin/_token_gen.py
yuchenshi: This should be no longer necessary with the latest Firebase CLI. Please let us know if otherwise. Integration tests can help here.

Keramblock:
@muru sorry for disturbance, but could you please take a look? I really wait for that MR  =)

firebase_admin/_auth_client.py
```diff
@@ -36,17 +39,42 @@ def __init__(self, app, tenant_id=None):
...
+        # If an emulator is present, check that the given value matches the expected format and set
+        # endpoint URLs to use the emulator. Also set a fake authorization token.
+        emulator_host = os.environ.get(_EMULATOR_HOST_ENV_VAR)
+        if emulator_host:
+            if '//' in emulator_host:
+                raise ValueError(
+                    'Invalid {0}: "{1}". It must follow format "host:port".'.format(
+                        _EMULATOR_HOST_ENV_VAR, emulator_host))
+            # Used instead of credential
+            http_headers["Authorization"] = "Bearer owner"
```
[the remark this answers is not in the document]
muru: This turned out to be a bit tricky - one of tests (`tests/test_user_mgt.py::TestImportUsers::test_import_users_http_error`) would fail trying to refresh the token. So I have settled in favour of using credentials if provided, or falling back to setting the token if not.

firebase_admin/_token_gen.py
muru: I'll dropped this commit for now, and add it back if the integration tests have a problem (which I don't think will happen).

muru:
@Keramblock thanks for the nudge. I have fixed the issues noted by @hiranya911 , but haven't gotten the integration tests to work. I'll try those again over the weekend.

firebase_admin/_token_gen.py
```diff
@@ -84,11 +84,13 @@ class TokenGenerator:
 
     ID_TOOLKIT_URL = 'https://identitytoolkit.googleapis.com/v1'
 
-    def __init__(self, app, http_client):
+    def __init__(self, app, http_client, id_toolkit_url=None):
```
hiranya911, line 87: A little readability nit: Call the new method argument `url_override`.

```py
def __init__(self, app, http_client, url_override=None):
    url_prefix = url_override if url_override else self.ID_TOOLKIT_URL
    self.base_url = '{0}/projects/{1}'.format(url_prefix, app.project_id)
```

firebase_admin/_auth_client.py
```diff
@@ -36,17 +42,45 @@ def __init__(self, app, tenant_id=None):
...
+            else:
+                # Not emulated, no credential, raise
+                raise credential_error
+
         http_client = _http_client.JsonHttpClient(
             credential=credential, headers={'X-Client-Version': version_header})
 
         self._tenant_id = tenant_id
-        self._token_generator = _token_gen.TokenGenerator(app, http_client)
+        self._token_generator = _token_gen.TokenGenerator(
```
hiranya911, line 77: I'd recommend using keyword args for the last argument for clarity. But not a dealbreaker.

hiranya911 (changes requested):
I think the way credentials are being used is still not quite correct. But overall looks pretty good.

firebase_admin/_auth_client.py
```diff
@@ -36,17 +42,45 @@ def __init__(self, app, tenant_id=None):
...
+        emulator_host = os.environ.get(_EMULATOR_HOST_ENV_VAR)
+        if emulator_host:
+            if '//' in emulator_host:
+                raise ValueError(
+                    'Invalid {0}: "{1}". It must follow format "host:port".'.format(
+                        _EMULATOR_HOST_ENV_VAR, emulator_host))
+            base_url = 'http://{0}/identitytoolkit.googleapis.com'.format(emulator_host)
+            id_toolkit_endpoints['v1'] = base_url + '/v1'
+            id_toolkit_endpoints['v2beta1'] = base_url + '/v2beta1'
+
```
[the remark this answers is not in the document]
muru: @hiranya911  That was how I approached it at first, but then `TestImportUsers::test_import_users_http_error` failed (well, I guess it expects to fail, but only in a particular way). I am not entirely sure of the implications of adding another expected failure mode there - that could cause real problems to creep past unnoticed.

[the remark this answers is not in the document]
hiranya911: What's the exact test failure you're setting? Can you share some details or a stacktrace?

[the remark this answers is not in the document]
hiranya911: Ok, having looked at the source I think I know what the problem is. That test case is simulating a 401 error, which forces the credential to be refreshed, and that's resulting in an error. You need to use something like the following:

https://github.com/firebase/firebase-admin-python/blob/c10c7471eda060ef16a11e568e3e503316e67793/firebase_admin/db.py#L972-L978

I'd recommend moving that class into `_utils.py`, and reusing it for both auth and db.

hiranya911:
@muru the last set of commits have broken the CI. I'd recommend reverting the breaking commits, and doing the bare minimal necessary to address the earlier code review feedback. Any other new capabilities (e.g. emulator based testing), can be implemented in separate future PRs.

muru:
@hiranya911 porting the token verification code from https://github.com/firebase/firebase-admin-node/pull/1148 breaks a few unit tests. Since that should probably be included with this PR, would it be fine if I marked those tests for skipping when tested with the emulator? And while I am it, I could do the same for integration tests too.

hiranya911:
Ideally we shouldn't have to skip any unit tests. 

We can live without emulator-based integration tests for now (that work is on hold in Node.js at https://github.com/firebase/firebase-admin-node/pull/1155 due to some problems).

muru:
@hiranya911 here are the tests that fail with the emulator (mostly those concerned with verifying signatures, invalid tokens, expired cookies, and the OIDC/SAML providers):

<details><summary><code>tests/test_token_gen.py</code> (39 failures)</summary>

```
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_valid_params[auth_app1-Basic] - ValueError: Could not verify token signature.
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_valid_params[auth_app1-NoDevClaims] - ValueError: Could not verify token signature.
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_valid_params[auth_app1-EmptyDevClaims] - ValueError: Could not verify token signature.
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_noncert_credential[user_mgt_app0] - Failed: DID NOT RAISE <class 'ValueError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_revoked_token_check_revoked[user_mgt_app0-BinaryToken] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.RevokedIdToke...
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_revoked_token_check_revoked[user_mgt_app0-TextToken] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.RevokedIdTokenE...
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app0-NoKid] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app0-WrongKid] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app0-FutureToken] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app0-ExpiredToken] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_expired_token[user_mgt_app0] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.ExpiredIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_certificate_request_failure[user_mgt_app0] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.CertificateFetchError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app0-NoKid] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCookieError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app0-WrongKid] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCookieErr...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app0-FutureCookie] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCooki...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app0-ExpiredCookie] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCook...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_expired_cookie[user_mgt_app0] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.ExpiredSessionCookieError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_certificate_request_failure[user_mgt_app0] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.CertificateFetchErr...
FAILED tests/test_token_gen.py::TestCertificateCaching::test_certificate_caching[user_mgt_app0] - AssertionError: assert 0 == 1
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_noncert_credential[user_mgt_app1] - Failed: DID NOT RAISE <class 'ValueError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_revoked_token_check_revoked[user_mgt_app1-BinaryToken] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.RevokedIdToke...
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_revoked_token_check_revoked[user_mgt_app1-TextToken] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.RevokedIdTokenE...
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-NoKid] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-WrongKid] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-FutureToken] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-ExpiredToken] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_expired_token[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.ExpiredIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_certificate_request_failure[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.CertificateFetchError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-NoKid] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCookieError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-WrongKid] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCookieErr...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-FutureCookie] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCooki...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-ExpiredCookie] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCook...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_expired_cookie[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.ExpiredSessionCookieError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_certificate_request_failure[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.CertificateFetchErr...
FAILED tests/test_token_gen.py::TestCertificateCaching::test_certificate_caching[user_mgt_app1] - AssertionError: assert 0 == 1
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_sign_with_iam - AssertionError: assert False
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_sign_with_iam_error - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.TokenSignError'>
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_sign_with_discovered_service_account - AssertionError: assert False
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_sign_with_discovery_failure - Failed: DID NOT RAISE <class 'ValueError'>
```
</details>

<details><summary><code>integration/test_auth.py</code> (13 failures)</summary>

```
ERROR integration/test_auth.py::test_create_oidc_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
ERROR integration/test_auth.py::test_get_oidc_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
ERROR integration/test_auth.py::test_list_oidc_provider_configs - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
ERROR integration/test_auth.py::test_create_saml_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
ERROR integration/test_auth.py::test_get_saml_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
ERROR integration/test_auth.py::test_list_saml_provider_configs - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
FAILED integration/test_auth.py::test_custom_token_without_service_account - google.auth.exceptions.RefreshError: ('invalid_grant: Invalid grant: account not found', '{"error":"inval...
FAILED integration/test_auth.py::test_verify_id_token_revoked - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.RevokedIdTokenError'>
FAILED integration/test_auth.py::test_import_users_with_password - requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://localhost:9099/www.googleapis.com/iden...
FAILED integration/test_auth.py::test_update_oidc_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
FAILED integration/test_auth.py::test_delete_oidc_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
FAILED integration/test_auth.py::test_update_saml_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
FAILED integration/test_auth.py::test_delete_saml_provider_config - firebase_admin.exceptions.NotFoundError: Error while calling Auth service (Not Found).
```
</details>
I have split up the recent commit, putting the changes skipping these tests (note - skipped only when the emulator is used) as a separate commit. Please have a look when you have the time - if there are some tests which mustn't be skipped with the emulator, we can fix them now. I think I have also addressed the other review comments.

hiranya911:
Any idea how this happened?

```
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_noncert_credential[user_mgt_app0] - Failed: DID NOT RAISE <class 'ValueError'>
...
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_noncert_credential[user_mgt_app1] - Failed: DID NOT RAISE <class 'ValueError'>
```

Shouldn't one of them pass?

hiranya911 (changes requested):
Let's not implement integration testing for emulator mode yet. Lets focus on getting the implementation and the unit tests absolutely correct.

muru:
@hiranya911 what you say is correct, but shouldn't matter here, since emulation is set in the second/final iteration (https://github.com/firebase/firebase-admin-python/pull/531/files#diff-81622636c5d10b6ab40396c608887473086a43b94621195117193ab81a4b5cdaR166 - `emulated` is `False` first, then `True`).

These tests broke due to the changes made to token validation to fit the emulator (https://github.com/firebase/firebase-admin-python/pull/531/commits/aacd27fac95415fc42f74ec744ec0894231d704d) - they were supposed to raise an exception but didn't. Note that all tests ran perfectly fine before that.

hiranya911:
I'm debugging the tests in your branch and I can clearly see that tests that shouldn't be running in the emulator mode, picking up the emulator mode from another fixture. The test case `test_noncert_credential[user_mgt_app0]` should pass by raising the expected error, but it doesn't because it sees the env variable set in a different fixture, and goes into the emulator mode.

tests/test_token_gen.py
```diff
@@ -133,23 +141,39 @@ def _overwrite_iam_request(app, request):
     client = auth._get_client(app)
     client._token_generator.request = request
 
-@pytest.fixture(scope='module')
-def auth_app():
+
+def _is_emulated():
```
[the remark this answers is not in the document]
muru: The intention is that once the tests are fixed to accomodate the emulator, if the emulator variable name is changed, these tests should break. But I suppose the environment variable name could be a parameter for the function or something.

tests/test_token_gen.py
```diff
@@ -614,6 +665,8 @@ def test_certificate_request_failure(self, user_mgt_app):
 class TestCertificateCaching:
 
     def test_certificate_caching(self, user_mgt_app, httpserver):
+        if _is_emulated():
```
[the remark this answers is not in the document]
muru: Yes, right now I don't have the understanding of the emulator required to know which of these tests are actually important for the emulator - what behaviour is supported and expected and what behaviour is undefined. The failing tests are all marked skipped for the emulator case here for review (so that the reviewer doesn't have to jump back and forth between test results and code). If somebody can triage them and note the important ones, I can then prioritise fixing those.

integration/test_auth.py
```diff
@@ -32,11 +33,18 @@
 from firebase_admin import credentials
 
 
-_verify_token_url = 'https://www.googleapis.com/identitytoolkit/v3/relyingparty/verifyCustomToken'
-_verify_password_url = 'https://www.googleapis.com/identitytoolkit/v3/relyingparty/verifyPassword'
-_password_reset_url = 'https://www.googleapis.com/identitytoolkit/v3/relyingparty/resetPassword'
-_verify_email_url = 'https://www.googleapis.com/identitytoolkit/v3/relyingparty/setAccountInfo'
-_email_sign_in_url = 'https://www.googleapis.com/identitytoolkit/v3/relyingparty/emailLinkSignin'
+_EMULATOR_HOST_ENV_VAR = 'FIREBASE_AUTH_EMULATOR_HOST'
```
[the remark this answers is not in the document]
muru: I suppose the question here is whether the knowledge that some tests do pass worth marking the failing tests as being skipped for emulator (they should still be run for the "real" integration tests). At the least, this does reduce future work a bit since we'll already know which tests to go about fixing, and *might* catch a breakage in the meantime.

muru:
Ah, I hadn't noticed that, thanks! I'll look into how that can be fixed - maybe have an existing value of the environment variable override the fixture parameters.

hiranya911:
I think the easiest solution would be to change the fixture scope to `function` (which is also pytest default). That will cause a new app and a monkeypatch instance to be created per test function, and they will get cleaned up immediately after. App initialization and clean up are pretty cheap operations, so it should be ok in terms of speed.

integration/test_auth.py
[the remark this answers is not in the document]
hiranya911: It's a lot of knowledge to capture in one PR. So let's move that into a separate PR, and keep this one focused on the implementation.

We also encountered some flaky tests when we tried this in Node.js, which is why that PR is still on hold. So a passing integration test doesn't necessarily mean everything is ok either.

tests/test_token_gen.py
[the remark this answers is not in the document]
hiranya911: Ideally everything except a few should pass on both modes. Some of the ones that are not relevant in the emulator mode are:

* test_sign_with_iam
* test_sign_with_iam_error
* test_sign_with_discovered_service_account
* test_sign_with_discovery_failure
* test_certificate_request_failure
* test_certificate_caching

hiranya911:
With function-scoped fixtures (plus a few other minor modifications), these are all the remaining test failures:

```
FAILED tests/test_token_gen.py::TestCreateCustomToken::test_noncert_credential[user_mgt_app1] - Failed: DID NOT RAISE <class 'ValueError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_revoked_token_check_revoked[user_mgt_app1-BinaryToken] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.Revoke...
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_revoked_token_check_revoked[user_mgt_app1-TextToken] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.RevokedI...
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-NoKid] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-WrongKid] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-FutureToken] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_invalid_token[user_mgt_app1-ExpiredToken] - Failed: DID NOT RAISE <class 'firebase_admin._auth_utils.InvalidIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_expired_token[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.ExpiredIdTokenError'>
FAILED tests/test_token_gen.py::TestVerifyIdToken::test_certificate_request_failure[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.CertificateFetchEr...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-NoKid] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCooki...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-WrongKid] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessionCo...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-FutureCookie] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSessi...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_invalid_cookie[user_mgt_app1-ExpiredCookie] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.InvalidSess...
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_expired_cookie[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.ExpiredSessionCookieError'>
FAILED tests/test_token_gen.py::TestVerifySessionCookie::test_certificate_request_failure[user_mgt_app1] - Failed: DID NOT RAISE <class 'firebase_admin._token_gen.CertificateF...
FAILED tests/test_token_gen.py::TestCertificateCaching::test_certificate_caching[user_mgt_app1] - assert 0 == 1
```

It's unfortunate our implementation doesn't handle expired emulated tokens correctly (this is because we bypass the JWT verification step entirely for emulated tokens). But this might be ok. I'm fine with skipping the above tests if we must.

muru:
@hiranya911 I removed the integration test and modified the fixtures to use the `function` scope, and where possible, I modified the tests to account for the current behaviour of the emulator mode instead of skipping them. If the emulator mode is improved, that might help prevent tests from being overlooked due to being skipped.

Also, I opted against using a module-level variable for `_auth_utils.is_emulated()` as it behaves like module fixture when a variable, causing test failures. Is that fine?

tests/test_token_gen.py
```diff
@@ -286,6 +347,14 @@ def test_sign_with_discovery_failure(self):
     def _verify_signer(self, token, signer):
         segments = token.split('.')
         assert len(segments) == 3
+        # See https://github.com/googleapis/google-auth-library-python/pull/324
```
hiranya911, line 350: What if we just did a `jwt.decode(verify=False)` here instead?

tests/test_token_gen.py
```diff
@@ -212,20 +244,30 @@ def test_invalid_params(self, auth_app, values):
...
+            # test_sign_with_iam below.
+            custom_token = auth.create_custom_token(MOCK_UID, app=user_mgt_app).decode()
+            self._verify_signer(custom_token, _token_gen.AUTH_EMULATOR_EMAIL)
+            return
         with pytest.raises(ValueError):
             auth.create_custom_token(MOCK_UID, app=user_mgt_app)
 
     def test_sign_with_iam(self):
-        options = {'serviceAccountId': 'test-service-account', 'projectId': 'mock-project-id'}
+        signer = _token_gen.AUTH_EMULATOR_EMAIL if _is_emulated() else 'test-service-account'
```
hiranya911, line 257: This test doesn't use any of the fixtures. Does it ever run in the emulator mode?

Update: I see a few other tests in this file that doesn't depend on the fixtures, and therefore never runs in the emulator mode. Let's not modify them.

tests/test_token_gen.py
```diff
@@ -440,17 +515,31 @@ def test_invalid_arg(self, user_mgt_app, id_token):
...
+            assert claims['uid'] == claims['sub']
+            return
         _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
         with pytest.raises(auth.InvalidIdTokenError) as excinfo:
             auth.verify_id_token(id_token, app=user_mgt_app)
         assert isinstance(excinfo.value, exceptions.InvalidArgumentError)
         assert excinfo.value.http_response is None
 
     def test_expired_token(self, user_mgt_app):
+        _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
```
hiranya911, line 534: Redundant?

tests/test_token_gen.py
```diff
@@ -440,17 +515,31 @@ def test_invalid_arg(self, user_mgt_app, id_token):
             auth.verify_id_token(id_token, app=user_mgt_app)
         assert 'Illegal ID token provided' in str(excinfo.value)
 
-    @pytest.mark.parametrize('id_token', invalid_tokens.values(), ids=list(invalid_tokens))
-    def test_invalid_token(self, user_mgt_app, id_token):
+    @pytest.mark.parametrize('id_token_key', list(invalid_tokens))
+    def test_invalid_token(self, user_mgt_app, id_token_key):
+        id_token = self.invalid_tokens[id_token_key]
+        if _is_emulated() and id_token_key in self.tokens_not_invalid_in_emulator:
+            # These tokens won't be invalid when using the emulator, check them like valid tokens.
```
hiranya911, line 522: There's quite a bit of duplication and commentary here. Let's try to reduce that as much as possible. For example:

```
if _is_emulated() and id_token_key in self.tokens_not_invalid_in_emulator:
    return _assert_valid_in_emulator(id_token, user_mgt_app)
```

Then you can use the same helper function in places like `test_expired_token`.

tests/test_token_gen.py
```diff
@@ -560,8 +659,15 @@ def test_invalid_args(self, user_mgt_app, cookie):
             auth.verify_session_cookie(cookie, app=user_mgt_app)
         assert 'Illegal session cookie provided' in str(excinfo.value)
 
-    @pytest.mark.parametrize('cookie', invalid_cookies.values(), ids=list(invalid_cookies))
-    def test_invalid_cookie(self, user_mgt_app, cookie):
+    @pytest.mark.parametrize('cookie_key', list(invalid_cookies))
+    def test_invalid_cookie(self, user_mgt_app, cookie_key):
+        cookie = self.invalid_cookies[cookie_key]
+        if _is_emulated() and cookie_key in self.cookies_not_invalid_in_emulator:
+            # These cookies won't be invalid when using the emulator, check them like valid cookies.
```
hiranya911, line 666: Same here. Extract duplicate code into helper functions. Avoid comments.

hiranya911 (changes requested):
Thanks for making the changes @muru. I think this looks pretty reasonable. I've left some comments on how to further clean up and reduce some of the code duplication.

firebase_admin/_token_gen.py
```diff
@@ -291,14 +311,16 @@ def verify(self, token, request):
                 '{0} expects {1}, but was given a custom '
                 'token.'.format(self.operation, self.articled_short_name))
         elif not header.get('kid'):
+            if _auth_utils.is_emulated():
```
hiranya911, line 314: Use a local variable please. Do not call `is_emulated()` more than once per verification.

tests/test_token_gen.py
muru: It might not use those fixtures, but the presence of those fixtures probably affects it via the environment variable. At any rate, I did get failures for this and the tests below, where the email address in the assertion failure definitely was the one used by the emulator code.

hiranya911: That sounds like a bug to me. With function-scoped fixtures, these tests should never see the emulator mode enabled.

muru: Ah, yes, with the change to function-scoped fixtures, they no longer see emulator mode enabled.

firebase_admin/_token_gen.py
```diff
@@ -285,20 +305,24 @@ def verify(self, token, request):
...
 
+        emulated = _auth_utils.is_emulated()
+
         error_message = None
         if audience == FIREBASE_AUDIENCE:
             error_message = (
                 '{0} expects {1}, but was given a custom '
                 'token.'.format(self.operation, self.articled_short_name))
         elif not header.get('kid'):
+            if emulated:
```
hiranya911, line 316:
```
elif not emulated and not header.get('kid'):
```

tests/test_token_gen.py
```diff
@@ -522,20 +579,27 @@ class TestVerifySessionCookie:
         'IDToken': TEST_ID_TOKEN,
     }
 
+    cookies_not_invalid_in_emulator = [
```
hiranya911, line 582: Let's call this `invalid_cookies_accepted_in_emulator` to avoid the double negative.

tests/test_token_gen.py
```diff
@@ -388,14 +425,23 @@ class TestVerifyIdToken:
         'BadFormatToken': 'foobar'
     }
 
-    @pytest.mark.parametrize('id_token', valid_tokens.values(), ids=list(valid_tokens))
-    def test_valid_token(self, user_mgt_app, id_token):
-        _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
-        claims = auth.verify_id_token(id_token, app=user_mgt_app)
+    tokens_not_invalid_in_emulator = [
```
hiranya911, line 428: invalid_tokens_accepted_in_emulator

tests/test_user_mgt.py
```diff
@@ -50,17 +50,30 @@
...
+EMULATED_ID_TOOLKIT_URL = 'http://{}/identitytoolkit.googleapis.com/v1'.format(AUTH_EMULATOR_HOST)
+URL_PROJECT_SUFFIX = '/projects/mock-project-id'
+USER_MGT_URLS = {
+    'ID_TOOLKIT': ID_TOOLKIT_URL,
+    'PREFIX': ID_TOOLKIT_URL + URL_PROJECT_SUFFIX,
+}
 
-@pytest.fixture(scope='module')
-def user_mgt_app():
+@pytest.fixture(scope='module', params=[{'emulated': False}, {'emulated': True}])
```
hiranya911, line 65: To be on the safe side should we change this to function scope as well?

hiranya911 (changes requested):
Thanks @muru. This looks great! I only had a few minor comments, and we can merge once those are addressed.

tests/test_token_gen.py
muru: That's what I originally had, but it breaks the linter as the name's too long:
```
tests/test_token_gen.py:428:4: C0103: Class attribute name "invalid_tokens_accepted_in_emulator" doesn't conform to '([A-Za-z_][A-Za-z0-9_]{2,30}|(__.*__))$' pattern (invalid-name)
tests/test_token_gen.py:583:4: C0103: Class attribute name "invalid_cookies_accepted_in_emulator" doesn't conform to '([A-Za-z_][A-Za-z0-9_]{2,30}|(__.*__))$' pattern (invalid-name)
```

(Same problem with the slightly shorter variant `invalid_tokens_valid_in_emulator`)

tests/test_user_mgt.py
muru: Done. Test times went from 7-7.5s to 7.5-8s. Not much of a problem, I think.

tests/test_token_gen.py
hiranya911: How about just `cookies_accepted_in_emulator` and `tokens_accepted_in_emulator`?

muru: Still feels not quite right to me, but I don't want to bikeshed on this, so `*_accepted_in_emulator` it is.

hiranya911 (approved):
Thanks @muru for patiently working through all the comments. This LGTM 👍

tests/test_token_gen.py
hiranya911: I understand the concern. It's a compromise for sure.

Issue #535, opened 2021-03-10
Auth call timeout option


gzamb:
Using firebase-admin sdk v4.0.0 for python3.8. 
We have recently noticed that one of our time sensitive services is struggling due to long calls on `auth.verify_id_token` function.  We'd like the option to have a timeout for the actual request call that happens here, since it does appear that the google-auth library allows you to set a [timeout](https://google-auth.readthedocs.io/en/latest/reference/google.auth.transport.requests.html#module-google.auth.transport.requests) to the request.  We do use the `httpTimeout` on initialization of the app already but it appears this config option is not used at all when making the request.

hiranya911:
We don't make any direct HTTP requests in `verify_id_token()`. We simply call [`google.oauth2.id_token.verify_token()`](https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.id_token.html#google.oauth2.id_token.verify_token). It is that library that makes the request to [fetch certificates](https://google-auth.readthedocs.io/en/latest/_modules/google/oauth2/id_token.html#_fetch_certs).

We do pass in a `request` object (which wraps a Session instance) when making the above call, but unfortunately the `requests.Session` API doesn't support injecting a timeout as far as I can tell. We will probably have to implement a [workaround](https://stackoverflow.com/questions/41295142/is-there-a-way-to-globally-override-requests-timeout-setting).

daniellehanks:
It looks like google.oauth2.id_token.verify_token() just takes a generic `google.auth.transport.Request` interface. Swapping out `requests` for `urllib3`, which does support setting a global timeout on its [connection pools](https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html) could be another option?

daniellehanks:
Additionally, it looks like [this call](https://github.com/firebase/firebase-admin-python/blob/873aa7c10f32d1f07ce28dde6dc5414e7583afd3/firebase_admin/_user_mgt.py#L596) (made when `check_revoked=True`) does not pass a timeout either?

hiranya911:
> Swapping out requests for urllib3, which does support setting a global timeout on its connection pools could be another option

I'd rather implement one of the workarounds (e.g. subclassing the `requests.Session`) than taking a direct dependency on urllib3. 

> Additionally, it looks like this call (made when check_revoked=True) does not pass a timeout either?

That call goes through our own `_http_client.HttpClient` which sets a default timeout on all requests.

[...]

PR #536, opened 2021-03-13
fix(auth): Make auth client respect app options httpTimeout

daniellehanks:
https://github.com/firebase/firebase-admin-python/issues/535

This takes care of half of the above ticket, the other half being to specify a timeout on the http transport given to google.oauth2.id_token.verify_token().

I tried to stay locally consistent and tried not to change the test setup too much.

### DIFF
--- a/tests/test_user_mgt.py
+++ b/tests/test_user_mgt.py
@@ -52,11 +52,23 @@ MOCK_ACTION_CODE_SETTINGS = auth.ActionCodeSettings(**MOCK_ACTION_CODE_DATA)
 
 USER_MGT_URL_PREFIX = 'https://identitytoolkit.googleapis.com/v1/projects/mock-project-id'
 
+TEST_TIMEOUT = 42
+
 
 @pytest.fixture(scope='module')
 def user_mgt_app():
     app = firebase_admin.initialize_app(testutils.MockCredential(), name='userMgt',
                                         options={'projectId': 'mock-project-id'})
+    yield app
+    firebase_admin.delete_app(app)
+
+@pytest.fixture(scope='module')
+def user_mgt_app_with_timeout():
+    app = firebase_admin.initialize_app(
+        testutils.MockCredential(),
+        name='userMgtTimeout',
+        options={'projectId': 'mock-project-id', 'httpTimeout': TEST_TIMEOUT}
+    )
     yield app
     firebase_admin.delete_app(app)
 
@@ -105,7 +117,7 @@ def _check_user_record(user, expected_uid='testuser'):
     assert provider.provider_id == 'phone'
 
 
-def _check_request(recorder, want_url, want_body=None):
+def _check_request(recorder, want_url, want_body=None, want_timeout=None):
     assert len(recorder) == 1
     req = recorder[0]
     assert req.method == 'POST'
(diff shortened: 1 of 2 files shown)


tests/test_user_mgt.py
```diff
@@ -105,14 +117,16 @@ def _check_user_record(user, expected_uid='testuser'):
...
+def _check_request(recorder, want_url, want_body=None, want_timeout=None):
     assert len(recorder) == 1
     req = recorder[0]
     assert req.method == 'POST'
     assert req.url == '{0}{1}'.format(USER_MGT_URL_PREFIX, want_url)
     if want_body:
         body = json.loads(req.body.decode())
         assert body == want_body
+    if want_timeout:
+        assert recorder[0]._extra_kwargs['timeout'] == pytest.approx(want_timeout, 0.001)
```
hiranya911, line 129: Should we add the following assertion?

```py
timeout_arg = recorder[0]._extra_kwargs['timeout']
if want_timeout:
    assert timeout_arg == pytest.approx(want_timeout, 0.01)
else:
    assert timeout_arg == pytest.approx(_http_client.DEFAULT_TIMEOUT, 0.01)
```

hiranya911 (approved):
Thanks for the PR @daniellehanks. Looks pretty good to me. One suggestion on tests, but we can merge without that change too. Let me know what you think.

tests/test_user_mgt.py
daniellehanks: The problem I see with that is it would force any test not using the default to pass that argument. I aligned with the other `want` args, if it's specified, check it, otherwise don't. That said, it's your code base so I defer to you. Lmk if you want me to change it.

hiranya911: > The problem I see with that is it would force any test not using the default to pass that argument.

It doesn't. It can default to `None`, in which case we assert against `_http_client.DEFAULT_TIMEOUT`.

daniellehanks: That's what I'm saying, if someone added a test using `user_mgt_app_with_timeout`, but didn't pass a timeout to this function (presumably because they didn't care about asserting the value), it would fail.

hiranya911: In the implementation we already default the client to `_http_client.DEFAULT_TIMEOUT`. In what scenario do we expect that assertion to not hold?

daniellehanks: If an app is used with a non-default timeout, but someone is either unaware of the `want_timeout` parameter or explicitly does not wish to make an assertion around it. Just biasing towards explicit > implicit [per BDFL](https://www.python.org/dev/peps/pep-0020/). Implementing the suggested behavior would implicitly check timeout. It also adds assertions to existing tests, whereas the original implementation results in no change to existing tests/assertions.

For a concrete example, if I removed the `TEST_TIMEOUT` from `test_get_user_with_timeout`, the test would fail, but, IMO, it would not be readily obvious just inspecting the test function why (nowhere did I make an assertion or ask an assertion to be made around timeout, so why is that happening?).

hiranya911: > It also adds assertions to existing tests, whereas the original implementation results in no change to existing tests/assertions.

That was my intended goal. Ideally, the `_check_request()` helper function should make assertions about the traits that are common to all requests made by the auth client. We probably should have had the timeout assertion there to begin with. The `want_body` shouldn't be an optional argument. It's always specified.

Anyway, all these changes can probably be lumped into a separate PR. For now I'm happy with where this stands.

PR #538, opened 2021-03-23
fix(auth): Setting httpTimeout on certificate fetch requests

### LINKED ISSUE #535: Auth call timeout option

hiranya911:
Passing the `httpTimeout` option to the request instance used to fetch public key certificates by `verify_id_token()` and `verify_session_cookie()` APIs.

Resolves #535

### DIFF
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -29,6 +29,7 @@ import google.oauth2.service_account
 
 from firebase_admin import exceptions
 from firebase_admin import _auth_utils
+from firebase_admin import _http_client
 
 
 # ID token constants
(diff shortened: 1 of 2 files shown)


lahirumaramba (approved):
LGTM!

PR #541, opened 2021-04-12
fix: Using alg=none header for custom tokens in emulator mode

hiranya911:
Setting the `alg` header to `none` when creating custom tokens for the emulator. This requires a recent bug fix in the google-auth package: https://github.com/googleapis/google-auth-library-python/pull/729

To enable that I'm also upgrading the corresponding dependencies.

### DIFF
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -53,6 +53,8 @@ RESERVED_CLAIMS = set([
 ])
 METADATA_SERVICE_URL = ('http://metadata.google.internal/computeMetadata/v1/instance/'
                         'service-accounts/default/email')
+ALGORITHM_RS256 = 'RS256'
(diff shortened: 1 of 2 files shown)


hiranya911:
This is unfortunate. Looks like google-auth discontinued Python 3.5 support a few minor versions ago: https://github.com/googleapis/google-auth-library-python/releases/tag/v1.24.0

hiranya911:
Tried to work around that issue in the tests, but it's not quite working out. Turns out google-cloud-firestore has also dropped Python 3.5 support as of their v2 release. I think we have to just drop Python 3.5 support before we can go ahead. 

@lahirumaramba FYA

hiranya911:
Rebased on the head of main branch.

lahirumaramba (approved):
LGTM!

PR #542, opened 2021-04-13
change: Drop Python 3.5 support

### LINKED ISSUE #522: EOL for Python 3.5
As per this release note for [3.5.10](https://www.python.org/downloads/release/python-3510/), v3.5 has reached end of life.

lahirumaramba:
- Drop Python 3.5 support (Python 3.5 has reached EoL)
- Resolves #522 

RELEASE NOTE: Dropped support for Python 3.5. Developers should to use Python 3.6 or higher when deploying the Admin SDK.

### DIFF
--- a/tests/testutils.py
+++ b/tests/testutils.py
@@ -16,7 +16,7 @@
 import io
 import os
 
-import pytest
+from _pytest.monkeypatch import MonkeyPatch
 
 from google.auth import credentials
 from google.auth import transport
@@ -61,12 +61,7 @@ def run_without_project_id(func):
 
 
 def new_monkeypatch():
-    try:
-        return pytest.MonkeyPatch()
-    except AttributeError:
-        # Fallback for Python 3.5
-        from _pytest.monkeypatch import MonkeyPatch
-        return MonkeyPatch()
+    return MonkeyPatch()
 
 
 class MockResponse(transport.Response):
--- a/tests/testutils.py
+++ b/tests/testutils.py
@@ -16,7 +16,7 @@
 import io
 import os
 
-from _pytest.monkeypatch import MonkeyPatch
+import pytest
 
 from google.auth import credentials
 from google.auth import transport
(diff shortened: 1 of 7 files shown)


tests/testutils.py
```diff
@@ -61,12 +61,7 @@ def run_without_project_id(func):
 
 
 def new_monkeypatch():
-    try:
-        return pytest.MonkeyPatch()
-    except AttributeError:
-        # Fallback for Python 3.5
-        from _pytest.monkeypatch import MonkeyPatch
-        return MonkeyPatch()
+    return MonkeyPatch()
```
hiranya911, line 64: Why not just `return pytest.MonkeyPatch()`?

hiranya911 (changes requested):
Looks good. I only had one suggestion.

Have you thought about how to release/version this? In the past we have bumped major version numbers when we discontinued support for a platform version. But GCP is following a more relaxed policy with respect to that (albeit unofficially).

tests/testutils.py
lahirumaramba: Looks like `pytest.MonkeyPatch()` works for `pytest` 6.2 and up. Our CIs use the latest version so I think it would be safe to update this code.

lahirumaramba: Updated!

hiranya911: Should we just set pytest version to 6.2 and up in requirements.txt file?

hiranya911 (approved):
LGTM 👍 

My personal preference is to bump the major version as well. GCP policy on this issue is still a work in progress, and might take a while to materialize. In the meantime, I'd be ok with bumping the major version for this change.

tests/testutils.py
lahirumaramba: Good idea! Updated.

lahirumaramba:
> LGTM 👍
> 
> My personal preference is to bump the major version as well. GCP policy on this issue is still a work in progress, and might take a while to materialize. In the meantime, I'd be ok with bumping the major version for this change.

Sounds good! Let's do a major version bump for this then. Thanks!

Issue #507, opened 2020-12-01
Messaging: Errors with multicast send


aplusm:
* App Engine environment
  * firebase-admin:  4.4.0
  * Firebase Product: Messaging

### Describe the problem

From time to time, while sending multicast notification from our app engine application, the following issues are reported, all at the same time:

```
KeyError: '8'
at execute (/layers/google.python.pip/pip/googleapiclient/http.py:1548)
at positional_wrapper (/layers/google.python.pip/pip/googleapiclient/_helpers.py:134)
at send_all (/layers/google.python.pip/pip/firebase_admin/messaging.py:390)
at send_multicast (/layers/google.python.pip/pip/firebase_admin/messaging.py:167)
at send (/srv/main.py:109)
at notification_send (/srv/main.py:55)
at dispatch_request (/layers/google.python.pip/pip/flask/app.py:1936)
at full_dispatch_request (/layers/google.python.pip/pip/flask/app.py:1950)
at reraise (/layers/google.python.pip/pip/flask/_compat.py:39)
at handle_user_exception (/layers/google.python.pip/pip/flask/app.py:1821)
at full_dispatch_request (/layers/google.python.pip/pip/flask/app.py:1952)
at wsgi_app (/layers/google.python.pip/pip/flask/app.py:2447)
```

```
OSError: [Errno 0] Error
at read (/opt/python3.8/lib/python3.8/ssl.py:1099)
at recv_into (/opt/python3.8/lib/python3.8/ssl.py:1241)
at readinto (/opt/python3.8/lib/python3.8/socket.py:669)
at _read_status (/opt/python3.8/lib/python3.8/http/client.py:268)
at begin (/opt/python3.8/lib/python3.8/http/client.py:307)
at getresponse (/opt/python3.8/lib/python3.8/http/client.py:1347)
at _conn_request (/layers/google.python.pip/pip/httplib2/__init__.py:1589)
at _request (/layers/google.python.pip/pip/httplib2/__init__.py:1650)
at request (/layers/google.python.pip/pip/httplib2/__init__.py:1985)
at request (/layers/google.python.pip/pip/google_auth_httplib2.py:200)
at _execute (/layers/google.python.pip/pip/googleapiclient/http.py:1469)
at execute (/layers/google.python.pip/pip/googleapiclient/http.py:1540)
at positional_wrapper (/layers/google.python.pip/pip/googleapiclient/_helpers.py:134)
[...]

PR #558, opened 2021-06-24
fix(fcm): A workaround for the concurrency issues in googleapiclient

hiranya911:
This is a possible solution to the problem outlined in #507. I wasn't quite able to repro the exception in the bug report, but the threading issues in `googleapiclient` are well known, and I can easily reproduce some concurrency issues by calling `send_all()` from just 3-4 parallel threads. The proposed solution is to instantiate a new transport instance for each invocation of `send_all()` (as opposed to using a shared transport across calls), and tests indeed confirm that it resolves the issues I managed to reproduce.

### DIFF
--- a/firebase_admin/messaging.py
+++ b/firebase_admin/messaging.py
@@ -330,9 +330,9 @@ class _MessagingService:
             'X-FIREBASE-CLIENT': 'fire-admin-python/{0}'.format(firebase_admin.__version__),
         }
         timeout = app.options.get('httpTimeout', _http_client.DEFAULT_TIMEOUT_SECONDS)
-        self._client = _http_client.JsonHttpClient(
-            credential=app.credential.get_credential(), timeout=timeout)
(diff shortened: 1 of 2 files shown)


firebase_admin/messaging.py
```diff
@@ -373,10 +373,11 @@ def batch_callback(_, response, error):
 
         batch = http.BatchHttpRequest(
             callback=batch_callback, batch_uri=_MessagingService.FCM_BATCH_URL)
+        transport = self._build_transport(self._credential)
```
lahirumaramba, line 376: Is the reason to use `_build_trasnport` here (instead of directly using `_auth.authorized_http`) so we can write tests easily?

lahirumaramba (approved):
Using a new transport makes sense to me. Thanks!

firebase_admin/messaging.py
hiranya911: Yes.

PR #559, opened 2021-07-06
correct `**kwargs` documentation in docstrings

skarzi:
Hi :wave: 

first of all thanks for maintaining and developing this great SDK!

This PR fixes `kwargs` documentation in `Args` section of docstrings.
According to Google Python Style Guide `*args` and `**kwargs` should be documented in [`Args`](https://google.github.io/styleguide/pyguide.html#doc-function-args) section using following format:

> If a function accepts *foo (variable length argument lists) and/or **bar (arbitrary keyword arguments), they should be listed as *foo and **bar.

### DIFF
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -284,7 +284,7 @@ class Client:
         """Creates a new user account with the specified properties.
 
         Args:
-            kwargs: A series of keyword arguments (optional).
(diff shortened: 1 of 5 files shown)


skarzi:
@googlebot I signed it!

lahirumaramba:
Thank you @skarzi ! LGTM
Adding @egilmorez as this might update reference docs.

lahirumaramba (approved):

PR #565, opened 2021-08-04
fix(auth): check if user disabled on check_revoked

bojeil-google:
When `verify_session_cookie` or `verify_id_token` is called with
`check_revoked` set to `True` we should also check if the user is
disabled.

If disabled the `UserDisabledError` is raised.

### DIFF
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -100,7 +100,8 @@ class Client:
 
         Args:
             id_token: A string of the encoded JWT.
-            check_revoked: Boolean, If true, checks whether the token has been revoked (optional).
+            check_revoked: Boolean, If true, checks whether the token has been revoked or
+                the user disabled (optional).
 
         Returns:
             dict: A dictionary of key-value pairs parsed from the decoded JWT.
@@ -115,6 +116,8 @@ class Client:
                 this ``Client`` instance.
             CertificateFetchError: If an error occurs while fetching the public key certificates
                 required to verify the ID token.
+            UserDisabledError: If ``check_revoked`` is ``True`` and the corresponding user
+                record is disabled.
         """
         if not isinstance(check_revoked, bool):
             # guard against accidental wrong assignment.
@@ -129,7 +132,8 @@ class Client:
                     'Invalid tenant ID: {0}'.format(token_tenant_id))
 
         if check_revoked:
-            self._check_jwt_revoked(verified_claims, _token_gen.RevokedIdTokenError, 'ID token')
+            self._check_jwt_revoked_or_disabled(
+                verified_claims, _token_gen.RevokedIdTokenError, 'ID token')
         return verified_claims
 
     def revoke_refresh_tokens(self, uid):
@@ -720,7 +724,9 @@ class Client:
         """
         return self._provider_manager.list_saml_provider_configs(page_token, max_results)
 
-    def _check_jwt_revoked(self, verified_claims, exc_type, label):
+    def _check_jwt_revoked_or_disabled(self, verified_claims, exc_type, label):
         user = self.get_user(verified_claims.get('uid'))
         if verified_claims.get('iat') * 1000 < user.tokens_valid_after_timestamp:
             raise exc_type('The Firebase {0} has been revoked.'.format(label))
+        if user.disabled:
+            raise _auth_utils.UserDisabledError('The user record is disabled.')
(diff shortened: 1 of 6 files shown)


lahirumaramba (approved):
Thanks! LGTM!

bojeil-google:
> Thanks! LGTM!

Thanks for the quick review. we are planning to fix this in Node.js, Java and Golang too.

firebase_admin/_auth_client.py
```diff
@@ -720,7 +724,9 @@ def list_saml_provider_configs(
         """
         return self._provider_manager.list_saml_provider_configs(page_token, max_results)
 
-    def _check_jwt_revoked(self, verified_claims, exc_type, label):
+    def _check_jwt_revoked_or_disabled(self, verified_claims, exc_type, label):
         user = self.get_user(verified_claims.get('uid'))
         if verified_claims.get('iat') * 1000 < user.tokens_valid_after_timestamp:
             raise exc_type('The Firebase {0} has been revoked.'.format(label))
+        if user.disabled:
```
lahirumaramba, line 731: Should `user.disabled` takes precedent over revoked token here? If the user is disabled, do we need to check the token at all? :)

lahirumaramba:
Added @egilmorez for docs changes. Thanks!

firebase_admin/_auth_client.py
lahirumaramba, line 731: @bojeil-google: Should `user.disabled` take precedence over revoked token here? If the user is disabled, do we need to check the token at all? :)

hiranya911 (changes requested):
Looks mostly good. Just had one suggestion about the base type for the new `UserDisabledError` class.

firebase_admin/_auth_client.py
bojeil-google: Good point. I moved the check. I also added a test to confirm the disabled check is done before.

tests/test_token_gen.py
```diff
@@ -208,6 +208,12 @@ def revoked_tokens():
     mock_user['users'][0]['validSince'] = str(int(time.time())+100)
     return json.dumps(mock_user)
 
+@pytest.fixture(scope='module')
+def get_user_disabled_response():
```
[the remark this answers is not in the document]
bojeil-google: Done

firebase_admin/_auth_utils.py
```diff
@@ -385,6 +385,15 @@ def __init__(self, message, cause=None, http_response=None):
         exceptions.NotFoundError.__init__(self, message, cause, http_response)
 
 
+class UserDisabledError(exceptions.FailedPreconditionError):
```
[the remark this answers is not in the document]
bojeil-google: Since revoked token error extends `exceptions.InvalidArgumentError`, I have updated `UserDisabledError` to also extend `exceptions.InvalidArgumentError`.

I prefer keeping the error in `_auth_utils.py` since there is also a backend error code `USER_DISABLED` which is currently used for client operation errors but could be re-used in a different context for the admin sdk.

tests/test_token_gen.py
```diff
@@ -471,6 +484,24 @@ def test_revoked_token_check_revoked(self, user_mgt_app, revoked_tokens, id_toke
...
+    @pytest.mark.parametrize('id_token', valid_tokens.values(), ids=list(valid_tokens))
+    def test_disabled_user_check_revoked(self, user_mgt_app, user_disabled, id_token):
+        _overwrite_cert_request(user_mgt_app, MOCK_REQUEST)
+        _instrument_user_manager(user_mgt_app, 200, user_disabled)
+        with pytest.raises(auth.UserDisabledError) as excinfo:
+            auth.verify_id_token(id_token, app=user_mgt_app, check_revoked=True)
+        assert str(excinfo.value) == 'The user record is disabled.'
+
+    @pytest.mark.parametrize('id_token', valid_tokens.values(), ids=list(valid_tokens))
+    def test_disabled_revoked_check_revoked(
```
hiranya911, line 496: Nit: We can call this `test_check_disabled_before_revoked()` and drop the comment if you like.

hiranya911 (approved):
LGTM with a suggestion.

egilmorez (approved):
Doc strings look good, thanks!

tests/test_token_gen.py
bojeil-google: Done

PR #568, opened 2021-08-13
Maceo regapic appcheck

maceonthompson:
Proof of concept demonstrating the ease of using a [REGAPIC](https://github.com/googleapis/gapic-generator) generated client to create a new API surface (in this case, Appcheck).
Definitely not prod ready, but shows that this works!


GabrieleCalarota:
What's missing for this to be released? I really would like to have in firebase-admin the possibility to have appCheck as for NodeJs

PR #572, opened 2021-09-08
fix(auth): Speed up the PageIterator by evaluating items once per page

### LINKED ISSUE #571: Very slow performance with firebase_admin.auth.list_users().iterate_all()
* Operating System version: n/a
  * Firebase SDK version: firebase-admin==5.0.2 
  * Firebase Product: auth, user management
  * Python version: 3.8
  * Pip version: n/a

### [REQUIRED] Step 3: Describe the problem

When fetching all user accounts using the firebase_admin.auth.list_users().iterate_all() helper, performance is very poor, taking about 5 seconds to fetch 1000 users.

I have a project which has several million users. There is a scheduled task that fetches all the users for processing, but the task takes way too long and runs into problems with App Engine request timeouts.

#### Steps to reproduce:

Here is an App Engine standard + Python 3.8 project which has 2 routes to show the difference between iterating through all users using the iterate all helper, versus fetching each page directly:

https://gist.github.com/davidwtbuxton/c6bbdc1e93f686cae931d1e38b5bd555

The code logs a message for every 1000 users. In my testing it takes about 5 seconds for every 1000 users on the slow path (using iterate_all) and about 1 second for every 1000 users on the fast path (not using iterate_all).

#### Relevant Code:

I've got a WIP patch here https://github.com/davidwtbuxton/firebase-admin-python/commit/20991aafcb007a9d4c7fc4bd8589ebff5de1ab16

I need to test that WIP patch some more, but would this be useful? Suggestions for improving the patch?

I think the cause of the poor performance is the [PageIterator class][1] and how next() is implemented. In the next() method, the code evaluates `self.items` 3 times. For the [_UserIterator class][2] the items list is actually a computed property, which accesses [ListUsersPage.users][3], which is also a computed property that builds a new list of user records.

So for each user in the list of users, the iterator ends up triggering a list of all users in the current page to be constructed 3 times, just to access 1 user. When a page has 1000 users, this ends up being much slower (1000 x 3) than the ideal case of constructing the list just 1 time for each page.

[1]: https://github.com/firebase/firebase-admin-python/blob/01db7eb8da6094e09fc0311930718deec5ccd4ad/firebase_admin/_auth_utils.py#L35
[2]: https://github.com/firebase/firebase-admin-python/blob/01db7eb8da6094e09fc0311930718deec5ccd4ad/firebase_admin/_user_mgt.py#L845
[3]: https://github.com/firebase/firebase-admin-python/blob/01db7eb8da6094e09fc0311930718deec5ccd4ad/firebase_admin/_user_mgt.py#L357

davidwtbuxton:
The `firebase_admin.auth.list_users().iterate_all()` method uses a sub-class
of PageIterator, which happens to access the .items computed property more
than once for every page of results. This has been changed so we take care
not to access the `self.items` property more than once per page.

This is a lot faster.

Fixes #571

### DIFF
--- a/firebase_admin/_auth_utils.py
+++ b/firebase_admin/_auth_utils.py
@@ -43,29 +43,31 @@ class PageIterator:
     def __init__(self, current_page):
         if not current_page:
             raise ValueError('Current page must not be None.')
-        self._current_page = current_page
-        self._index = 0
 
-    def next(self):
-        if self._index == len(self.items):
+        self._current_page = current_page
+        self._iter = None
+
+    def __next__(self):
+        if self._iter is None:
+            self._iter = iter(self.items)
+
+        try:
+            return next(self._iter)
+        except StopIteration:
             if self._current_page.has_next_page:
                 self._current_page = self._current_page.get_next_page()
-                self._index = 0
-        if self._index < len(self.items):
-            result = self.items[self._index]
-            self._index += 1
-            return result
-        raise StopIteration
+                self._iter = iter(self.items)
+
+                return next(self._iter)
+
+            raise
+
+    def __iter__(self):
+        return self
 
     @property
     def items(self):
         raise NotImplementedError
-
-    def __next__(self):
-        return self.next()
-
-    def __iter__(self):
-        return self
 
 
 def get_emulator_host():


firebase_admin/_auth_utils.py
```diff
@@ -43,30 +43,29 @@ class PageIterator:
     def __init__(self, current_page):
         if not current_page:
             raise ValueError('Current page must not be None.')
+
         self._current_page = current_page
-        self._index = 0
+        self._iter = iter(self.items)
```
hiranya911, line 48: Should we defer this until `next()` is called?

```py
def __init__(self):
  self._iter = None

def __next__(self):
  if self._iter is None:
    self._iter = iter(self.items)
```

hiranya911 (commented):
Thanks @davidwtbuxton for looking into this. Fix looks pretty good. Just had one suggestion. WDYT?

firebase_admin/_auth_utils.py
davidwtbuxton: I've pushed a change on this patch that includes your suggestion. Thanks for taking time to review the patch and for your advice!

hiranya911 (approved):
LGTM 👍

Issue #578, opened 2021-09-20
[FR] add ability to remove provider via python admin sdk


hardikns:
**Is your feature request related to a problem? Please describe.**
`deleteProvider` field for updateUser has been documented and implemented in other admin SDK like nodejs. Its missing in the python admin sdk

**Describe the solution you'd like**
Add the feature in the python admin sdk as well that will allow users to delete providers (like google.com, facebook.com) when email address changes for the user. Basically add delete_providers as input to update_user method in auth.

hardikns:
@hiranya911 - I have created a PR for the same. #579 pls have a look and see if we can accept it.

PR #579, opened 2021-09-20
feat(auth): ability to delete provider in auth

hardikns:
This address #578, 
Adds the feature in the python admin sdk (which has been present in the nodejs sdk) that will allow users to delete providers (like google.com, facebook.com). Basically add delete_providers as input to update_user method in auth.

### DIFF
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -336,6 +336,7 @@ class Client:
             valid_since: An integer signifying the seconds since the epoch (optional). This field
                 is set by ``revoke_refresh_tokens`` and it is discouraged to set this field
                 directly.
+            delete_providers: The list of provider IDs to unlink, eg: 'google.com', 'password', etc. 
 
         Returns:
             UserRecord: An updated UserRecord instance for the user.
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -336,7 +336,7 @@ class Client:
             valid_since: An integer signifying the seconds since the epoch (optional). This field
                 is set by ``revoke_refresh_tokens`` and it is discouraged to set this field
                 directly.
-            delete_providers: The list of provider IDs to unlink, eg: 'google.com', 'password', etc. 
+            delete_providers: The list of provider IDs to unlink, eg: 'google.com', 'password', etc.
 
         Returns:
             UserRecord: An updated UserRecord instance for the user.
--- a/firebase_admin/_auth_client.py
+++ b/firebase_admin/_auth_client.py
@@ -336,7 +336,8 @@ class Client:
             valid_since: An integer signifying the seconds since the epoch (optional). This field
                 is set by ``revoke_refresh_tokens`` and it is discouraged to set this field
                 directly.
-            delete_providers: The list of provider IDs to unlink, eg: 'google.com', 'password', etc.
+            providers_to_delete: The list of provider IDs to unlink,
+                eg: 'google.com', 'password', etc.
 
         Returns:
             UserRecord: An updated UserRecord instance for the user.
--- a/firebase_admin/_auth_utils.py
+++ b/firebase_admin/_auth_utils.py
@@ -267,9 +267,9 @@ def validate_action_type(action_type):
     return action_type
 
 def validate_provider_ids(provider_ids, required=False):
-    if provider_ids is None:
+    if not provider_ids:
         if required:
-            raise ValueError('Invalid provider IDs. The list must be non-empty.')
+            raise ValueError('Invalid provider IDs. Provider ids should be provided')
         return []
     for provider_id in provider_ids:
         validate_provider_id(provider_id, True)
--- a/firebase_admin/_auth_utils.py
+++ b/firebase_admin/_auth_utils.py
@@ -266,6 +266,15 @@ def validate_action_type(action_type):
             Valid values are {1}'.format(action_type, ', '.join(VALID_EMAIL_ACTION_TYPES)))
     return action_type
 
+def validate_provider_ids(provider_ids, required=False):
+    if provider_ids is None:
+        if required:
+            raise ValueError('Invalid provider IDs. The list must be non-empty.')
+        return []
+    for provider_id in provider_ids:
+        validate_provider_id(provider_id, True)
+    return provider_ids
+
 def build_update_mask(params):
     """Creates an update mask list from the given dictionary."""
     mask = []
--- a/firebase_admin/_user_mgt.py
+++ b/firebase_admin/_user_mgt.py
@@ -688,7 +688,7 @@ class UserManager:
 
     def update_user(self, uid, display_name=None, email=None, phone_number=None,
                     photo_url=None, password=None, disabled=None, email_verified=None,
-                    valid_since=None, custom_claims=None):
+                    valid_since=None, custom_claims=None, delete_providers=None):
         """Updates an existing user account with the specified properties"""
         payload = {
             'localId': _auth_utils.validate_uid(uid, required=True),
@@ -700,6 +700,7 @@ class UserManager:
         }
 
         remove = []
+        removeProvider = []
         if display_name is not None:
             if display_name is DELETE_ATTRIBUTE:
                 remove.append('DISPLAY_NAME')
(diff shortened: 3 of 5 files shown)


hiranya911:
Thank for the PR @hardikns. There's an open PR #383 that implements this feature, but I'm not sure if the original authors are still available to finish up the work.

@rsgowman @bojeil-google @nrsim are any of you available to finish and merge #383? If not we can accept this PR by @hardikns (it will likely need some minor changes to meet the approved API design).

rsgowman:
Ah, sorry; I forgot this was still outstanding. I don't think @nrsim will be available to finish #383, at least not for some time.

@hardikns's PR has the advantage of not having to resolve 1y+ worth of merge conflicts. Though #383 also does linking (whereas this one just deletes).

Since this PR likely only needs minor revision, we can probably get this one submitted quicker than #383. (And I assume @hardikns is waiting on this feature.) My vote is to proceed with this, and then rebase 383 ontop of this one.

My only hesitation is that the `provider_to_link` parameter should ideally appear before the `providers_to_delete` parameter in the `update_user` function. Or at least, I think that's the ordering on the other platforms.

firebase_admin/_user_mgt.py
```diff
@@ -688,7 +688,7 @@ def create_user(self, uid=None, display_name=None, email=None, phone_number=None
 
     def update_user(self, uid, display_name=None, email=None, phone_number=None,
                     photo_url=None, password=None, disabled=None, email_verified=None,
-                    valid_since=None, custom_claims=None):
+                    valid_since=None, custom_claims=None, delete_providers=None):
```
hiranya911, line 691: This should be called `providers_to_delete`.

firebase_admin/_auth_client.py
```diff
@@ -336,6 +336,7 @@ def update_user(self, uid, **kwargs): # pylint: disable=differing-param-doc
             valid_since: An integer signifying the seconds since the epoch (optional). This field
                 is set by ``revoke_refresh_tokens`` and it is discouraged to set this field
                 directly.
+            delete_providers: The list of provider IDs to unlink, eg: 'google.com', 'password', etc.
```
hiranya911, line 339: Should be `providers_to_delete`

firebase_admin/_user_mgt.py
```diff
@@ -726,6 +727,11 @@ def update_user(self, uid, display_name=None, email=None, phone_number=None,
                 custom_claims, dict) else custom_claims
             payload['customAttributes'] = _auth_utils.validate_custom_claims(json_claims)
 
+        if delete_providers is not None and isinstance(delete_providers, list):
```
hiranya911, line 730: I think we want to validate the contents of the list too. #383 added a separate util method for this purpose:

```py
def validate_provider_ids(provider_ids, required=False):
    if provider_ids is None:
        if required:
            raise ValueError('Invalid provider IDs. The list must be non-empty.')
        else:
            return None
    for provider_id in provider_ids:
        validate_provider_id(provider_id, True)
    return provider_ids
```

Can we do something similar here?

integration/test_auth.py
```diff
@@ -496,6 +496,14 @@ def test_disable_user(new_user_with_params):
     assert user.disabled is True
     assert len(user.provider_data) == 1
 
+def test_remove_provider(new_user_with_provider):
+    provider_ids = map(lambda x: x.provider_id, new_user_with_provider.provider_data)
```
hiranya911, line 500: Prefer list comprehension to `map()`:

```py
provider_ids = [provider.provider_id for provider in new_user_with_provider.provider_data]
```

integration/test_auth.py
```diff
@@ -496,6 +496,14 @@ def test_disable_user(new_user_with_params):
...
+    assert 'google.com' in provider_ids
+    user = auth.update_user(new_user_with_provider, delete_providers=['google.com'])
+    assert user.uid == new_user_with_params.uid
+    new_provider_ids = map(lambda x: x.provider_id, user.provider_data)
```
hiranya911, line 504: Here too.

tests/test_user_mgt.py
```diff
@@ -663,6 +663,21 @@ def test_update_user_valid_since(self, user_mgt_app, arg):
...
+        assert set(request['deleteProvider']) == set(arg)
+
+    @pytest.mark.parametrize('arg', [['phone'], ['google.com', 'phone']])
+    def test_update_user_delete_provider_and_phone(self, user_mgt_app, arg):
+        user_mgt, recorder = _instrument_user_manager(user_mgt_app, 200, '{"localId":"testuser"}')
+        user_mgt.update_user('testuser', delete_providers=arg, phone_number=auth.DELETE_ATTRIBUTE)
+        request = json.loads(recorder[0].body.decode())
+        assert 'phone' in request['deleteProvider']
+        assert len(set(request['deleteProvider'])) == len(request['deleteProvider'])
+        assert set(arg) - set(request['deleteProvider']) == set()
```
hiranya911, line 680: Why not just `assert set(arg) == set(request['deleteProvider'])` ?

hiranya911 (changes requested):
Thanks everybody for the comments. I've reviewed the PR with the expectation that we will accept and merge this before #383. 

@hardikns PR looks pretty good. I've pointed out a few things that can be improved.

tests/test_user_mgt.py
hardikns: arg may be [], but since we are removing the phone number its provider gets added.
* `arg == []` , `request['deleteProvider'] == ['phone']`
* `arg == ['phone']` , `request['deleteProvider'] == ['phone']`

firebase_admin/_user_mgt.py
hardikns: done.

firebase_admin/_auth_client.py
hardikns: done

firebase_admin/_user_mgt.py
hardikns: done.

integration/test_auth.py
hardikns: done

integration/test_auth.py
hardikns: done

firebase_admin/_auth_utils.py
```diff
@@ -266,6 +266,15 @@ def validate_action_type(action_type):
             Valid values are {1}'.format(action_type, ', '.join(VALID_EMAIL_ACTION_TYPES)))
     return action_type
 
+def validate_provider_ids(provider_ids, required=False):
+    if provider_ids is None:
+        if required:
+            raise ValueError('Invalid provider IDs. The list must be non-empty.')
```
hiranya911, line 272: This error message is not exactly correct. How about "Provider ID list must not be None".?

tests/test_user_mgt.py
```diff
@@ -663,6 +663,23 @@ def test_update_user_valid_since(self, user_mgt_app, arg):
...
         assert request == {'localId': 'testuser', 'validSince': int(arg)}
 
+    @pytest.mark.parametrize('arg', [['phone'], ['google.com', 'phone']])
+    def test_update_user_delete_provider(self, user_mgt_app, arg):
+        user_mgt, recorder = _instrument_user_manager(user_mgt_app, 200, '{"localId":"testuser"}')
+        user_mgt.update_user('testuser', providers_to_delete=arg)
+        request = json.loads(recorder[0].body.decode())
+        assert set(request['deleteProvider']) == set(arg)
+
+    @pytest.mark.parametrize('arg', [['phone'], ['google.com', 'phone']])
```
hiranya911, line 673: Let's also add `['google.com']` to the parameter list.

tests/test_user_mgt.py
hiranya911: Fair enough. In that case let's also add `[]` to the parameter list.

hiranya911 (approved):
Thanks @hardikns. LGTM with a few nits. I can merge once those comments are addressed.

firebase_admin/_auth_utils.py
hardikns: I am changing the check to this so that it will check for empty list and None both 
```
    if not provider_ids:
        if required:
            raise ValueError('Invalid provider IDs. Provider ids should be provided')
        return []
```

tests/test_user_mgt.py
hardikns: done

tests/test_user_mgt.py
hardikns: done

hardikns:
@hiranya911 I have addressed your comments and also updated my branch. Let me know if anything else is pending.

PR #581, opened 2021-09-30
fix: Extracting GAPIC API calls into a new module

hiranya911:
GAPIC is our largest dependency today (~54MB), and is only used by FCM for batch notification delivery. Currently it's imported by our shared `_utils` module, which puts it in the dependency chain of everything. In this PR I'm extracting the GAPIC API calls into a new `_gapic_utils` module for better isolation. This way, developers who don't need it (and care about the installation footprint), can at least manually delete it with no impact on rest of the SDK.


lahirumaramba (approved):
Thanks Hiranya! LGTM!

PR #582, opened 2021-10-06
fix: Fixing a broken integration test

hiranya911:
Fixing a broken integration test introduced in #579


lahirumaramba (approved):
Thank you, Hiranya!

PR #602, opened 2022-01-07
chore: Update Node.js version in CI workflow

lahirumaramba:
- Firebase CLI dropped support for Node.js 10. Updating the Node.js version to ~12~ 16 for emulator test setup.


hiranya911 (approved):
Thanks. 

Should we go all the way to Node 16 for this?

lahirumaramba:
> Thanks.
> 
> Should we go all the way to Node 16 for this?

Good call. Yes, let's do 16!

Issue #104, opened 2017-12-11
FR: Async requests


eranhirs:
I am using `firebase-admin` on an `asyncio` project and when there are a lot of requests it slows down the whole API.

Surprisingly, I didn't see any other issues about this, I hope I am not duplicating.

```python
ref = yield from db.reference
result = yield from ref.get()
```

Thanks

hiranya911:
Can this be not handled in the application code? Basically execute Admin SDK methods like `ref.get()` on separate threads.

eranhirs:
It won't scale as I would like and, to my opinion, complicates the code and is harder to track.

hiranya911:
Most of our users are on Python 2.7, so we cannot directly expose async coroutines from the SDK anytime soon. In the meantime you should be able to get around the scalability issues by using the [`run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.run_in_executor) method of the asyncio event loop.

```
ref = db.reference(path) # Doesn't make RPC calls

executor = concurrent.futures.ThreadPoolExecutor(max_workers=8)
loop = asyncio.get_event_loop()
loop.run_in_executor(executor, ref.get) # Make RPC call on a separate thread
```

eranhirs:
Thanks! I will consider using it.

Meanwhile, I believe this should be reopened and prioritized.
[There are more users](https://stackoverflow.com/questions/46378139/aiohttp-and-firebase-and-firebase-admin-is-it-non-blocking-python-3-6) who need this.
It is a standard in python 3. It's not like python 2.7 that there are a lot of different libraries to support.

How about supporting both by exposing different methods?

eranhirs:
In case anyone runs into the same problem -
Eventually, I am using for now the [REST API](https://firebase.google.com/docs/database/rest/retrieve-data).
Calling it asynchronously with `aiohttp`.

Natim:
@eranhirs would you feel like starting a async-firebase project that we could contribute to?

hyperknot:
I wanted to report that having async support for an auth SDK is very important. This is exactly the area where async frameworks make a huge difference, where you do a client side request inside a server side request.

Please consider adding core support and examples for using with ASGI frameworks, like Starlette:
https://github.com/encode/starlette

Lawouach:
I agree with the general feeling here. Not having async support is making this library a tad less useful nowadays. Not very future proof unfortunately.

[...]

PR #623, opened 2022-07-13
Adds integration tests for the Async Firstore module

jkyle109:
Adds integration tests for the async Firstore module.

Related to https://github.com/firebase/firebase-admin-python/issues/104

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2022 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


integration/conftest.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2022 Google Inc.
```
lahirumaramba, line 1: We only set this for new files (this is the creation date)

lahirumaramba (approved):
LGTM!

lahirumaramba:
Added `staging` label to trigger integration tests

integration/conftest.py
jkyle109: Got it, thought this was last time edited. I'll keep that in mind!

Issue #624, opened 2022-07-14
auth.verify_id_token fails with 'Token used too early'


fschaeck:
The function auth.verify_id_token may fail for tokens that were issued by servers which have clocks running a little early.

If verification is done right after the token was issued by such a server, then the call to function google.oauth2.id_token.verify_token in _JWTVerifier.verify in firebase_admin/_token_gen.py may be early enough for the 'issued-at-time' timestamp of the token still being in the future.

That will lead to the error 'Token used too early' from google.oauth2.id_token.verify_token, because the function _JWTVerifier.verify is not passing along the optional parameter clock_skew_in_seconds, which would make the google...verify_token function allow for slightly off clock settings.

The worst part about this is, that depending on the server's clock setting and the speed, googles verify_token function is called and the time it takes, to get the api request through to the verification server, the verification of such tokens sometimes may work and sometimes may not. Making this a very hard to understand - and account for - issue.

Adding the optional parameter clock_skew_in_seconds=60 to the call to google.oauth2.id_token.verify_token would allow for the servers clock to be off by up to a minute and still allow verification of the issued token immediately after it being issued. 

Otherwise, developers working with firebase_admin's function auth.verify_id_token would have to add a slight delay before calling the function to account for such clock skew.

siddie:
Any news? It's almost a year since the report and PR.

foolishsailor:
crazy that this hasnt been merged yet?????

omalave:
Try to set google ntp server as your datetime source.
https://developers.google.com/time/guides

GaryFrewin:
Is this ever going to be updated. It seems like an absolutely fundamental part of this package. 

Firebase issues a token and then we try to use it immediately to get some information that the user needs on the front end, and they have to deal with errors because of a clock issue? Seriously? This request is a year old and is a basic and completely required piece of functionality.  

What is going on?

ccppoo:
I added

```py
await asyncio.sleep(1)
```

GaryFrewin:
> I added
> 
> ```python
> await asyncio.sleep(1)
> ```

This doesn't always work. I've tried various sleep times and anything up to 20 seconds can still fail. 

I ended up building my own skew function to catch the error and parse out the times.

ccppoo:
> > I added
> > ```python
> > await asyncio.sleep(1)
> > ```
> 
> This doesn't always work. I've tried various sleep times and anything up to 20 seconds can still fail.
> 
[...]

PR #625, opened 2022-07-14
Added option clockSkewInSeconds to allow setting clock_skew_in_seconds parameter for token verification

fschaeck:
Adding the optional parameter clock_skew_in_seconds=60 to the call to google.oauth2.id_token.verify_token now allows for the token-issuing server's clock to be off by up to a minute without the token becoming invalid due to a 'issued-at-time' timestamp that is in the future.

### Discussion

Issue #624 

### Testing

No additional tests required.

### API Changes

n/a

### DIFF
--- a/firebase_admin/__init__.py
+++ b/firebase_admin/__init__.py
@@ -28,7 +28,7 @@ _clock = datetime.datetime.utcnow
 
 _DEFAULT_APP_NAME = '[DEFAULT]'
 _FIREBASE_CONFIG_ENV_VAR = 'FIREBASE_CONFIG'
-_CONFIG_VALID_KEYS = ['databaseAuthVariableOverride', 'databaseURL', 'httpTimeout', 'projectId',
+_CONFIG_VALID_KEYS = ['clockSkewInSeconds', 'databaseAuthVariableOverride', 'databaseURL', 'httpTimeout', 'projectId',
                       'storageBucket']
 
 def initialize_app(credential=None, options=None, name=_DEFAULT_APP_NAME):
@@ -48,9 +48,10 @@ def initialize_app(credential=None, options=None, name=_DEFAULT_APP_NAME):
       credential: A credential object used to initialize the SDK (optional). If none is provided,
           Google Application Default Credentials are used.
       options: A dictionary of configuration options (optional). Supported options include
-          ``databaseURL``, ``storageBucket``, ``projectId``, ``databaseAuthVariableOverride``,
-          ``serviceAccountId`` and ``httpTimeout``. If ``httpTimeout`` is not set, the SDK
-          uses a default timeout of 120 seconds.
+          ``clockSkewInSeconds``, ``databaseURL``, ``storageBucket``, ``projectId``,
+          ``databaseAuthVariableOverride``, ``serviceAccountId`` and ``httpTimeout``.
+          If ``httpTimeout`` is not set, the SDK uses a default timeout of 120 seconds.
+          If ``clockSkewInSeconds`` is not set, 0 is used when verifying a token.
       name: Name of the app (optional).
     Returns:
       App: A newly initialized instance of App.
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -393,7 +393,8 @@ class _JWTVerifier:
                     token,
                     request=request,
                     audience=self.project_id,
-                    certs_url=self.cert_url)
+                    certs_url=self.cert_url,
+                    clock_skew_in_seconds=60)
             verified_claims['uid'] = verified_claims['sub']
             return verified_claims
         except google.auth.exceptions.TransportError as error:
--- a/firebase_admin/_token_gen.py
+++ b/firebase_admin/_token_gen.py
@@ -55,6 +55,7 @@ METADATA_SERVICE_URL = ('http://metadata.google.internal/computeMetadata/v1/inst
                         'service-accounts/default/email')
 ALGORITHM_RS256 = 'RS256'
 ALGORITHM_NONE = 'none'
+DEFAULT_CLOCK_SKEW_IN_SECONDS = 0
 
 # Emulator fake account
 AUTH_EMULATOR_EMAIL = 'firebase-auth-emulator@example.com'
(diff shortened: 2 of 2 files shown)


DanielJerrehian:
Hi, I wanted to follow up on this issue and ask when it will be merged?

fschaeck:
I changed the pull request a little. The parameter clock_skew_in_seconds for the Google token verification API call is no longer hard-coded with 60 seconds but can be set as an App option. If that option is not set, a default of 0 is used.

This makes all uses of firebase_admin work the same way as before. So the change won't introduce a change of behaviour for those existing applications. But users that need the clock skew setting can add an option to their App and overwrite the default of 0 that way.

I think that is the better way to introduce using the clock_skew_in_seconds parameter of the Google API.

timur737 (approved):
lets merge this changes
many users in stuck 
because of this bug

DanielJerrehian (approved):
A default clock skew is set to 0 seconds and the server's clock is allowed to be off by up to 60 seconds

DanielJerrehian (approved):

Nakachi-S:
Why is this PR not merge?
I am facing this problem too.

lahirumaramba:
Thank you @fschaeck for your contribution! The changes to App options requires an internal API review and this might take a while.

Hi, @prameshj, @renkelvin let me know your thoughts on this, I can initiate the internal changes. Thank you!

lahirumaramba (commented):
Thank you @fschaeck for your contribution! This looks good to me, but as an alternative, what do you think about adding `clock_skew_in_seconds =0` to `verify_id_token()` and `verify_session_cookie` as an optional parameter?
 https://github.com/firebase/firebase-admin-python/blob/b9e95e8248eb1473ca5a13bf64e8a33b79dc9db3/firebase_admin/auth.py#L194

Yudai-Saito:
Hello. 
I hope this PR will be merged because I had the same problem with the same bug.

By the way, we have a similar bug with this method.
```python
auth.verify_session_cookie(session, check_revoked=True)
```

I solved it this way.
```python
session_cookie_request = requests.Request()
CERTS_URL = "https://www.googleapis.com/identitytoolkit/v3/relyingparty/publicKeys"
id_token.verify_token(session_token=session, request=session_cookie_request, certs_url=CERTS_URL, clock_skew_in_seconds=10)
```

How about modifying this one this way as well?
https://github.com/firebase/firebase-admin-python/blob/b9e95e8248eb1473ca5a13bf64e8a33b79dc9db3/firebase_admin/_token_gen.py#L392-L396

proposed amendment
```python
 verified_claims = google.oauth2.id_token.verify_token(
     token,
     request=request,
     audience=self.project_id,
     certs_url=self.cert_url,
     clock_skew_in_seconds=self._clock_skew_in_seconds
)
```

Yudai-Saito (approved):
LGTM!

ryanlinnane:
@fschaeck are you able to merge this now? Subscribing.

smartexpert:
appreciate if this could be merged

lahirumaramba:
Thank you for your patience everyone. We just completed the internal API review and we think a better approach would be to add a new optional `clock_skew_seconds` parameter to `verify_id_token()` and `verify_session_cookie()` APIs.

**API:**
```
def verify_id_token(id_token, app=None, check_revoked=False, clock_skew_seconds=0):

def verify_session_cookie(session_cookie, check_revoked=False, app=None, clock_skew_seconds=0):
```

**Usage:**
```
import firebase_admin
from firebase_admin import auth

app = firebase_admin.initialize_app()

decoded_token = auth.verify_id_token(id_token, clock_skew_in_seconds=60)
decoded_claims = auth.verify_session_cookie(session_cookie, check_revoked=True, clock_skew_in_seconds=10)
```

@fschaeck would you be able to make the changes? Thank you!

barel-mishal:
Hi mean while this need to be changed what the work around? could not figure out yat how to skew my clock with node js

ryanlinnane:
> Hi mean while this need to be changed what the work around? could not figure out yet how to skew my clock with node js

Fork the fix and install the git repo with pip works for now.

fschaeck:
> @fschaeck would you be able to make the changes? Thank you!

@lahirumaramba 

You do realize, that this means to change the signatures of methods TokenVerifier.verfify_id_token, TokenVerifier.verify_session_cookie and _JWTVerifier.verify from firebase_admin/_token_gen.py by adding the optional parameter clock_skew_seconds there as well, right?

Since TokenVerifier and _JWTVerifier are classes (doubly) abstracting the token verification and I was not able to assess, if there are any other implementations that might be used here, I chose the way of the additional app property instead of changing the signature of those methods, because other such abstracting classes may not understand the additional parameter and then apps using this optional parameter all the sudden stop working.

If you okay THOSE change as well, I will change this pull request right the way!

beef-stek:
> ```python
> verify_token
> ```

Can you please go into more detail on your fix? What class is id_token? Thanks.

This bug has been annoying as it would happen randomly. Please merge this as soon as possible! THANKS GOOGLE.

fschaeck:
@lahirumaramba  Any news here?

lahirumaramba:
> 
> You do realize, that this means to change the signatures of methods TokenVerifier.verfify_id_token, TokenVerifier.verify_session_cookie and _JWTVerifier.verify from firebase_admin/_token_gen.py by adding the optional parameter clock_skew_seconds there as well, right?
> 
> Since TokenVerifier and _JWTVerifier are classes (doubly) abstracting the token verification and I was not able to assess, if there are any other implementations that might be used here, I chose the way of the additional app property instead of changing the signature of those methods, because other such abstracting classes may not understand the additional parameter and then apps using this optional parameter all the sudden stop working.
> 
> If you okay THOSE change as well, I will change this pull request right the way!

Hi @fschaeck thank you for being so thoughtful of this change. Since `_JWTVerifier` and `TokenVerifier` are internal classes and not exposed in the public API surface, we are safe to change the method signature by adding the optional parameter `clock_skew_seconds`. Please add unit tests (and integration tests) where possible. Thanks!

karkir0003:
@fschaeck when will this pr get reviewed/approved to be merged? this change would be so nice to have for firebase

theobouwman:
Any updates on this? Or how to avoid it?

jonathanedey:
Addressed in #714

PR #629, opened 2022-07-26
feat(credentials): Add async credentials.

jkyle109:
Added async credential instances alongside existing sync credentials.
Added type hints.

Related to https://github.com/firebase/firebase-admin-python/issues/104

### DIFF
--- a/firebase_admin/credentials.py
+++ b/firebase_admin/credentials.py
@@ -17,7 +17,10 @@ import collections
 import json
 import pathlib
 
-from typing import Type
+from typing import (
+    Type,
+    List
+)
 
 import google.auth  # type: ignore
 from google.auth import default
@@ -32,7 +35,7 @@ from google.oauth2 import _service_account_async as service_account_async
 
 _request: requests.Request = requests.Request()
 _request_async: aiohttp_requests.Request = aiohttp_requests.Request()
-_scopes: list[str] = [
+_scopes: List[str] = [
     'https://www.googleapis.com/auth/cloud-platform',
     'https://www.googleapis.com/auth/datastore',
     'https://www.googleapis.com/auth/devstorage.read_write',
--- a/firebase_admin/credentials.py
+++ b/firebase_admin/credentials.py
@@ -17,14 +17,22 @@ import collections
 import json
 import pathlib
 
-import google.auth
-from google.auth.transport import requests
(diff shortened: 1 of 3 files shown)


firebase_admin/credentials.py
```diff
@@ -95,44 +120,54 @@ def __init__(self, cert):
...
 class ApplicationDefault(Base):
     """A Google Application Default credential."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         """Creates an instance that will use Application Default credentials.
 
-        The credentials will be lazily initialized when get_credential() or
-        project_id() is called. See those methods for possible errors raised.
+        The credentials will be lazily initialized when get_credential(), get_credential_async()
```
egilmorez, line 162: Suggest code font for these literal references to functions, like "..when <code>get_credential()</code>,..."

egilmorez (approved):
LG with one formatting nit, thanks!

firebase_admin/credentials.py
lahirumaramba: @jkyle109 I think you can try double ` here. See https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/messaging.py#L130

jkyle109: Sounds good, Done!

requirements.txt
```diff
@@ -4,6 +4,7 @@ pytest >= 6.2.0
 pytest-cov >= 2.4.0
 pytest-localserver >= 0.4.1
 
+aiohttp == 3.8.1
```
lahirumaramba, line 7: Is there a reason to pin the version to `3.8.1` or is it safer to to `>=` here as well?

lahirumaramba (approved):
LGTM, Thank you!

lahirumaramba:
@jkyle109 I added the release:stage label to trigger integration tests just to be safe. Could you do an empty commit to trigger the CIs again, please? Thank you!
```
git commit --allow-empty -m "Trigger CIs"
```

requirements.txt
jkyle109: My reasoning for it originally was that it was since the latest release and the only one which is compatible. But I think it's still safe to have it be ``>=``.

jkyle109:
The failing integration test is also failing in master. Is this free to merge @lahirumaramba ?

PR #642, opened 2022-09-07
feat: Add function to verify an App Check token

dwyfrequency:
Add Firebase functionality to verify an App Check token

### DIFF
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -43,6 +43,7 @@ class _AppCheckService:
     _APP_CHECK_ISSUER = 'https://firebaseappcheck.googleapis.com/'
     _JWKS_URL = 'https://firebaseappcheck.googleapis.com/v1/jwks'
     _project_id = None
+    _scoped_project_id = None
 
     def __init__(self, app):
         # Validate and store the project_id to validate the JWT claims
@@ -52,6 +53,8 @@ class _AppCheckService:
                 'Project ID is required to access App Check service. Either set the '
                 'projectId option, or use service account credentials. Alternatively, set the '
                 'GOOGLE_CLOUD_PROJECT environment variable.')
+        self._scoped_project_id = 'projects/' + app.project_id 
+
 
     def verify_token(self, token: str) -> Dict[str, Any]:
         """Verifies a Firebase App Check token."""
@@ -64,7 +67,7 @@ class _AppCheckService:
         jwks_client = PyJWKClient(self._JWKS_URL, lifespan=21600)
         signing_key = jwks_client.get_signing_key_from_jwt(token)
         self._has_valid_token_headers(jwt.get_unverified_header(token))
-        verified_claims = self._decode_and_verify(token, signing_key.get('key'))
+        verified_claims = self._decode_and_verify(token, signing_key.key)
 
         # The token's subject will be the app ID, you may optionally filter against
         # an allow list
@@ -91,7 +94,8 @@ class _AppCheckService:
             payload = jwt.decode(
                 token,
                 signing_key,
-                algorithms
+                algorithms,
+                audience=self._scoped_project_id
             )
         except (DecodeError, InvalidKeyError):
             ValueError(
@@ -108,9 +112,9 @@ class _AppCheckService:
             algorithms=["RS256"]
         )
 
-        scoped_project_id = 'projects/' + self._project_id
+        
         audience = payload.get('aud')
-        if not isinstance(audience, list) and scoped_project_id not in audience:
+        if not isinstance(audience, list) and self._scoped_project_id not in audience:
             raise ValueError('Firebase App Check token has incorrect "aud" (audience) claim.')
         if not payload.get('iss').startswith(self._APP_CHECK_ISSUER):
             raise ValueError('Token does not contain the correct "iss" (issuer).')
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -60,6 +60,7 @@ class _AppCheckService:
         # Obtain the Firebase App Check Public Keys
         # Note: It is not recommended to hard code these keys as they rotate,
         # but you should cache them for up to 6 hours.
+        # TODO(dwyfrequency): update cache lifespan
         jwks_client = PyJWKClient(self._JWKS_URL)
         signing_key = jwks_client.get_signing_key_from_jwt(token)
         self._has_valid_token_headers(jwt.get_unverified_header(token))
@@ -70,17 +71,17 @@ class _AppCheckService:
         verified_claims['app_id'] = verified_claims.get('sub')
         return verified_claims
 
-    def _has_valid_token_headers(self, header: Any) -> None:
+    def _has_valid_token_headers(self, headers: Any) -> None:
         """Checks whether the token has valid headers for App Check."""
         # Ensure the token's header has type JWT
-        if header.get('typ') != 'JWT':
+        if headers.get('typ') != 'JWT':
             raise ValueError("The provided App Check token has an incorrect type header")
         # Ensure the token's header uses the algorithm RS256
-        algorithm = header.get('alg')
+        algorithm = headers.get('alg')
         if algorithm != 'RS256':
             raise ValueError(
-                f'The provided App Check token has an incorrect algorithm. '
-                'Expected RS256 but got {algorithm}.'
+                'The provided App Check token has an incorrect algorithm. '
+                f'Expected RS256 but got {algorithm}.'
                 )
 
     def _decode_token(self, token: str, signing_key: str, algorithms: List[str]) -> Dict[str, Any]:
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -29,35 +29,21 @@ from typing import Any, Dict, List
 from firebase_admin import _utils
 
 _APP_CHECK_ATTRIBUTE = '_app_check'
-_APP_CHECK_API_URL = "https://firebaseappcheck.googleapis.com/"
 
 def _get_app_check_service(app) -> Any:
     return _utils.get_app_service(app, _APP_CHECK_ATTRIBUTE, _AppCheckService)
 
-# we'll need a public method like def send(message, dry_run=False, app=None):
-# that lives outside of the class. This method will refer to the class instance
-# when calling
-
-# outside of the class we will need all the public methods that 
-# in this case just `verify_token`
-#  
-
-# should i accept an app or just always make it none
+# should i accept an app (design doc doesn't have one) or just always make it none
 def verify_token(token: str, app=None) -> Dict[str, Any]:
     return _get_app_check_service(app).verify_token(token)
 
-# this can be deleted, it was only for the example
-def run_checks():
-    app_id = verify_app_check(request.headers.get('X-Firebase-AppCheck'))
-    if app_id is None:
-        abort(401)
-    app.config['APP_ID'] = app_id
-
 class _AppCheckService:
     """Service class that implements Firebase App Check functionality."""
-    # Then we insert the code sample that uses flask but we do not need
-    # the actual flask stuff
-    # import all of verify_app_check from https://github.com/lahirumaramba/codecloud/blob/2f9ea1e206c740ed4c01c3277ee6745a39f8ee21/app-check-verify/python/server.py
+    
+    _APP_CHECK_GCP_API_URL = "https://firebaseappcheck.googleapis.com"
+    _APP_CHECK_BETA_JWKS_RESOURCE = "/v1beta/jwks" 
+
+
     def __init__(self, app):
         # the verification method should go in the service
         project_id = app.project_id
@@ -66,6 +52,7 @@ class _AppCheckService:
                 'Project ID is required to access App Check service. Either set the '
                 'projectId option, or use service account credentials. Alternatively, set the '
                 'GOOGLE_CLOUD_PROJECT environment variable.')
+        # Unsure what I should include in this constructor, or even if I should include one
     
     @classmethod
     def verify_token(self, token: str) -> Dict[str, Any]:
@@ -75,14 +62,13 @@ class _AppCheckService:
         # Obtain the Firebase App Check Public Keys
         # Note: It is not recommended to hard code these keys as they rotate,
         # but you should cache them for up to 6 hours.
-        url = "https://firebaseappcheck.googleapis.com/v1beta/jwks"
+        url = f'{self._APP_CHECK_GCP_API_URL}{self._APP_CHECK_BETA_JWKS_RESOURCE}'
 
         jwks_client = PyJWKClient(url)
         signing_key = jwks_client.get_signing_key_from_jwt(token)
 
         header = jwt.get_unverified_header(token)
-        if not self._has_valid_verify_token_headers(header):
-            return None
+        self._has_valid_token_headers(header) 
 
         
         # I don't see any method or property to just get key from signing_key /*/lib/python3.10/site-packages/jwt/api_jwk.py
@@ -92,14 +78,13 @@ class _AppCheckService:
         # an allow list
         return payload.get('sub')
     
-    def _has_valid_verify_token_headers(header: Any) -> bool:
-        # Ensure the token's header uses the algorithm RS256
-        if header.get('alg') != 'RS256':
-            return False
+    def _has_valid_token_headers(header: Any) -> None:
         # Ensure the token's header has type JWT
         if header.get('typ') != 'JWT':
-            return False
-        return True
+            raise ValueError("The token received is not a JWT")
+        # Ensure the token's header uses the algorithm RS256
+        if header.get('alg') != 'RS256':
+            raise ValueError("JWT's algorithm does not have valid token headers")
 
     def _decode_token(token: str, signing_key: str, algorithms:List[str]=["RS256"]) -> Dict[str, Any]:
         payload = {}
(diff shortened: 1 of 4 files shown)


firebase_admin/app_check.py
```diff
@@ -0,0 +1,123 @@
...
+        if not project_id:
+            raise ValueError(
+                'Project ID is required to access App Check service. Either set the '
+                'projectId option, or use service account credentials. Alternatively, set the '
+                'GOOGLE_CLOUD_PROJECT environment variable.')
+        # Unsure what I should include in this constructor, or even if I should include one
+    
+    @classmethod
+    def verify_token(self, token: str) -> Dict[str, Any]:
+        if token is None:
```
dwyfrequency, line 59: Do I even need this check or would it fail on compile if a str wasn't passed in?

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+    def _decode_and_verify(self, token: str, signing_key: str):
+        """Decodes and verifies the token from App Check."""
+        payload = self._decode_token(
+            token,
+            signing_key,
+            algorithms=["RS256"]
+        )
+
+        # within the aud property, there will be an array of project id & number
+        if len(payload.get('aud')) <= 1:
```
dwyfrequency, line 111: when should I use get and when should I be using dot syntax? Are they interchangeable?

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+            return None
+
+        # Obtain the Firebase App Check Public Keys
+        # Note: It is not recommended to hard code these keys as they rotate,
+        # but you should cache them for up to 6 hours.
+        url = f'{cls._APP_CHECK_GCP_API_URL}{cls._APP_CHECK_BETA_JWKS_RESOURCE}'
+        jwks_client = PyJWKClient(url)
+        signing_key = jwks_client.get_signing_key_from_jwt(token)
+
+        # Getting error "No value for argument 'header' in unbound
```
dwyfrequency, line 71: Not sure why the linter is angry here

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+    @classmethod
+    def verify_token(cls, token: str) -> Dict[str, Any]:
+        """Verifies a Firebase App Check token."""
+        if token is None:
+            return None
+
+        # Obtain the Firebase App Check Public Keys
+        # Note: It is not recommended to hard code these keys as they rotate,
+        # but you should cache them for up to 6 hours.
+        url = f'{cls._APP_CHECK_GCP_API_URL}{cls._APP_CHECK_BETA_JWKS_RESOURCE}'
```
dwyfrequency, line 67: this feels kind of clunky but unsure if it's ok to just make it a pure string literal

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    return _get_app_check_service(app).verify_token(token)
+
+class _AppCheckService:
+    """Service class that implements Firebase App Check functionality."""
+
+    _APP_CHECK_GCP_API_URL = 'https://firebaseappcheck.googleapis.com'
+    _APP_CHECK_BETA_JWKS_RESOURCE = '/v1beta/jwks'
+
+    def __init__(self, app):
```
dwyfrequency, line 48: Unsure what I should include in this constructor, or even if I should include one

dwyfrequency:
I also see the messaging.py has extensive error mapping. Should I be creating this as well?

firebase_admin/app_check.py
lahirumaramba: We need to check the token here and throw if it is not a valid string (the type hints won't validate it). You can check how we validate the string in Node.js as an example https://github.com/firebase/firebase-admin-node/blob/318f0e43d65f47dfae81decf5166ddd198d99680/src/app-check/token-verifier.ts#L53

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+        signing_key = jwks_client.get_signing_key_from_jwt(token)
+
+        # Getting error "No value for argument 'header' in unbound
+        # method call (no-value-for-parameter)"
+        cls._has_valid_token_headers(jwt.get_unverified_header(token))
+        payload = cls._decode_and_verify(token, signing_key.key, "project_number")
+
+        # The token's subject will be the app ID, you may optionally filter against
+        # an allow list
+        return payload.get('sub')
```
lahirumaramba, line 78: We need to return the full decoded token here (a dictionary with all the claims). Additionally we need to add `app_id` to the payload.
```
verified_claims['app_id'] = verified_claims['sub']
return verified_claims
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _has_valid_token_headers(self, header: Any) -> None:
+        """Checks whether the token has valid headers for App Check."""
+        # Ensure the token's header has type JWT
+        if header.get('typ') != 'JWT':
+            raise ValueError("The token received is not a JWT")
+        # Ensure the token's header uses the algorithm RS256
+        if header.get('alg') != 'RS256':
+            raise ValueError("JWT's algorithm does not have valid token headers")
```
lahirumaramba, line 87: `The provided App Check token has incorrect algorithm. Expected RS256 but got ...`

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # The token's subject will be the app ID, you may optionally filter against
+        # an allow list
+        return payload.get('sub')
+
+    def _has_valid_token_headers(self, header: Any) -> None:
+        """Checks whether the token has valid headers for App Check."""
+        # Ensure the token's header has type JWT
+        if header.get('typ') != 'JWT':
+            raise ValueError("The token received is not a JWT")
```
lahirumaramba, line 84: `The provided App Check token has incorrect type header`

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+        """Decodes the JWT received from App Check."""
+        payload = {}
+        try:
+            payload = jwt.decode(
+                token,
+                signing_key,
+                algorithms
+            )
+        except DecodeError:
+            ValueError('Unable to decode the token')
```
lahirumaramba, line 99: `Decoding App Check token failed. Make sure you passed the entire string JWT which represents the Firebase App Check token.`

firebase_admin/app_check.py
lahirumaramba: It is a safer way of accessing a dictionary. `get()` will return `None` if the key does not exist and `.key` will crash with an error.

lahirumaramba, line 111: In some JWTs `aud` can be just a string... in that case `len(payload.get('aud'))` will throw an error. I think it is better to first check the type of `payload.get('aud')`, and then check to see if it contains the value we are looking for.

```
scopedProjectId = 'projects/' + self.project_id

if not isinstance(payload.get('aud'), list) and scopedProjectId not in payload.get('aud'):
   raise 'Firebase App Check token has incorrect "aud" (audience) claim.'
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+        payload = self._decode_token(
+            token,
+            signing_key,
+            algorithms=["RS256"]
+        )
+
+        # within the aud property, there will be an array of project id & number
+        if len(payload.get('aud')) <= 1:
+            raise ValueError('Project ID and Project Number are required to access App Check.')
+        if self._APP_CHECK_GCP_API_URL not in payload.get('issuer'):
```
lahirumaramba, line 113: Since we don't have access to the project number we are unable to compare the complete issuer claim here. We can instead use `startswith` to see if the value starts with the Issuer URL.

```
if not payload.get('issuer').startswith(self._APP_CHECK_GCP_API_URL:
```

lahirumaramba (changes requested):
Thank you @dwyfrequency !
We are headed in the correct direction. I added a few comments. Thanks!

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+from jwt import PyJWKClient, DecodeError
+from firebase_admin import _utils
+
+_APP_CHECK_ATTRIBUTE = '_app_check'
+
+def _get_app_check_service(app) -> Any:
+    return _utils.get_app_service(app, _APP_CHECK_ATTRIBUTE, _AppCheckService)
+
+# should i accept an app (design doc doesn't have one) or just always make it none
+def verify_token(token: str, app=None) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
dwyfrequency: Is optional ok for app or should it be not have a default value?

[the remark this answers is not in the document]
lahirumaramba: It's okay to have a default value there. `app=None` looks good!

firebase_admin/app_check.py
dwyfrequency: Do you also want me to add `Expected JWT but got ...` for this one?

firebase_admin/app_check.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @classmethod
+    def verify_token(cls, token: str) -> Dict[str, Any]:
+        """Verifies a Firebase App Check token."""
+        if token is None:
+            return None
+
+        # Obtain the Firebase App Check Public Keys
+        # Note: It is not recommended to hard code these keys as they rotate,
+        # but you should cache them for up to 6 hours.
```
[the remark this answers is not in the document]
dwyfrequency: It looks like it is cached as refresh is by default set to False https://github.com/jpadilla/pyjwt/blob/fc5b94eb3575254caba599218246616c75fecdc7/jwt/jwks_client.py#L96

firebase_admin/app_check.py
dwyfrequency: Based on https://github.com/firebase/firebase-admin-node/blob/318f0e43d65f47dfae81decf5166ddd198d99680/src/app-check/token-verifier.ts#L62, should we instead do 
```
        payload = self._decode_and_verify(token, signing_key.get('key'))
        # The token's subject will be the app ID, you may optionally filter against
        # an allow list
        payload['app_id'] = payload.get('sub')
        return payload
```

At least that's what it looks like the node code is doing.

dwyfrequency: else I'm happy to just update the naming of payload to verified_claims

firebase_admin/app_check.py
dwyfrequency: Added a `_Validators` class

firebase_admin/app_check.py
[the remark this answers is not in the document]
lahirumaramba: Looks like the default is 5 mins. Can we set it to 6 hours with `JWKSetCache`? 
https://github.com/jpadilla/pyjwt/blob/fc5b94eb3575254caba599218246616c75fecdc7/jwt/jwks_client.py#L23

lahirumaramba (commented):
Looks great! Thanks @dwyfrequency!
I think we are good to work on the tests now.

Should we also add a new error type instead of raising value errors?

firebase_admin/app_check.py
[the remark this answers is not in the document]
dwyfrequency: Added comment asking when they're release this to prod https://github.com/jpadilla/pyjwt/pull/781#issuecomment-1249681497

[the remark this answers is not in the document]
dwyfrequency: update requirements to use latest release

firebase_admin/app_check.py
```diff
@@ -0,0 +1,132 @@
...
+def verify_token(token: str, app=None) -> Dict[str, Any]:
+    """Verifies a Firebase App Check token.
+
+    Args:
+        token: A token from App Check.
+        app: An App instance (optional).
+
+    Returns:
+        Dict[str, Any]: A token's decoded claims
+        if the App Check token is valid; otherwise, a rejected promise.
```
dwyfrequency, line 36: Now, it does not return a rejected promise. What should I be returning in case of failure.

firebase_admin/app_check.py
```diff
@@ -0,0 +1,132 @@
...
+    def _decode_token(self, token: str, signing_key: str, algorithms: List[str]) -> Dict[str, Any]:
+        """Decodes the JWT received from App Check."""
+        payload = {}
+        try:
+            payload = jwt.decode(
+                token,
+                signing_key,
+                algorithms
+            )
+        except (DecodeError, InvalidKeyError):
```
dwyfrequency, line 96: having difficulty testing these errors

firebase_admin/app_check.py
lahirumaramba: I think the rejected promise part is a copypasta from Node.js. In Python we should raise an error for any invalid tokens.

example in it token verification: https://github.com/firebase/firebase-admin-python/blob/9e5b8e383e066c319a483285df903150d5029a34/firebase_admin/auth.py#L209

firebase_admin/app_check.py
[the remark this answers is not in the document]
lahirumaramba: That's great! Thank you!

firebase_admin/app_check.py
lahirumaramba: Yeah.. I think to test these we will have to mock `jwt.decode()` (or wrap it in a helper class and throw from there). Let's continue with the rest and come back to this later.

lahirumaramba (commented):
LGTM! Thanks!

firebase_admin/app_check.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def verify_token(self, token: str) -> Dict[str, Any]:
+        """Verifies a Firebase App Check token."""
+        _Validators.check_string("app check token", token)
+
+        # Obtain the Firebase App Check Public Keys
+        # Note: It is not recommended to hard code these keys as they rotate,
+        # but you should cache them for up to 6 hours.
+        # Default lifespan is 300 seconds (5 minutes) so we change it to 21600 seconds (6 hours).
+        jwks_client = PyJWKClient(self._JWKS_URL, lifespan=21600)
```
lahirumaramba, line 70: Sorry for not catching earlier, I think we should initialize this in the constructor and make jwks_client a class member. Otherwise, every time we call `verify_token()` it will initialize a new instance and I don't think that is necessary plus it will also break the caching mechanism.

firebase_admin/app_check.py
```diff
@@ -0,0 +1,146 @@
...
+        signing_key = jwks_client.get_signing_key_from_jwt(token)
+        self._has_valid_token_headers(jwt.get_unverified_header(token))
+        verified_claims = self._decode_and_verify(token, signing_key.key)
+
+        # The token's subject will be the app ID, you may optionally filter against
+        # an allow list
```
lahirumaramba, line 76: I think we can remove `you may optionally filter against an allow list` part. It was a comment in the code sample to help developers, but we don't do the filtering in the SDK.

firebase_admin/app_check.py
```diff
@@ -0,0 +1,146 @@
...
+                f'Expected payload to include {self._scoped_project_id}.'
+                )
+        except InvalidIssuerError:
+            raise ValueError(
+                'The provided App Check token has incorrect "iss" (issuer) claim.'
+                f'Expected claim to include {self._APP_CHECK_ISSUER}'
+                )
+        except ExpiredSignatureError:
+            raise ValueError(
+                'The provided App Check token signature has expired.'
```
lahirumaramba, line 119:
```suggestion
                'The provided App Check token has expired.'
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            payload = jwt.decode(
+                token,
+                signing_key,
+                algorithms=["RS256"],
+                audience=self._scoped_project_id
+            )
+        except InvalidSignatureError:
+            raise ValueError(
+                'The provided App Check token signature cannot be verified.'
```
lahirumaramba, line 105:
```suggestion
                'The provided App Check token has invalid signature.'
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,146 @@
...
+        except InvalidTokenError as exception:
+            raise ValueError(
+                f'Decoding App Check token failed. Error: {exception}'
+                )
+
+        audience = payload.get('aud')
+        if not isinstance(audience, list) or self._scoped_project_id not in audience:
+            raise ValueError('Firebase App Check token has incorrect "aud" (audience) claim.')
+        if not payload.get('iss').startswith(self._APP_CHECK_ISSUER):
+            raise ValueError('Token does not contain the correct "iss" (issuer).')
```
lahirumaramba, line 130: Can we also validate the `sub` claim?

example from Node.js:
```
} else if (typeof payload.sub !== 'string') {
      errorMessage = 'The provided App Check token has no "sub" (subject) claim.';
    } else if (payload.sub === '') {
      errorMessage = 'The provided App Check token has an empty string "sub" (subject) claim.';
    }
    ```

**Review by lahirumaramba — APPROVED** [review]
LGTM! Thank you!

File: firebase_admin/app_check.py [hunk 20]
```diff
@@ -0,0 +1,148 @@
...
+
+def verify_token(token: str, app=None) -> Dict[str, Any]:
+    """Verifies a Firebase App Check token.
+
+    Args:
+        token: A token from App Check.
+        app: An App instance (optional).
+
+    Returns:
+        Dict[str, Any]: A token's decoded claims if the App Check token
```
kevinthecheung, line 36:
```suggestion
        Dict[str, Any]: The token's decoded claims.
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Raises:
+        ValueError: If ``project_id``, headers, or the decoded token payload is invalid.
```
kevinthecheung, line 39:
```suggestion
        ValueError: If the app's ``project_id`` is invalid or unspecified, or if the token's headers or payload are invalid.
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,148 @@
...
+    _jwks_client = None
+
+    def __init__(self, app):
+        # Validate and store the project_id to validate the JWT claims
+        self._project_id = app.project_id
+        if not self._project_id:
+            raise ValueError(
+                'Project ID is required to access App Check service. Either set the '
+                'projectId option, or use service account credentials. Alternatively, set the '
+                'GOOGLE_CLOUD_PROJECT environment variable.')
```
kevinthecheung, line 59:
```suggestion
            raise ValueError(
                'A project ID must be specified to access the App Check '
                'service. Either set the projectId option, use service '
                'account credentials, or, set the '
                'GOOGLE_CLOUD_PROJECT environment variable.')
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,148 @@
...
+    def _has_valid_token_headers(self, headers: Any) -> None:
+        """Checks whether the token has valid headers for App Check."""
+        # Ensure the token's header has type JWT
+        if headers.get('typ') != 'JWT':
+            raise ValueError("The provided App Check token has an incorrect type header")
+        # Ensure the token's header uses the algorithm RS256
+        algorithm = headers.get('alg')
+        if algorithm != 'RS256':
+            raise ValueError(
+                'The provided App Check token has an incorrect algorithm. '
```
kevinthecheung, line 88:
```suggestion
                'The provided App Check token has an incorrect alg header. '
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,148 @@
...
+        try:
+            payload = jwt.decode(
+                token,
+                signing_key,
+                algorithms=["RS256"],
+                audience=self._scoped_project_id
+            )
+        except InvalidSignatureError:
+            raise ValueError(
+                'The provided App Check token has invalid signature..'
```
kevinthecheung, line 104:
```suggestion
                'The provided App Check token has invalid signature.'
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,148 @@
...
+                )
+        except InvalidAudienceError:
+            raise ValueError(
+                'The provided App Check token has incorrect "aud" (audience) claim.'
```
kevinthecheung, line 108:
```suggestion
                'The provided App Check token has incorrect "aud" (audience) claim. '
```

firebase_admin/app_check.py
```diff
@@ -0,0 +1,148 @@
...
+                f'Expected payload to include {self._scoped_project_id}.'
+                )
+        except InvalidIssuerError:
+            raise ValueError(
+                'The provided App Check token has incorrect "iss" (issuer) claim.'
```
kevinthecheung, line 113:
```suggestion
                'The provided App Check token has incorrect "iss" (issuer) claim. '
```

kevinthecheung (approved):

kevinthecheung (approved):

egilmorez (approved):
With Kevin's suggested changes, LGTM, thanks!

PR #645, opened 2022-09-22
change: Drop Python 3.6 support

lahirumaramba:
Drop support for Python 3.6 (Python 3.6 has reached EoL)

RELEASE NOTE: Dropped support for Python 3.6. Developers should to use Python 3.7 or higher when deploying the Admin SDK.


dwyfrequency (approved):
LGTM

Issue #666, opened 2023-01-04
[FR] Improve exception handling for rate limiting


skion:
**Is your feature request related to a problem? Please describe.**

When using the SDK to create email action links, such as for password reset or email verifications, Firebase is strongly rate limiting for good reason. However, it is currently impossible to (programmatically) detect whether the reason for failure was indeed rate limiting, or something else, since in both cases a generic `InvalidArgumentError` exception is raised.

**Describe the solution you'd like**

A way to distinguish rate limiting from other client or server side errors, so that my API is able to return 429 or 503 responses with a `Retry-After` header instead of a generic 500 response in those cases.

**Describe alternatives you've considered**

In order to prevent my API from returning generic 500 I'm considering to implement my own rate limiting in front now. However, that is less elegant since I would need to heuristically match Firebase's rate limiting algorithm in order to prevent emitting 500's anyway.

**Additional context**

Currently the `RESET_PASSWORD_EXCEED_LIMIT` error code from Firebase is swallowed and converted into a `INVALID_ARGUMENT` code, because this type of exception is not special-cased:

    InvalidArgumentError('Error while calling Auth service (RESET_PASSWORD_EXCEED_LIMIT).')

Note that the same holds for building email verification links:

    InvalidArgumentError('Error while calling Auth service (TOO_MANY_ATTEMPTS_TRY_LATER).')

In both cases the original code is not introspectable in the exception, except by parsing the error string.

mohitCodepy:
@lahirumaramba , Is anyone actively working on this issue ? If not then please assign me, thanks

skion:
@lahirumaramba @pragatimodi FWIW I linked a PR a while ago that I believe solves this problem. Can someone have a look at it?

PR #695, opened 2023-04-13
Add rate limiting exceptions

skion:
This adds two exception types to enable users to detect rate limited calls which were previously swallowed and converted into a generic `InvalidArgumentError`, making it impossible to detect rate limiting situations and pass that information through to the end user.

This is being discussed in #666.

I chose to inherit from the `ResourceExhaustedError` base exception, which seemed to make most sense.

### DIFF
--- a/firebase_admin/_auth_utils.py
+++ b/firebase_admin/_auth_utils.py
@@ -405,6 +405,20 @@ class UserDisabledError(exceptions.InvalidArgumentError):
         exceptions.InvalidArgumentError.__init__(self, message, cause, http_response)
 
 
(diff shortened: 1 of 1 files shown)


skion:
@pragatimodi Since you are assigned to #666, would you be able to check this out?

skion:
@jonathanedey I see you are mostly maintaining this repo now, is this something you could have a look at?

lahirumaramba (approved):

pragatimodi (approved):

jonathanedey (approved):
Hi @skion, This looks great. Thank you so much for your patience in getting this added!

PR #706, opened 2023-06-07
feat(fcm): Add `send_each` and `send_each_for_multicast` for FCM batch send

Doris-Ge:
* Deprecate send_all() and send_multicast() APIs.
* Add new send_each() and send_each_for_multicast() APIs.

`send_each` vs `send_all`

1. `send_each` sends one HTTP request to V1 Send endpoint for each message in the list. `send_all` sends only one HTTP request to V1 Batch Send endpoint to send all messages in the array.
2. `send_each` uses concurrent.futures.ThreadPoolExecutor to run and wait for all `request` calls to complete and construct a `BatchResponse`. An `request` call to V1 Send endpoint either completes with a success or throws an exception. So if an exception is thrown out, the exception will be caught in `send_each` and turned into a `SendResponse` with an exception. Therefore, unlike `send_all`, `send_each` does not always throw an exception for a total failure. It can also return a `BatchResponse` with only exceptions in it.

`send_each_for_multicast` calls `send_each` under the hood.

RELEASE NOTE: `send_all()` and `send_multicast()` APIs are now deprecated. Use `send_each()` and `send_each_for_multicast()` APIs instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.7', '3.8', '3.9', '3.10', 'pypy3.7']
+        python: ['3.7', '3.8', '3.9', '3.10', 'pypy3.8']
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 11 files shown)


kevinthecheung (approved):

lahirumaramba (approved):
LGTM! Thank you!

PR #714, opened 2023-08-23
feat: add clockSkewInSeconds

stillmatic:
### Discussion

  * See #624 and #625
  * Implements the feedback in  https://github.com/firebase/firebase-admin-python/pull/625#issuecomment-1331197410 - with the suggested API and additional unit/integration tests

### Testing

  * All unit tests pass with `pytest`. 
  * Lint passes.
  * Integration tests are pretty annoying. Not sure how to create an expired token to use, and the session cookies have a minimum expiry time of 300 seconds - the naive sleep 300 means that tests take 5 min to run. Happy to take guidance on improving that.

### API Changes

  * The API changes here were suggested/approved in the linked discussion.

### DIFF
--- a/firebase_admin/auth.py
+++ b/firebase_admin/auth.py
@@ -256,7 +256,7 @@ def verify_session_cookie(session_cookie, check_revoked=False, app=None, clock_s
         check_revoked: Boolean, if true, checks whether the cookie has been revoked or the
             user disabled (optional).
         app: An App instance (optional).
-        clock_skew_seconds: The number of seconds to tolerate when checking the cookie
+        clock_skew_seconds: The number of seconds to tolerate when checking the cookie.
 
     Returns:
         dict: A dictionary of key-value pairs parsed from the decoded JWT.
--- a/firebase_admin/auth.py
+++ b/firebase_admin/auth.py
@@ -191,7 +191,7 @@ def create_custom_token(uid, developer_claims=None, app=None):
     return client.create_custom_token(uid, developer_claims)
 
 
-def verify_id_token(id_token, app=None, check_revoked=False):
+def verify_id_token(id_token, app=None, check_revoked=False, clock_skew_in_seconds=0):
     """Verifies the signature and data for the provided JWT.
 
     Accepts a signed token string, verifies that it is current, and issued
@@ -202,6 +202,7 @@ def verify_id_token(id_token, app=None, check_revoked=False):
         app: An App instance (optional).
         check_revoked: Boolean, If true, checks whether the token has been revoked or
             the user disabled (optional).
+        clock_skew_in_seconds: The number of seconds to tolerate when checking the token.
 
     Returns:
         dict: A dictionary of key-value pairs parsed from the decoded JWT.
@@ -217,7 +218,8 @@ def verify_id_token(id_token, app=None, check_revoked=False):
             record is disabled.
     """
     client = _get_client(app)
-    return client.verify_id_token(id_token, check_revoked=check_revoked)
+    return client.verify_id_token(
+        id_token, check_revoked=check_revoked, clock_skew_in_seconds=clock_skew_in_seconds)
 
 
 def create_session_cookie(id_token, expires_in, app=None):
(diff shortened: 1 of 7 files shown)


jonathanedey (changes requested):
Thanks @stillmatic for your contribution and getting this feature moving again.

This looks great! I have attached a few comments to be addressed to fully match the approved internal API proposal so please take a look. 

Additionally, we would like to add validation on the `clock_skew_seconds` to limit it to between `0 and 60` seconds and throw a `Value Error` otherwise. Could you add that here with supporting test cases? Thank you!

stillmatic:
I've addressed the feedback in the code change.

Note: this dependency uses `clock_skew_in_seconds` https://github.com/googleapis/google-auth-library-python/blob/7039beb63b8644be748cfc2fc79a2b8b643cda9f/google/oauth2/id_token.py#L112C5-L112C26, So if that gets updated, then this should also be updated. I'm going to assume that moving forward `clock_skew_seconds` is preferred and that you'll handle the discrepancy there.

integration/test_auth.py
```diff
@@ -577,6 +600,19 @@ def test_verify_id_token_revoked(new_user, api_key):
     claims = auth.verify_id_token(id_token, check_revoked=True)
     assert claims['iat'] * 1000 >= user.tokens_valid_after_timestamp
 
+def test_verify_id_token_tolerance(new_user, api_key):
+    expired_id_token = _sign_in_with_password(new_user.email, 'password', api_key)
```
jonathanedey, line 604: Should this be `new_user_with_params` here?

I'm guessing it's passing `None` to get an invalid token which is expired by default. Is this the only way to obtain an expired `id_token`?

integration/test_auth.py
```diff
@@ -577,6 +600,19 @@ def test_verify_id_token_revoked(new_user, api_key):
...
+    time.sleep(1)
+    # Verify the ID token with a tolerance of 0 seconds. This should
+    # raise an exception because the token is expired.
+    with pytest.raises(auth.InvalidIdTokenError) as excinfo:
```
jonathanedey, line 608: Can we catch `ExpiredIdTokenError` here instead?

integration/test_auth.py
```diff
@@ -617,6 +653,44 @@ def test_verify_session_cookie_revoked(new_user, api_key):
...
 
+def test_verify_session_cookie_tolerance(new_user, api_key):
+    expired_session_cookie = auth.create_session_cookie(
+        _sign_in(auth.create_custom_token(new_user.uid), api_key),
+        expires_in=datetime.timedelta(seconds=3)
+    )
+    time.sleep(3)
+    # Verify the session cookie with a tolerance of 0 seconds. This should
+    # raise an exception because the cookie is expired.
+    with pytest.raises(auth.InvalidSessionCookieError) as excinfo:
```
jonathanedey, line 664: Same as above.

jonathanedey (commented):
A few small questions but otherwise it looks good! 
@lahirumaramba do you have anything to add that we might have missed

Thanks for your concern around the discrepancy. We are aware of the convention there but preferred to default to being more in line with our other Firebase SDKs

integration/test_auth.py
stillmatic: I'm honestly not sure -- I couldn't find any integration examples with expired id tokens

lahirumaramba (approved):
LGTM! Thank you! Please take a look at the failing lint tests

stillmatic:
cool - I removed that test and the lint is passing locally now

integration/test_auth.py
stillmatic: yep, lints failed on this test, so deleting this actually fixes that as well!

jonathanedey: `test_session_cookies_with_tolerance`, `test_verify_session_cookie_tolerance` and `test_verify_session_cookie_clock_skew_seconds_range` should also be dropped for the same reason. Thank you!

jonathanedey (approved):
LGTM! Thank you for your contribution.

stillmatic:
thanks for the help reviewing! I've rebased on upstream main and everything should be green.

firebase_admin/auth.py
```diff
@@ -254,6 +256,7 @@ def verify_session_cookie(session_cookie, check_revoked=False, app=None):
         check_revoked: Boolean, if true, checks whether the cookie has been revoked or the
             user disabled (optional).
         app: An App instance (optional).
+        clock_skew_seconds: The number of seconds to tolerate when checking the cookie
```
egilmorez, line 259: Suggest period after "cookie."

egilmorez (approved):
LG with tiny nit, thanks!

Issue #693, opened 2023-04-01
[FR] Connect to firebase storage emulator


prggTheProgrammer:
**Is your feature request related to a problem? Please describe.**
I have a problem described by this [question](https://stackoverflow.com/questions/75420029/connnect-firebase-admin-storage-to-firebbase-emulator-in-python/75894891#comment133884767_75894891). But to summarize I am using firebase storage admin with python and I want to connect it to a firebase emulator running on localhost. I followed the docs [here](https://firebase.google.com/docs/emulator-suite/connect_storage#admin_sdks), added the environment variable, but now I need to get the bucket.

`bucket = storage.bucket("What do I put here?")`
If I don't put anything it gives an error, if I put a real bucket it uses that and I even tried localhost ("127.0.0.1:9199") as bucket name and it couldn't find it because it was searching for a real bucket.

**Describe the solution you'd like**
I'd like a way to get a emulated storage bucket. I think that if the environment variable is set `storage.bucket()` should return that.

**Describe alternatives you've considered**
Any other way it would be possible to use firebase storage admin with the emulator suite would be appreciated.

**Additional context**
As far as I know there is no solution for the problem described above. I don't know if this is a bug or if I'm just being dumb, so I would appreciate if you could point me in the right direction if this is not the right place.

NeoScript:
I am having a similar problem. Similar to #570

lahirumaramba:
Duplicate of #570

Issue #570, opened 2021-09-04
[FR] Add support for storage emulator


PaulRudin:
The python admin sdk does not appear to support the firebase storage emulator. Other emulators are supported, and the node sdk has this support.

qhaas:
We encountered this today, it is listed as 'future' in the [admin sdk emulator support matrix](https://firebase.google.com/docs/emulator-suite/install_and_configure#admin_sdk_availability), hopefully this feature is available soon.

rbarrette:
bump.

lahirumaramba:
Hey @yuchenshi, @christhompsongoogle FYI.

christhompsongoogle:
Hi @vizaiapp, no timeline to share on this just yet but it's still on our radar.

lahirumaramba:
If setting `STORAGE_EMULATOR_HOST` works that means emulator support is now added to the storage library. We still need to map `FIREBASE_STORAGE_EMULATOR_HOST` to `STORAGE_EMULATOR_HOST` internally in the Admin SDK. The change should be similar to: https://github.com/firebase/firebase-admin-go/issues/520

In the meantime, please use `STORAGE_EMULATOR_HOST`

bowenng:
Can we please update the documentation for Python SDK at least?

mcelroyengineering:
I was able to get it working in python by setting my `STORAGE_EMULATOR_HOST` environment variable equal to `http://127.0.0.1:9199`. Can I get confirmation that this is a safe work around for now?

christhompsongoogle:
I spoke to Lahiru, setting STORAGE_EMULATOR_HOST is okay to use the emulator for now. We still need to map FIREBASE_STORAGE_EMULATOR_HOST to STORAGE_EMULATOR_HOST in the SDK

PR #717, opened 2023-08-27
feat(emulator): add support for storage emulator 

Spirans:
#570 #693


Spirans:
@lahirumaramba hi! Is there something I can do to help with this CL? Would you like more information about the problem? Perhaps some stack traces or other details?

Issue #718, opened 2023-09-15
[FR] Add multiple Firestore support


KSokhal:
**Is your feature request related to a problem? Please describe.**
I have a Firebase project that utilised the recent feature to allow for multiple Firestores for a single project and am using Python to perform some firebase admin function, but it currently doesn't allow access to any non-default databases.

**Describe the solution you'd like**
I would like to be able to specify which Firestore to connect to when initializing the connection.

**Describe alternatives you've considered**
Cannot find any alternative to this feature with this package. The only other alternative would be to use Node.js SDK instead since it does have this feature.

liamvdv:
Update?

juansep17:
Hello, are you working on this?

PR #719, opened 2023-09-15
feat(firestore): Added multiple firestore database support

KSokhal:
#718 Added support for multiple Firestore databases within a single app

### DIFF
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -93,7 +93,7 @@ def _get_initialized_app(app):
 
 
 
-def get_app_service(app, options, name, initializer):
+def get_app_service(app, name, initializer, options=None):
     app = _get_initialized_app(app)
     if options:
         app.options._options.update(options)
--- a/firebase_admin/_utils.py
+++ b/firebase_admin/_utils.py
@@ -93,8 +93,10 @@ def _get_initialized_app(app):
 
 
 
-def get_app_service(app, name, initializer):
+def get_app_service(app, options, name, initializer):
     app = _get_initialized_app(app)
+    if options:
+        app.options._options.update(options)
     return app._get_service(name, initializer) # pylint: disable=protected-access
 
 
--- a/firebase_admin/firestore.py
+++ b/firebase_admin/firestore.py
@@ -29,7 +29,7 @@ except ImportError:
                       'to install the "google-cloud-firestore" module.')
 
 from firebase_admin import _utils
-
+from packaging.version import Version
 
 _FIRESTORE_ATTRIBUTE = '_firestore'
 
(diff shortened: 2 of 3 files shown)


firebase_admin/firestore.py
```diff
@@ -50,15 +51,22 @@ def client(app=None):
...
 
 
 class _FirestoreClient:
     """Holds a Google Cloud Firestore client instance."""
 
-    def __init__(self, credentials, project):
-        self._client = firestore.Client(credentials=credentials, project=project)
+    def __init__(self, credentials, project, database_id=None):
+        if database_id:
+            self._client = firestore.Client(credentials=credentials, project=project, database=database_id)
```
daniel-sanche, line 66: note: the `database` parameter is only upported in firestore [v2.12.0+](https://github.com/googleapis/python-firestore/releases/tag/v2.12.0), but this repo supports [v2.9.1](https://github.com/firebase/firebase-admin-python/blob/b992604fb831f19a91430bc699946323e86fb344/setup.py#L43). 

It's probably ok, but we might want to double-check that it fails gracefully when used with an older version

firebase_admin/_utils.py
```diff
@@ -93,8 +93,10 @@ def _get_initialized_app(app):
 
 
 
-def get_app_service(app, name, initializer):
+def get_app_service(app, options, name, initializer):
```
daniel-sanche, line 96: I see this is a private module, but adding a new positional argument into the middle of the list seems dangerous. Can this be an optional kwarg?

firebase_admin/firestore.py
```diff
@@ -50,15 +51,22 @@ def client(app=None):
...
+        else:
+            self._client = firestore.Client(credentials=credentials, project=project)
+
```
daniel-sanche, line 69: we should add some tests for this new functionality

daniel-sanche (commented):
The usage of the firestore client LGTM overall, but I don't have enough context on the rest of the project to give a full approval

firebase_admin/firestore.py
KSokhal: Wasn't sure the best way to do this but I included a version check if attempting to use older that v2.12.0

firebase_admin/_utils.py
KSokhal: Done

firebase_admin/firestore.py
KSokhal: I added a test to cover when an explicit database id is given

dr-nyt:
Any ETA on when this will be pushed? Urgently need this feature.

dr-nyt:
Any ETA on when this will be pushed? I urgently need this feature.

firebase_admin/firestore.py
lahirumaramba: I think we can update the firestore version to the latest in the next release so this check probably won't be necessary (we don't check the version in other SDKs). I understand this PR has been sitting here for a while so I will first merge this and then remove the version check in a follow up PR :)

lahirumaramba (approved):

DracarysBZH:
Hello, is there any news on this PR or is there another way to do it that have been implemented already ? Thanks 🙂

jonathanedey:
Hey folks, want to provide an update on this. We are current going through the API approval process since this alters the public API surface. While doing this we decided to go with a different implementation to fully support caching of the Firestore clients and avoids modifying the app options. With that we will be closing this PR and will complete this change in a separate PR.

Issue #609, opened 2022-02-16
Using Auth Client with Emulator without Credentials Doesn't Work


dgildeh:
After a lot of debugging, it looks like this is the line that is the bug - instead of catching a google.auth.exceptions.DefaultCredentialsError and passing, its trying to catch an AttributeError:

https://github.com/firebase/firebase-admin-python/blob/5e88d927757e61023e274cd8b00270d899fdf513/firebase_admin/__init__.py#L260

This is crashing my requests to get a user via a Rest API using the library on a local environment with the emulator running but no local credentials installed/local SDK. 

Please make sure tests for using the emulator assume no credentials are on the machine or they will break CI/CD pipelines. Any work arounds before a fix is released appreciated!

lahirumaramba:
Hey @dgildeh could you share a bit more on how you set up your environment? Below is what I tried and it seems to be working for me.
```
export FIREBASE_AUTH_EMULATOR_HOST="localhost:9099"
firebase emulators:start --only auth --project fake-project-id

```

dgildeh:
Hi sure:

1. Firebase SDK container installed with Auth/Firestore emulators running inside
2. Backend Python service built inside container with no authentication credentials/gcloud SDK inside with Firebase Admin SDK calling the Firebase Emulator successfully (I've setup the environment variables on the container as you specified already)
3. Both containers running with Docker-Compose locally
4. Client Python service running locally on desktop (not in container so has access to local gcloud SDK on my machine) calling the backend service via gRPC.

I setup remote debugging on the backend service inside the container, which after stepping through the request I saw it failing at that line above and raised this issue.

If you read the code comment for the method I highlighted, its supposed to fall back to environment variables to get the project-id if it can't get local credentials - but because its catching the wrong exception type - its basically crashing the request each time.

I may try to detect if running in Emulator and do firebase_admin.initialize_app() to set the project ID manually as a work around (haven't tried it yet) but ideally my code should just work without any authentication to use the emulator in local dev/test environments. 

What I'm aiming for is dev/test don't require any authentication and just use a local emulator, pushing out via Github actions with tests, and in Google Cloud Run (where I'm deploying) it just picks up the instance service account metadata to make calls to Google Identity Platform/FireStore without any code changes so I don't accidentially put authentication workarounds for local development/testing that cause security issues in prod.

Hope that helps!

dgildeh:
[...]

PR #720, opened 2023-09-19
fix: Correctly catch DefaultCredentialsError when looking up project_id

jonathanedey:
`DefaultCredentialsError` was not being caught when trying to load `project_id` from credentials.

Added a catch for this error along with unit tests.

Related: #609

### DIFF
--- a/firebase_admin/__init__.py
+++ b/firebase_admin/__init__.py
@@ -18,6 +18,7 @@ import json
 import os
 import threading
 
+from google.auth.exceptions import DefaultCredentialsError
 from firebase_admin import credentials
 from firebase_admin.__about__ import __version__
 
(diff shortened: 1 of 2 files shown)


firebase_admin/__init__.py
```diff
@@ -257,7 +258,7 @@ def _lookup_project_id(self):
         if not project_id:
             try:
                 project_id = self._credential.project_id
-            except AttributeError:
+            except (AttributeError, DefaultCredentialsError):
                 pass
```
lahirumaramba, line 262: Do we want to throw here or pass?

lahirumaramba (approved):

firebase_admin/__init__.py
jonathanedey: We want to pass and look for other ways of getting a project_id.

jonathanedey:
Noticed that I am not explicitly triggering the exceptions in the unit tests. Doing that caught that I also needed to clear gcloud env to ensure the exception is thrown.

lahirumaramba (approved):

Issue #726, opened 2023-10-03
Missing ibi (iOS bundle ID) parameter on generated email link authentication URLs


jesuslosada:
### Describe your environment

  * Operating System version: Ubuntu 22.04
  * Firebase SDK version: 6.2.0
  * Firebase Product: auth
  * Python version: 3.11
  * Pip version: 23.2.1

### Describe the problem

The URLs generated by `firebase_admin.auth.generate_sign_in_with_email_link()` include the relevant Android parameters (`apn` - Android package name, `amv` - Android minimum version) but the iOS bundle ID passed through `ActionCodeSettings.ios_bundle_id` is ignored and there is no `ibi` parameter in the resulting URL.

#### Steps to reproduce:

1. Initialize a `firebase_admin.auth.ActionCodeSettings` object assigning a `ios_bundle_id` value.
2. Generate an email link authentication URL by calling `firebase_admin.auth.generate_sign_in_with_email_link()` with the previously generated ActionCodeSettings. 

The parameter named currently used by `firebase-admin` is `iosBundleId` (see https://github.com/firebase/firebase-admin-python/blob/aef52be90951204bc2ce966656fca796f4f87228/firebase_admin/_user_mgt.py#L543), while the expected name is `iOSBundleId` according to the documentation here: https://cloud.google.com/identity-platform/docs/reference/rest/v1/projects.accounts/sendOobCode

#### Relevant Code:

```
import firebase_admin
import firebase_admin.auth


firebase_admin.initialize_app()


def main():
    action_code_settings = firebase_admin.auth.ActionCodeSettings(
        url="https://example.firebaseapp.com/email-link-test",
        handle_code_in_app=True,
        ios_bundle_id="your.app.bundle",
        android_package_name="your.app.package",
        android_install_app=True,
        android_minimum_version="1",
        dynamic_link_domain="example.page.link",
    )
    email = "example@example.com"
    link = firebase_admin.auth.generate_sign_in_with_email_link(
        email, action_code_settings
    )
    print(link)


if __name__ == "__main__":
    main()
```

PR #727, opened 2023-10-03
Fix iOSBundleId parameter name

jesuslosada:
Fix the iOSBundleId parameter name: https://github.com/firebase/firebase-admin-python/issues/726


lahirumaramba (approved):
LGTM! Thank you for the fix!

PR #728, opened 2023-10-04
Added multiple Firestore databases support

KarenjanM:
Hey there! 
I saw an issue, where was discussed adding support for new feature ( Multiple database in a single project )
Here's a link to it: https://github.com/firebase/firebase-admin-python/issues/718

I've tested my solution and it works, would be happy if it could be added to the main package


lahirumaramba:
Duplicate of #719

KarenjanM:
> Duplicate of #719

actually it's not quite a duplicate, because the order of arguments in function get_app_service in 719 is not right and breaks if you try to launch that. The new argument `options` should be as last argument

Issue #704, opened 2023-05-23
app_check.verify_token can also raise PyJWKClientError, but that is not documented


dmelo:
The documentation of `app_check.verify_token` states that the method can raise `ValueError`. But looking into the documentation, we can see that it can also raise an `PyJWKClientError`.

I believe we should add something like:

```
PyJWKClientError: If PyJWKClient fails to fetch the signing key
```

#### Relevant Code:

from https://github.com/firebase/firebase-admin-python/blob/f0865f7493a2c642d2e551efdf19da1e53c1a8c3/firebase_admin/app_check.py#LL38C1-L38C1:

```
    Raises:
        ValueError: If the app's ``project_id`` is invalid or unspecified,
        or if the token's headers or payload are invalid.
```

PR #733, opened 2023-10-24
fix: Add `PyJWKClientError` to raised errors documentation and handle possible uncaught errors.

jonathanedey:
Related to #704


firebase_admin/app_check.py
```diff
@@ -38,6 +38,7 @@ def verify_token(token: str, app=None) -> Dict[str, Any]:
     Raises:
         ValueError: If the app's ``project_id`` is invalid or unspecified,
         or if the token's headers or payload are invalid.
+        PyJWKClientError: If PyJWKClient fails to fetch a valid the signing key.
```
egilmorez, line 41: Omit "the" for "valid signing key."

egilmorez (approved):
With one small nit, good from a verbiage perspective, thanks!

PR #734, opened 2023-10-24
feat(fcm): Enabled the `direct_boot_ok` parameter for FCM Android Config.

jonathanedey:
Enable `direct_boot_ok` in FCM Android Config. 
If set to true, messages will be allowed to be delivered to the app while the device is in direct boot mode.
This allows developers to the parameter through the Admin SDK, a functionality which [already exists and works in HTTP v1](https://firebase.google.com/docs/cloud-messaging/send-message#send_direct_boot-enabled_messages_android_only).


lahirumaramba (approved):
LGTM! Thanks!

egilmorez (approved):
Comments LGTM, thanks!

Issue #678, opened 2023-03-03
Unable to remove email from userRecord


chaitanya4288:
* Operating System version: Ubuntu 22.04
  * Firebase SDK version: (admin-sdk) 6.0.1
  * Firebase Product: auth
  * Python version: 3.10.6
  * Pip version: 22.3.1

unable to remove the email from userRecord using update_user
auth.update_user("xyz", providers_to_delete=['password', 'google.com', 'apple.com', 'facebook.com'])
It only removes the provider data and does not remove the email. 

But using the above method, phone number can be removed. It removes both provider and the phone_number as well
auth.update_user("xyz", providers_to_delete=['phone'])

Is there no other way to remove/delete email from a userRecord?

Adeline09:
Hi I want to work on this issue. Could you assign it to me pls

PR #744, opened 2023-12-11
#678 fixed removing email from userRecord via update_user

smithlauren910:
**Discussion**

This PR resolves issue https://github.com/firebase/firebase-admin-python/issues/678.

**Testing**

Suggestions are welcomed for further testing.

### DIFF
--- a/tests/test_user_mgt.py
+++ b/tests/test_user_mgt.py
@@ -631,12 +631,13 @@ class TestUpdateUser:
             'testuser',
             display_name=auth.DELETE_ATTRIBUTE,
             photo_url=auth.DELETE_ATTRIBUTE,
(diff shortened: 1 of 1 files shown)


Adeline09:
**Description of changes**: 
In firebase_admin/_user_mgt.py we changed the email validation in the payload to include the condition that checks whether email should be deleted via DELETE_ATTRIBUTE. Also we wrote the function to delete email from the database upon checking DELETE_ATTRIBUTE.

**Description of tests**: 
In tests/test_user_mgt.py we included corresponding fields for email in the function test_update_user_delete_fields. Then we wrote the function test_update_user_delete_provider_and_email that verifies email deletion from the database in the same manner as phone number deletion.

prameshj:
cc @renkelvin

PR #747, opened 2023-12-15
Add missing return type for firestore.client()

wliumelb:
firestore.client() should have return type google.cloud.firestore.Client


lahirumaramba (approved):
Thank you! LGTM

PR #748, opened 2023-12-19
chore: Bump github actions CI to use Node 20

jonathanedey:
- `firebase-tools` v13 drops support for node 16 which breaks Continuous Integration.
- Updating CI to use Node 20:
  - actions/checkout: v4
  - actions/setup-node: v4


lahirumaramba (approved):
LGTM! Thanks for the PR!

PR #751, opened 2024-01-02
feat(functions): Add task queue API support

jonathanedey:
Includes implementation of:
- Functions Service
- taskqueue
- enqueue
- delete
- TaskOptions

### DIFF
--- a/firebase_admin/functions.py
+++ b/firebase_admin/functions.py
@@ -1,0 +1,400 @@
+# Copyright 2023 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Firebase Functions module."""
+
+from __future__ import annotations
+from datetime import datetime, timedelta
+from urllib import parse
+import re
+import json
+from base64 import b64encode
+from typing import Any, Optional
+from dataclasses import dataclass
+from google.auth.compute_engine import Credentials as ComputeEngineCredentials
+
+import requests
+import firebase_admin
+from firebase_admin import App
+from firebase_admin import _http_client
+from firebase_admin import _utils
+from firebase_admin import _functions_utils
+
+_FUNCTIONS_ATTRIBUTE = '_functions'
+
+__all__ = [
+    'TaskOptions',
+
+    'task_queue',
+]
+
+
+_CLOUD_TASKS_API_RESOURCE_PATH = \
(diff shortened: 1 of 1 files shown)


firebase_admin/functions.py
```diff
@@ -0,0 +1,409 @@
+# Copyright 2023 Google Inc.
```
lahirumaramba, line 1: nit: 2024

firebase_admin/functions.py
```diff
@@ -0,0 +1,409 @@
...
+
+
+_CLOUD_TASKS_API_RESOURCE_PATH = \
+    'projects/{project_id}/locations/{location_id}/queues/{resource_id}/tasks'
+_CLOUD_TASKS_API_URL_FORMAT = \
+    'https://cloudtasks.googleapis.com/v2/' + _CLOUD_TASKS_API_RESOURCE_PATH
+_FIREBASE_FUNCTION_URL_FORMAT = \
+    'https://{location_id}-{project_id}.cloudfunctions.net/{resource_id}'
+
+_FUNCTIONS_HEADERS = {
```
lahirumaramba, line 50: I think we can move some of these constants out of the global scope and into a mode limited scope. For example, can we defined `_FUNCTIONS_HEADERS` inside the `TaskQueue` class?

firebase_admin/functions.py
```diff
@@ -0,0 +1,409 @@
...
+
+    Raises:
+        ValueError: If the input arguments are invalid.
+    """
+    return _get_functions_service(app).task_queue(function_name, extension_id)
+
+class _FunctionsService:
+    """Service class that implements Firebase Functions functionality."""
+    def __init__(self, app: App):
+        self._project_id = app.project_id
```
lahirumaramba, line 96: let's validate the project id here. For example:
```
if not project_id:
            raise ValueError(
                'Project ID is required to access Cloud Messaging service. Either set the '
                'projectId option, or use service account credentials. Alternatively, set the '
```

firebase_admin/functions.py
```diff
@@ -0,0 +1,409 @@
...
+                headers=_FUNCTIONS_HEADERS,
+                json={'task': task_payload.__dict__}
+            )
+            task_name = resp.get('name', None)
+            task_resource = \
+                self._parse_resource_name(task_name, f'queues/{self._resource.resource_id}/tasks')
+            return task_resource.resource_id
+        except requests.exceptions.RequestException as error:
+            # TODO: Error handle
+            raise FirebaseError(400, 'Error', http_response=error.response)
```
lahirumaramba, line 170: We might be able to use https://github.com/firebase/firebase-admin-python/blob/23765843a124065dcb8fd5a86a3d8f3b784df34c/firebase_admin/_utils.py#L101 here to handle the GCP errors with the correct error codes

lahirumaramba (commented):
Thanks! Looks great overall! Added a few comments.
Let's make sure we have enough test coverage for resource name parsing functions

firebase_admin/functions.py
lahirumaramba: Actually, ignore this. I think the module level constants are okay!

jonathanedey:
Thanks for the review! Going to throw the unit and integration tests into the PR after this one.

lahirumaramba (approved):

PR #752, opened 2024-01-02
Update release and ci workflows to use `GITHUB_OUTPUT` and bump `actions/checkout` to v4.

jonathanedey:


jonathanedey:
Currently we use a forked version of a (no longer maintained) release github action which still uses `set-output`: 
https://github.com/firebase/firebase-admin-python/blob/master/.github/workflows/release.yml#L110

Would we want to consider moving this to a new maintained source like https://github.com/ncipollo/release-action?

lahirumaramba (approved):
Thanks! LGTM!

lahirumaramba:
> Currently we use a forked version of a (no longer maintained) release github action which still uses `set-output`: https://github.com/firebase/firebase-admin-python/blob/master/.github/workflows/release.yml#L110
> 
> Would we want to consider moving this to a new maintained source like https://github.com/ncipollo/release-action?

good call! I think we should. Let's do it in a separate PR though. Is there an official github action for this now?

jonathanedey:
> good call! I think we should. Let's do it in a separate PR though. Is there an official github action for this now?

[The original](https://github.com/actions/create-release) repo [the fork](https://github.com/fleskesvor/create-release/commit/1a72e235c178bf2ae6c51a8ae36febc24568c5fe) was based on was the official one [until they stopped maintaining it in 2021](https://github.com/actions/create-release/issues/119).

They left suggested alternatives with the one i linked being the most active but the thread of mentions shows a few repos who decided to use https://github.com/softprops/action-gh-release instead.

PR #759, opened 2024-01-19
chore: Update PyPi to use trusted publisher for authentication and correctly escape change log body.

jonathanedey:
PyPi no longer accepts username and password authentication for publishing with 2FA enabled accounts. They have additionally made the change to require 2FA on all accounts starting January 1st 2024:
- https://blog.pypi.org/posts/2023-06-01-2fa-enforcement-for-upload/
- https://blog.pypi.org/posts/2024-01-01-2fa-enforced/

Switching our release to use PyPi's [Trusted Publisher](https://docs.pypi.org/trusted-publishers/) method for authentication.

---
Escaped change log body to allow it to correctly show in in the release tag body.


lahirumaramba (approved):
Thanks! LGTM

PR #768, opened 2024-02-13
[chore] Rename pytest teardown methods from deprecated `teardown` to `teardown_method`

jonathanedey:
- `teardown` deprecation warning is now an error in pytest 8.0.0
   - https://github.com/pytest-dev/pytest/releases

- Renamed to suggested `teardown_method`
   - https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose


lahirumaramba (approved):
Thanks! LGTM

PR #764, opened 2024-01-25
feat(functions): Add unit and integration tests for task queue api support

jonathanedey:
Unit and Integration tests

PR #770, opened 2024-02-15
feat(functions): Add task queue API support

jonathanedey:
Task Queue Support for:
- `enqueue`
- `delete`

Code reviewed previously in : #751, #764 and #767

### DIFF
--- a/firebase_admin/functions.py
+++ b/firebase_admin/functions.py
@@ -1,0 +1,422 @@
+# Copyright 2024 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Firebase Functions module."""
+
+from __future__ import annotations
+from datetime import datetime, timedelta
+from urllib import parse
+import re
+import json
+from base64 import b64encode
+from typing import Any, Optional, Dict
+from dataclasses import dataclass
+from google.auth.compute_engine import Credentials as ComputeEngineCredentials
+
+import requests
+import firebase_admin
+from firebase_admin import App
+from firebase_admin import _http_client
+from firebase_admin import _utils
+
+_FUNCTIONS_ATTRIBUTE = '_functions'
+
+__all__ = [
+    'TaskOptions',
+
+    'task_queue',
+]
+
+
+_CLOUD_TASKS_API_RESOURCE_PATH = \
+    'projects/{project_id}/locations/{location_id}/queues/{resource_id}/tasks'
+_CLOUD_TASKS_API_URL_FORMAT = \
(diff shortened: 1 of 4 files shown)


firebase_admin/functions.py
```diff
@@ -0,0 +1,431 @@
...
+            recently then the call will throw an error with code "functions/task-already-exists".
+            Another task with the same ID can't be created for ~1hour after the original task was
+            deleted or executed.
+
+            Because there is an extra lookup cost to identify duplicate task IDs, setting ID
+            significantly increases latency. Using hashed strings for the task ID or for the prefix
+            of the task ID is recommended. Choosing task IDs that are sequential or have sequential
+            prefixes, for example using a timestamp, causes an increase in latency and error rates
+            in all task commands. The infrastructure relies on an approximately uniform distribution
+            of task IDs to store and serve tasks efficiently.
```
kevinthecheung, line 375:
```suggestion
            Because there is an extra lookup cost to identify duplicate task IDs, setting ID
            significantly increases latency.
            
            Also, note that the infrastructure relies on an approximately uniform distribution
            of task IDs to store and serve tasks efficiently.
            For this reason, using hashed strings for the task ID or for the prefix
            of the task ID is recommended. Choosing task IDs that are sequential or have sequential
            prefixes, for example using a timestamp, causes an increase in latency and error rates
            in all task commands. 
```

firebase_admin/functions.py
```diff
@@ -0,0 +1,431 @@
...
+
+            "Push IDs" from the Firebase Realtime Database make poor IDs because they are based on
+            timestamps and will cause contention (slowdowns) in your task queue. Reversed push IDs
+            however form a perfect distribution and are an ideal key. To reverse a string in python
```
kevinthecheung, line 379:
```suggestion
            however form a perfect distribution and are an ideal key. To reverse a string in Python
```

firebase_admin/functions.py
```diff
@@ -0,0 +1,431 @@
...
+            use ``reversedString = someString[::-1]``
+
+        headers: HTTP request headers to include in the request to the task queue function. These
+            headers represent a subset of the headers that will accompany the task's HTTP request.
+            Some HTTP request headers will be ignored or replaced, e.g. Authorization, Host,
+            Content-Length, User-Agent etc. cannot be overridden.
```
kevinthecheung, line 385:
```suggestion
            Some HTTP request headers will be ignored or replaced: `Authorization`, `Host`,
            `Content-Length`, `User-Agent` and others cannot be overridden.
```

kevinthecheung: Also, a complete list would be better

kevinthecheung (approved):

firebase_admin/functions.py
jonathanedey: These are specifically referencing the the headers that Cloud Tasks ignores and replaces. Maybe adding a note to referencing it would be enough?

> A complete list of these ignored or replaced headers can be found in the following definition of the HttpRequest.headers property: https://cloud.google.com/tasks/docs/reference/rest/v2/projects.locations.queues.tasks#httprequest

firebase_admin/functions.py
```diff
@@ -0,0 +1,431 @@
...
+
+            By default, Content-Type is set to 'application/json'.
```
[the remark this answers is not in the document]
jonathanedey: We use the TaskOptions to override this later, so default is correct.
https://github.com/firebase/firebase-admin-python/blob/0482294ff52c427608eaab4fa606ce669d00efb9/firebase_admin/functions.py#L246

kevinthecheung (approved):

PR #806, opened 2024-08-26
chore: Remove Python 3.7 from workflows

lahirumaramba:
- GCP auth libraries have dropped support for Python 3.7
- Our CIs and nightly builds are failing as a result
- Since Python 3.7 is deprecated in the SDK we are removing it from the workflows
- We will drop support for Python 3.7 in the next major release
- Additionally, skip integration tests on batch send APIs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', 'pypy3.8']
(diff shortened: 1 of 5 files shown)


jonathanedey (commented):
We should also update [setup.py](https://github.com/firebase/firebase-admin-python/blob/master/setup.py)

lahirumaramba:
> We should also update [setup.py](https://github.com/firebase/firebase-admin-python/blob/master/setup.py)

I think we can leave that till we officially drop 3.7 support in the next major version. This PR is just to remove the 3.7 versions from our CIs to unblock the release pipeline

jonathanedey (approved):
LGTM!

PR #850, opened 2025-01-09
Adding methods for fetching JSON representation of server template and the value source of the config values

pijushcs:
Hey there! So you want to contribute to a Firebase SDK? 
Before you file this pull request, please read these guidelines:

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here. 
    If not, go file an issue about this **before creating a pull request** to discuss.

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help 
    us make Firebase APIs better, please propose your change in an issue so that we 
    can discuss it together.

### DIFF
--- a/firebase_admin/remote_config.py
+++ b/firebase_admin/remote_config.py
@@ -17,6 +17,7 @@ This module has required APIs for the clients to use Firebase Remote Config with
 """
 
 import asyncio
+import json
 import logging
 from typing import Dict, Optional, Literal, Union, Any
 from enum import Enum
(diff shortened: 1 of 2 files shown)


firebase_admin/remote_config.py
```diff
@@ -161,22 +171,40 @@ def evaluate(self, context: Optional[Dict[str, Union[str, int]]] = None) -> 'Ser
...
-    def set(self, template: ServerTemplateData):
+    def set(self, template_data_json: str):
         """Updates the cache to store the given template is of type ServerTemplateData.
 
         Args:
-          template: An object of type ServerTemplateData to be cached.
+          template_data_json: A json string representing ServerTemplateData to be cached.
         """
-        self._cache = template
+        template_data_map = json.loads(template_data_json)
```
lahirumaramba, line 194: do we need to validate the json here or are we okay if `json.loads` throws on invalid formats?

lahirumaramba (approved):
LGTM! Thanks!
Added one comment

firebase_admin/remote_config.py
pijushcs: We are okay if json.loads throws a ValueError in this case so that the developer can handle it later.

PR #851, opened 2025-01-13
chore: Add tests for `Reference.listen()`

jonathanedey:
Follow up PR for #845 with unit and integration tests.

These tests ensure that the `sseClient` is setup with the necessary parameters in both production and emulator mode.


lahirumaramba (approved):
Thanks! LGTM!

PR #868, opened 2025-03-18
feat(fcm): Support `proxy` field in FCM `AndroidNotification`

jonathanedey:
* Adds `proxy` field to `AndroidNotification`

### DIFF
--- a/firebase_admin/_messaging_utils.py
+++ b/firebase_admin/_messaging_utils.py
@@ -137,7 +137,8 @@ class AndroidNotification:
             If ``default_light_settings`` is set to ``True`` and ``light_settings`` is also set, the
             user-specified ``light_settings`` is used instead of the default value.
         visibility: Sets the visibility of the notification. Must be either ``private``, ``public``,
-            or ``secret``. If unspecified, default to ``private``.
+            or ``secret``. If unspecified, it remains undefined in the Admin SDK, and defers to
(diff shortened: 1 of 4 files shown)


firebase_admin/_messaging_utils.py
```diff
@@ -145,6 +145,8 @@ class AndroidNotification:
             want the count here to represent the number of total new messages. If zero or
             unspecified, systems that support badging use the default, which is to increment a
             number displayed on the long-press menu each time a new notification arrives.
+        proxy: Sets if the notification may be proxied. Must be one of ``allow``, ``deny``, or
+            ``if_priority_lowered``. If unspecified, defaults to ``if_priority_lowered``.
```
OrlandriaH-G, line 149: Suggested update for the default sentence: "If unspecified, it will remain undefined and no default value will be applied."

OrlandriaH-G, line 149: Suggested text changes: If unspecified, it remains undefined in the Admin SDK, and defers to the FCM backend's default mapping.

OrlandriaH-G (changes requested):

firebase_admin/_messaging_utils.py
jonathanedey: Done!

OrlandriaH-G (approved):

PR #871, opened 2025-04-14
Refactored retry config into `_retry.py` and added support for exponential backoff and `Retry-After` header

jonathanedey:
Adds more retry logic support

This PR covers:
- Moving  and refactoring `HttpxRetryTransport` to `_retry.py`
- Adding new class `HttpxRetry` to manage retry state
- Adding support for setting some retry configurations on `HttpxRetryTransport` creation
- Performing exponential backoff before request retries
- Respecting `Retry-After` headers from request responses
- Unit tests for `HttpxRetry` and `HttpxRetryTransport`
- Unit test for handing HTTPX request errors

### DIFF
--- a/firebase_admin/_retry.py
+++ b/firebase_admin/_retry.py
@@ -35,8 +35,8 @@ logger = logging.getLogger(__name__)
 class HttpxRetry:
     """HTTPX based retry config"""
     # TODO: Decide
-    # urllib3.Retry ignores the status_forcelist and only respects Retry-After header
-    # for 413, 429 and 503 errors.
+    # urllib3.Retry ignores the status_forcelist when respecting Retry-After header
+    # Only 413, 429 and 503 errors are retried with the Retry-After header.
     # Should we do the same?
     # Default status codes to be used for ``status_forcelist``
     RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])
@@ -123,7 +123,7 @@ class HttpxRetry:
 
     def get_retry_after(self, response: httpx.Response) -> float | None:
         """Determine the Retry-After time needed before sending the next request."""
-        retry_after_header = response.headers.get('Retry_After', None)
+        retry_after_header = response.headers.get('Retry-After', None)
         if retry_after_header:
             # Convert retry header to a float in seconds
             return self._parse_retry_after(retry_after_header)
@@ -131,9 +131,12 @@ class HttpxRetry:
 
     def get_backoff_time(self):
         """Determine the backoff time needed before sending the next request."""
-        # request_count is the number of previous request attempts
-        request_count = len(self.history)
-        backoff = self.backoff_factor * (2 ** (request_count-1))
+        # attempt_count is the number of previous request attempts
+        attempt_count = len(self.history)
+        # Backoff should be set to 0 until after first retry.
+        if attempt_count <= 1:
+            return 0
+        backoff = self.backoff_factor * (2 ** (attempt_count-1))
         if self.backoff_jitter:
             backoff += random.random() * self.backoff_jitter
         return float(max(0, min(self.backoff_max, backoff)))
@@ -141,7 +144,7 @@ class HttpxRetry:
     async def sleep_for_backoff(self) -> None:
         """Determine and wait the backoff time needed before sending the next request."""
         backoff = self.get_backoff_time()
-        logger.debug('Sleeping for %f seconds following failed request', backoff)
+        logger.debug('Sleeping for backoff of %f seconds following failed request', backoff)
         await asyncio.sleep(backoff)
 
     async def sleep(self, response: httpx.Response) -> None:
@@ -149,6 +152,10 @@ class HttpxRetry:
         if self.respect_retry_after_header:
             retry_after = self.get_retry_after(response)
             if retry_after:
+                logger.debug(
+                    'Sleeping for Retry-After header of %f seconds following failed request',
+                    retry_after
+                )
                 await asyncio.sleep(retry_after)
                 return
         await self.sleep_for_backoff()
--- a/firebase_admin/_retry.py
+++ b/firebase_admin/_retry.py
@@ -1,0 +1,246 @@
+# Copyright 2025 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Internal retry logic module
+
+This module provides utilities for adding retry logic to HTTPX requests
+"""
+
+from __future__ import annotations
+import copy
+import email.utils
+import random
+import re
+import time
+from types import CoroutineType
+from typing import Any, Callable, List, Optional, Tuple
+import logging
+import asyncio
+import httpx
+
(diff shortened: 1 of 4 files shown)


firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+                self.respect_retry_after_header
+                and has_retry_after
+                and response.status_code in self.RETRY_AFTER_STATUS_CODES
+        ):
+            return True
+
+        return False
+
+    # Placeholder for exception retrying
+    def is_retryable_error(self, error: Exception):
```
lahirumaramba, line 94: It is best not to leave placeholders in PRs if we can. If this is currently a no-op we can remove it and add it when we need it

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+            response: Optional[httpx.Response] = None,
+            error: Optional[Exception] = None
+    ) -> None:
+        """Update the retry state based on request attempt."""
+        if response and self.is_retryable_response(response):
+            self.status -= 1
+        self.history.append((request, response, error))
+
+
+# TODO: Remove comments
```
lahirumaramba, line 175: You can remove these. Or add a comment on what it doesn't cover as a TODO for future. I prefer we use a bug/issue to track these instead of TODO comments when possible

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+    # Should we do the same?
+    # Default status codes to be used for ``status_forcelist``
+    RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])
+
+    #: Default maximum backoff time.
+    DEFAULT_BACKOFF_MAX = 120
+
+    def __init__(
+            self,
+            status: int = 10,
```
lahirumaramba, line 49: Would it make sense to call this `max_retries` instead?

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+
+            # First retry
+            if response:
+                await retry.sleep(response)
+
+            # Need to reset here so only last attempt's error or response is saved.
+            response, error = None, None
+
+            try:
+                logger.debug('Sending request in _dispatch_with_retry(): %r', request)
```
lahirumaramba, line 231: Do we plan to keep these logs in the production code?

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+                logger.debug('Received error: %r', err)
+                error = err
+
+            if response and not retry.is_retryable_response(response):
+                return response
+
+            if error and not retry.is_retryable_error(error):
+                raise error
+
+            retry.increment(request, response)
```
lahirumaramba, line 244: Should we pass the `error` here? `retry.increment(request, response, error)`

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+
+        if response:
+            return response
+        if error:
+            raise error
+        raise Exception('_dispatch_with_retry() ended with no response or exception')
```
lahirumaramba, line 250: Let's narrow this down to an Exception type `RuntimeError` or `AssertionError` might work better here.

firebase_admin/messaging.py
```diff
@@ -38,7 +39,9 @@
     exceptions,
     App
 )
+from firebase_admin._retry import HttpxRetryTransport
 
+logger = logging.getLogger(__name__)
```
lahirumaramba, line 44: Is there a way to set the log level in production code?

tests/test_retry.py
```diff
@@ -0,0 +1,451 @@
...
+
+        # Simulate attempt 5 completed
+        retry.increment(self._TEST_REQUEST, response)
+        # History len 5, attempt 6 -> 0.5*(2^4) = 8.0
+        assert retry.get_backoff_time() == pytest.approx(8.0)
+
+        # Simulate attempt 6 completed
+        retry.increment(self._TEST_REQUEST, response)
+        # History len 6, attempt 7 -> 0.5*(2^4) = 10.0
+        assert retry.get_backoff_time() == pytest.approx(10.0)
```
lahirumaramba, line 451: Should this be `0.5 * (2^5) = 16.0`? and clammed at 10 due to `backoff_max`? If that's the case let's clarify it here :)

lahirumaramba (commented):
Thanks Jonathan! Overall looking good! Added a few comments!

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+                Optional[httpx.Response],
+                Optional[Exception]
+            ]]] = None,
+            respect_retry_after_header: bool = False,
+    ) -> None:
+        self.status = status
+        self.status_forcelist = status_forcelist
+        self.backoff_factor = backoff_factor
+        self.backoff_max = backoff_max
+        self.raise_on_status = raise_on_status
```
lahirumaramba, line 66: If `raise_on_status` is unused we should just remove it

firebase_admin/_retry.py
```diff
@@ -0,0 +1,253 @@
...
+import asyncio
+import httpx
+
+logger = logging.getLogger(__name__)
+
+
+class HttpxRetry:
+    """HTTPX based retry config"""
+    # TODO: Decide
+    # urllib3.Retry ignores the status_forcelist when respecting Retry-After header
```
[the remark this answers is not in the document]
jonathanedey: I think I worded that comment poorly so you can disregard it.

What I wanted to point out that in `urllib3.Retry` there is no default status codes that are retried if `status_forcelist` and `respect_retry_after_header` both are not set. These `RETRY_AFTER_STATUS_CODES` are only applied if `respect_retry_after_header` is set.

I decided to do the same in this implementation but still wanted to highlight it because [the urllib3 docs were misleading](https://urllib3.readthedocs.io/en/2.3.0/reference/urllib3.util.html#urllib3.util.Retry.RETRY_AFTER_STATUS_CODES) and implied these are used as `status_forcelist` defaults.

firebase_admin/_retry.py
jonathanedey: In the current state yes because status retries are our only retries we have. This was more of a future proof decision for when we added error retires where we would have total, status and error counters. 

Similar to the placeholder comment I think we can use `max_retries` (or maybe `retries_left` since we decrease this value) and add the other options later.

firebase_admin/_retry.py
jonathanedey: I would want to keep these for future debugging. This would be helpful to catch issues as we iterate. Wdyt?

firebase_admin/messaging.py
jonathanedey: Developers should be able to set the logging level in their production apps by using the following code:
```py
import logging

logging.basicConfig()
firebase_admin_logger = logging.getLogger('firebase_admin')
firebase_admin_logger.setLevel(logging.DEBUG)
```

firebase_admin/_retry.py
jonathanedey: Done!

firebase_admin/_retry.py
jonathanedey: Sounds good, removed!

firebase_admin/_retry.py
jonathanedey: Done, these todos were mainly markers so as to not miss removing some thought process comments before final merge.

Moving the missing features notes to a bugs.

firebase_admin/_retry.py
jonathanedey: Yes, good catch! Fixed

firebase_admin/_retry.py
jonathanedey: Let's go with `AssertionError` since this case should never be reached if the logic above is correct.

tests/test_retry.py
jonathanedey: Yeah that should be the correct calculation, fixed!

PR #873, opened 2025-04-22
chore: Update README to deprecate Python 3.8 support

lahirumaramba:
Python 3.8 has EoL'ed. Update README to deprecate Python 3.8 support

Updated the 'Supported Python Versions' section in README.md to indicate that Python 3.7 and Python 3.8 support is deprecated, advising users to use Python 3.9 or higher.


jonathanedey (approved):
LGTM, Thanks!

PR #876, opened 2025-05-01
chore: Correct `x-goog-api-client` header logic

jonathanedey:
* Lower cased `x-goog-api-client`
* Moved `x-goog-api-client` to be applied on request rather that on session
* Fixed tests to simulate `x-goog-api-client` mutation by `AuthorizedSession`


lahirumaramba (approved):
Thanks for this fix! LGTM!

PR #877, opened 2025-05-20
Add type annotations

ViktorSky:
This pull request introduces **comprehensive type annotations** to the `firebase_admin` codebase using [PEP 484](https://peps.python.org/pep-0484/) standards. All changes have been validated with **Pyright** in **strict** mode to enforce consistency and catch potential type errors early.

#### Goals

- Enhance developer experience with precise autocomplete and type hints.
- Catch type-related bugs upfront via static analysis.

#### Tools

- **Pyright** in **strict** mode for robust type-checking.
- Manual review of ambiguous or potentially problematic sections.

#### Configuration Additions
- `pyrightconfig.json`: config file created to declare strict-mode settings.

- `firebase_admin/_typing.py`: A new central module that defines all custom type aliases, TypedDicts, and Protocols. By consolidating these shared types in one place, we simplify extension and reuse across the package and avoid circular-import issues.

### DIFF
--- a/firebase_admin/__about__.py
+++ b/firebase_admin/__about__.py
@@ -14,7 +14,7 @@
 
 """About information (version, etc) for Firebase Admin SDK."""
 
-__version__ = '6.8.0'
+__version__ = '6.9.0'
 __title__ = 'firebase_admin'
 __author__ = 'Firebase'
 __license__ = 'Apache License 2.0'
--- a/firebase_admin/__init__.py
+++ b/firebase_admin/__init__.py
@@ -17,23 +17,31 @@ import datetime
 import json
 import os
 import threading
+import typing
 
 from google.auth.credentials import Credentials as GoogleAuthCredentials
 from google.auth.exceptions import DefaultCredentialsError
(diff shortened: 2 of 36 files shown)


ViktorSky:
Before approving this request, I must warn you that you cannot use pylint, because the configured version of `astroid` can only be used on versions lower than py3.11.
I also left some comments in the code about potential runtime issues.

- `firebase_admin._messaging_encoder._Validators`
Why not `@staticmethod`? Perhaps there were past issues or subclassing considerations?

- `firebase_admin.remote_config._ConditionEvaluator.evaluate_conditions`
You probably need to change condition.get('name') to condition['name']

- `firebase_admin.remote_config._ConditionEvaluator._compare_semantic_versions`
bad docstring

- `firebase_admin.remote_config._ConditionEvaluator.evaluate`
parameter_value_wrapper.get('value') is only valid if None is a valid argument to `_Value`

- `firebase_admin.functions.TaskQueue._update_task_payload`
`service_account_email` is accessed on a type that might not have that attribute

Issue #883, opened 2025-05-30
[FR] Update ActionCodeSettings to support link_domain


huwmartin:
**Is your feature request related to a problem? Please describe.**

[Firebase Dynamic Links will stop working August 25th 2025](https://firebase.google.com/support/dynamic-links-faq).

[Migration instructions indicate that to use a custom domain one should update usage of `ActionCodeSettings` to use `linkDomain`](https://firebase.google.com/docs/auth/ios/email-link-migration#reuse-custom-dynamic-links-domain), see also in the [Firebase Admin docs](https://firebase.google.com/docs/auth/admin/email-action-links#initialize-actioncodesettings).

We currently depend upon `firebase-admin-python` `generate_sign_in_with_email_link` to generate email links, [`firebase-admin-python` hasn't been updated with `link_domain` and currently only supports `dynamic_link_domain`](https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/_user_mgt.py#L492), [`firebase-admin-node` has been updated with support for `linkDomain`](https://github.com/firebase/firebase-admin-node/blob/master/src/auth/action-code-settings-builder.ts#L109) so I assume this should be officially supported, please let me know if this is not the case.

**Describe the solution you'd like**

[`ActionCodeSettings` should be updated to accept `link_domain`](https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/_user_mgt.py#L496) and [`encode_action_code_settings` should be updated to handle `link_domain`](https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/_user_mgt.py#L531-L536).

PR #884, opened 2025-06-03
feat(auth) Update `ActionCodeSettings` to support `link_domain` and deprecate `dynamic_link_domain`

huwmartin:
### Discussion

[Firebase Dynamic Links will stop working August 25th 2025](https://firebase.google.com/support/dynamic-links-faq).

[Migration instructions indicate that to use a custom domain one should update usage of `ActionCodeSettings` to use `linkDomain`](https://firebase.google.com/docs/auth/ios/email-link-migration#reuse-custom-dynamic-links-domain), see also in the [Firebase Admin docs](https://firebase.google.com/docs/auth/admin/email-action-links#initialize-actioncodesettings).

We currently depend upon `generate_sign_in_with_email_link` to generate email links, [`firebase-admin-python` hasn't been updated with `link_domain` and currently only supports `dynamic_link_domain`](https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/_user_mgt.py#L492), [`firebase-admin-node` has been updated with support for `linkDomain`](https://github.com/firebase/firebase-admin-node/blob/master/src/auth/action-code-settings-builder.ts#L109) so I assume this should be officially supported, please let me know if this is not the case.

See https://github.com/firebase/firebase-admin-python/issues/883.

This PR updates `ActionCodeSettings` to accept `link_domain` and updates `encode_action_code_settings` to handle `link_domain` when constructing request payload. I have also added definition for new `InvalidHostingLinkDomainError` error.


errkk (approved):

firebase_admin/_auth_utils.py
```diff
@@ -328,6 +328,15 @@ def __init__(self, message, cause, http_response):
         exceptions.InvalidArgumentError.__init__(self, message, cause, http_response)
 
 
+class InvalidHostingLinkDomainError(exceptions.InvalidArgumentError):
+        """Hosting link domain in ActionCodeSettings is not authorized."""
```
Xiaoshouzi-gh, line 332: Can you update both msgs to be
"Hosting link domain in ActionCodeSettings is not authorized or is not owned by the current project."? to be more aligned with node SDK.  See: https://github.com/firebase/firebase-admin-node/blob/26b884f70300ef8e3213dfbbd72351ce7f68c695/src/utils/error.ts#L510

Xiaoshouzi-gh (commented):
Thanks for putting this together!

firebase_admin/_auth_utils.py
huwmartin: Certainly, appreciate you putting eyes on this. Updated in https://github.com/firebase/firebase-admin-python/pull/884/commits/bc7fcfcf7f052b9c569fc0cf17c1bf596db73804.

Xiaoshouzi-gh (commented):
Please fix the lint: https://github.com/firebase/firebase-admin-python/actions/runs/16007848807/job/45494220646?pr=884

jonathanedey (changes requested):
Thanks @huwmartin for putting this together! Overall LGTM with one minor change below.

PR #892, opened 2025-06-16
chore(deps): Bump minimum supported Python version to 3.9 and add 3.13 to CIs

jonathanedey:
* Bumped minimum supported Python version to 3.9
* Added python 3.13 to CIs
* Bumped test dependancies
* Removed usage of `setup.py` [as a command line tool since this is now deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#is-setup-py-deprecated). 
* Replaced usage of deprecated `download_as_string()` with `download_as_bytes()` in storage tests.

RELEASE NOTE: Dropped support for Python 3.7 and 3.8. Developers should use Python 3.9 or higher with the Admin SDK.

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -16,7 +16,6 @@
 import json
 
 import pytest
-from pytest_asyncio import is_async_test
 
 import firebase_admin
 from firebase_admin import credentials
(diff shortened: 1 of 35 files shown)


integration/conftest.py
```diff
@@ -71,9 +70,3 @@ def api_key(request):
                          'command-line option.')
     with open(path) as keyfile:
         return keyfile.read().strip()
-
-def pytest_collection_modifyitems(items):
```
lahirumaramba, line 75: Is this what causing the CI errors?

lahirumaramba (approved):
LGTM! Thanks!

PR #908, opened 2025-08-26
fix(auth): Fixed auth error code parsing

jonathanedey:
The auth error code parser incorrectly mapped known error codes to unknown errors due to additional whitespace from the expected auth error message. This whitespace is now trimmed and correctly mapped to the corresponding auth error code.


lahirumaramba (approved):
Thank you!

PR #919, opened 2025-11-04
Add pipeline snippets for Firestore

morganchen12:


markarndt (approved):

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+
+def functions_example():
+    # [START functions_example]
+    # Type 1: Scalar (for use in non-aggregation stages)
+    # Example: Return the min store price for each book.
+    results = client.pipeline().collection("books") \
+        .select(
+            Field.of("current").logical_minimum(Field.of("updated")).as_("price_min")
+        ) \
+        .execute()
```
daniel-sanche: nit: I would probably merge this with the above line

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+            Field.of("latitudeDifference").add(Field.of("longitudeDifference")).sqrt()
+                # Inaccurate for large distances or close to poles
+                .as_("approximateDistanceToGoogle")
+        ) \
+        .execute()
+    # [END pow_function]
+    for res in result:
+        print(res)
+
+def sqrt_function():
```
daniel-sanche: This seems like a duplicate of the above? If that's really what you want, I think you can stack multiple region tags over the same sample

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+    # [END exists_function]
+    for res in result:
+        print(res)
+
+def and_function():
+    # [START and_function]
+    result = client.pipeline() \
+        .collection("books") \
+        .select(
+            And(
```
daniel-sanche: Maybe you should include the import in the sample? `from google.cloud.firestore_v1.pipeline_expressions import And`. I guess this feedback would apply to Field and Constant too though...

Our usual policy was that samples should all include relevant imports. Most of these pipeline tests don't really need imports. But when we start adding Ands and Ors and Nots, I could see users getting confused, since they look so close to built-in methods

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+    # [START vector_length]
+    result = client.pipeline() \
+        .collection("books") \
+        .select(
+            Field.of("embedding").vector_length().as_("vectorLength")
+        ) \
+        .execute()
+    # [END vector_length]
+    for res in result:
+        print(res)
```
daniel-sanche: We usually use the [`black` code formatter](https://pypi.org/project/black/) for our Python code. If you haven't used it, it might be good to run it over this file to see if it has any changes

Nothing really stands out to me as badly formatted, and I don't think using that formatting should be a requirement for samples. But it might be worth checking

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+
+def aggregate_groups():
+    # [START aggregate_groups]
+    results = client.pipeline() \
+        .collection("books") \
+        .aggregate(
+            Field.of("rating").average().as_("avg_rating"),
+            groups=[Field.of("genre")]
+        ) \
+        .execute()
```
daniel-sanche: If you want to get rid of all the \\, IIRC think you could do something like this:

```
results = (
    client.pipeline()
    .collection("books")
    .aggregate(
        Field.of("rating").average().as_("avg_rating"),
        groups=[Field.of("genre")]
    ).execute()
)
```

(But test it first)

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+
+import firebase_admin
+from firebase_admin import firestore
+
+default_app = firebase_admin.initialize_app()
+client = firestore.client(default_app, "your-new-enterprise-database")
+
+# pylint: disable=invalid-name
+def pipeline_concepts():
+    # [START pipeline_concepts]
```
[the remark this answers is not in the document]
morganchen12: Yeah, I omitted `firestore_` where the filepath contained `firestore` somewhere. The snippets tool and the includecode tool both reference the filepath when importing statements so it'll still be findable as a Firestore snippet.

snippets/firestore/firestore_pipelines.py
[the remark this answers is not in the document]
morganchen12: Will defer to another PR

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1289 @@
...
+    pipeline = client.pipeline() \
+        .collection("cities") \
+        .where(Field.of("population").greater_than(100000)) \
```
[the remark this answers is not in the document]
morganchen12: Done here and throughout

snippets/firestore/firestore_pipelines.py
daniel-sanche: My understanding is no, and I've definitely seen other samples that double up like that. (Here's [an example](https://github.com/googleapis/python-logging/blob/a7a28d1b9396e472f153fd76d0d846338c107d5a/samples/snippets/handler.py#L20))

But if you want to keep them separate, that's fine

daniel-sanche (approved):
Consider adding more imports for Vector, and others that may be missing through the file

But overall LGTM

snippets/firestore/firestore_pipelines.py
```diff
@@ -0,0 +1,1753 @@
...
+    # [END timestamp_unix_seconds]
+    for res in result:
+        print(res)
+
+
+def cosine_distance_function():
+    # [START cosine_distance]
+    from google.cloud.firestore_v1.pipeline_expressions import Field
+
+    sample_vector = Vector([0.0, 1.0, 2.0, 3.0, 4.0, 5.0])
```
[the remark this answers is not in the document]
morganchen12: Done.

daniel-sanche:
1. One of those is `from google.cloud import firestore`, and the other is `from firebase_admin import firestore`. I'm not super familiar with the `firebase_admin` wrapper, but I don't think those are fully equivalent

2. In the end, I think Array was actually implemented as a stand-alone class:

```
from google.cloud.firestore_v1.pipeline_expressions import Array
Array(["bar", Field.of("baz")])
```
It looks like there's an [out-dated docstring](https://github.com/googleapis/python-firestore/blob/4848fbe29fbed70f35067de694314dc91c2b68f1/google/cloud/firestore_v1/pipeline_expressions.py#L1853). I can fix that. 

Let me know if you think it's important to have `array` exposed on Expression as well though. I removed it in an effort to keep the API surface manageable, instead of exposing multiple ways of expressing the same thing. But I can add it back if that's more consistent with other languages.

4. array_get was a late addition, but it should be in the preview branch as of yesterday. Maybe try running again

snippets/firestore/firestore_pipelines.py
```diff
@@ -47,6 +20,19 @@
...
+
+def query_explain():
+    # [START query_explain]
+    from google.cloud.firestore import Query, FieldFilter, ExplainOptions
+
+    results = client.collection("cities") \
+        .where(filter=FieldFilter("capital", "==", True)) \
+        .execute(explain_options=ExplainOptions(analyze=False))
+    metrics = results.explain_metrics()
+    summary = metrics.plan_summary()
```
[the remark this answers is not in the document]
morganchen12: Should be fixed now. You can republish the docs on devsite and it'll pick up the fix.

PR #926, opened 2025-12-12
chore: Update default branch to `main`

lahirumaramba:
Updated default branch to `main`

If you have a local clone, you can update it by running the following commands:
`master is now named main`
```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

Optionally, run the following command to remove tracking references to the old branch name.
```
git remote prune origin
```

Side note: Update the CI workflow to set the Java version to match the supported versions in `firebase-tools`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,12 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: 20
+    - name: Set up Java 21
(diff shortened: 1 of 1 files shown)


jonathanedey (approved):
LGTM, Thanks! Could we add a note to the PR description to highlight why we needed to set the java version here for future reference.

Issue #860, opened 2025-02-06
[FR] Missing type hints


JohnGale87:
**Is your feature request related to a problem? Please describe.**
Mypy is reporting: 
```
Skipping analyzing "firebase_admin": module is installed, but missing library stubs or py.typed marker
Mypy[import-untyped](https://mypy.readthedocs.io/en/latest/_refs.html#code-import-untyped)

See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Mypy[note](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports)
```

**Describe the solution you'd like**
I would like Mypy to be able to provide valid type hints and stop warning me that it can't.

**Describe alternatives you've considered**
Use `  # type: ignore` to ignore the warning.

andersonaddo:
See https://github.com/firebase/firebase-admin-python/pull/713

PR #929, opened 2026-01-05
Add type hints to FirebaseError in exceptions.py

Niharika-Sathish26:
Just added few type hints to the FirebaseError class in exceptions.py to make it easier to work with IDEs and catch type issues. The constructor now clearly shows the types for code, message, cause, and http_response. No functionality has changed, and all existing tests still pass. This is a small quality-of-life improvement that should help anyone using this module in their projects.

Related issue: #860


ViktorSky:
https://github.com/firebase/firebase-admin-python/pull/923

PR #952, opened 2026-05-07
chore: Resolved CI build failures on PyPy 3.9 by pinning `cryptography`

lahirumaramba:
Newer versions of the `cryptography` package (>= 44.0.0) upgraded their internal Rust `PyO3` dependency to a version that requires PyPy >= 3.11. Consequently, building `cryptography` from source fails in older PyPy environments like `pypy3.9`. 

This PR introduces an environment marker constraint to restrict `cryptography < 44.0.0` on PyPy interpreters older than 3.11, ensuring stable dependency resolution and restoring successful builds.


jonathanedey (approved):
LGTM, Thanks!

PR #965, opened 2026-07-09
feat(dataconnect): Implementation and Testing of the _DataConnectApiClient class

mk2023:
## Overview
✨ Implements `_DataConnectApiClient` core request preparation helpers (`_validate_graphql_options`, `_prepare_graphql_payload`, `_get_firebase_dataconnect_service_url`, `_get_headers`), refactors `Impersonation` to inherit from `dict`, and adds unit tests for input validation and payload serialization.

## Highlights
✨ Key changes:
* **Option & Variable Validation (`_validate_graphql_options`)**: Added strict runtime validation for GraphQL options, variables (mappings and dataclasses with generic type checking via `typing.get_origin`), operation names, and impersonation options.
* **Payload Serialization (`_prepare_graphql_payload`)**: Added GraphQL JSON payload construction supporting raw dictionaries, dataclasses (`asdict`), operation names, and impersonation extensions.
* **Service URL & Headers (`_get_firebase_dataconnect_service_url`, `_get_headers`)**: Added dynamic endpoint URL construction for production/emulator environments and injected standard SDK metrics headers (`X-Firebase-Client` and `x-goog-api-client`).
* **Impersonation Refactoring (`Impersonation`)**: Updated `Impersonation` to inherit directly from `dict` for improved type compatibility while retaining dictionary runtime behavior.
* **Unit Tests (`tests/test_data_connect.py`)**: Added unit test coverage for `_DataConnectApiClient` constructor, validation methods, header generation, and nested dataclass payload serialization.

<details>

### Core SDK Changes
  - Implemented `_DataConnectApiClient` constructor and property setup.
  - Implemented `_validate_variables_type`, `_validate_impersonation_options`, and `_validate_graphql_options`.
  - Implemented `_prepare_graphql_payload` to convert query, variables, operation name, and impersonation claims into a JSON payload.
  - Implemented `_get_firebase_dataconnect_service_url` and `_get_headers`.
  - Updated `Impersonation` class hierarchy (`class Impersonation(dict)`).
- **`firebase_admin/_utils.py` & `firebase_admin/functions.py`**:
  - Added shared `get_emulator_host` utility helper for reading emulator host environment variables.

### Unit Test Changes
- **`tests/test_data_connect.py` & `tests/test_utils.py`**:
  - Added unit test cases for client constructor (invalid app, missing project ID, emulator host validation).
  - Added unit test cases for `_validate_graphql_options` (valid/invalid variables, dataclasses, generic mappings, operation names, impersonation).
  - Added unit test cases for `_prepare_graphql_payload` (queries, raw dict variables, nested dataclasses, operation names, impersonation).
  - Added unit test cases for `get_emulator_host` utility.

</details>

---

### Testing Strategy
- **Unit Tests**: `pytest tests/test_data_connect.py tests/test_utils.py` (PASSED)

---

Context Sources Used:
- id: firebase-admin-python


tests/test_data_connect.py
```diff
@@ -331,3 +331,285 @@ def test_overall_client_retrieval_and_caching(self):
...
+    def setup_method(self):
+        self.cred = testutils.MockCredential()
+
+    def teardown_method(self, method):
+        del method
+        testutils.cleanup_apps()
+
+    def test_constructor_invalid_app(self):
+        with pytest.raises(ValueError, match="First argument passed to DataConnectApiClient must be a valid Firebase app instance."):
+            dataconnect._DataConnectApiClient(BASE_CONFIG, None)
```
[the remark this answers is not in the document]
mk2023: Addded in new commit!

tests/test_data_connect.py
```diff
@@ -331,3 +331,285 @@ def test_overall_client_retrieval_and_caching(self):
         assert client1_app2.app is self.app2
         assert client1_app2.config is self.config1
         assert client1_app2 is not client1a
+
+
+class TestDataConnectApiClientConstructor:
```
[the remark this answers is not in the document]
mk2023: Added!

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+class Impersonation:
+    """Represents impersonation configuration for DataConnect requests."""
+
+    @staticmethod
+    def unauthenticated() -> Dict[str, bool]:
+        """Returns impersonation configuration for unauthenticated requests."""
+        return {"unauthenticated": True}
+
+    @staticmethod
+    def authenticated(auth_claims: Dict[str, Any]) -> Dict[str, Any]:
```
stephenarosaj: is there a way to more strongly type this? In [node](https://github.com/firebase/firebase-admin-node/blob/bcae57ae8b919cfa630f190d9de28f91066a6c0d/src/data-connect/data-connect-api.ts#L98) we use this type:

```
/**
 * Type representing the partial claims of a Firebase Auth token used to evaluate the
 * Data Connect auth policy.
 */
export type AuthClaims = Partial<DecodedIdToken>;
```

firebase_admin/dataconnect.py
```diff
@@ -18,14 +18,39 @@
...
+    '/services/{service_id}:{endpoint_id}'
+)
+
+_EMULATOR_SERVICES_URL_FORMAT = (
+    'http://{host}/{version}/projects/{project_id}/locations/{location_id}'
+    '/services/{service_id}:{endpoint_id}'
+)
+
+# Generic Type Parameters
+_T = TypeVar("_T")
```
stephenarosaj: let's make these type names more descriptive. how about `_Data` and `_Variables`?

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+
+@dataclass
+class ExecuteGraphqlResponse(Generic[_T]):
+    data: _T
+
+
+def _get_emulator_host() -> Optional[str]:
+    emulator_host = os.environ.get("DATA_CONNECT_EMULATOR_HOST")
+    if emulator_host:
+        if "//" in emulator_host:
```
stephenarosaj: we don't make this check in the admin SDK - but maybe we should :P

if we're going to enforce that it's not using `"//"`, could we just use a regex to ensure it's not doing anything else we don't expect? i believe we expect either `localhost:<port>` or `<IP address>:<port>`

stephenarosaj (changes requested):
Partial review - just dropping comments early

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+
+        self._emulator_host = _get_emulator_host()
+        if self._emulator_host:
+            self._credential = _utils.EmulatorAdminCredentials()
+        else:
+            self._credential = app.credential.get_credential()
+
+        self._http_client = _http_client.JsonHttpClient(credential=self._credential)
+
+    def _validate_inputs(
```
stephenarosaj: this function does validate inputs - but only for `execute_graphql()` since it expects `graphql_options` - but in the future there will be other methods whose inputs need to be validated that don't use `graphql_options` - so the name `_validate_inputs` will probably have to be more specific

thinking a little deeper on this, it seems the majority of this function deals with validating `graphql_options` - and string validation is only a few lines - perhaps instead we should make this just `_validate_graphql_options`, and leave the string validation in-line in the funciton. WDYT?

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+        self,
+        graphql_query: str,
+        graphql_options: Optional[GraphqlOptions[Any]],
+        variable_type: Any = None
+    ) -> None:
+        """Validates query and GraphqlOptions inputs at runtime."""
+        # Validate the Query
+        if not isinstance(graphql_query, str) or not graphql_query.strip():
```
stephenarosaj: seems like the `or not graphql_queryl.strip()` is checking if it's an empty string - but it would be simpler to just check `or not graphql_query`

if your goal was to also remove the leading/trailing whitespace, heads up that `.strip()` doesn't remove that in place, it returns the edited string ([documentation](https://docs.python.org/3.4/library/stdtypes.html#str.strip))

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+            # Validate Variables against expected variable_type
+            variables = graphql_options.variables
+            if variables is not None and variable_type is not None:
+                if not isinstance(variables, variable_type):
+                    raise ValueError(f"variables must be of type {variable_type.__name__}")
+
+            # Validate Operation Name (if it exists)
+            operation_name = graphql_options.operation_name
+            if operation_name is not None:
+                if not isinstance(operation_name, str) or not operation_name.strip():
```
stephenarosaj: see above comment about `.strip()`

firebase_admin/dataconnect.py
stephenarosaj: Actually, looking at the API proposal, we said `T_Variables` and `T_Returned_Data` - but let's just use `T_Data` instead.

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+
+        # Validate Options (if they exist)
+        if graphql_options is not None:
+            if not isinstance(graphql_options, GraphqlOptions):
+                raise ValueError('options must be a GraphqlOptions instance')
+
+            # Validate Variables against expected variable_type
+            variables = graphql_options.variables
+            if variables is not None and variable_type is not None:
+                if not isinstance(variables, variable_type):
```
stephenarosaj: `isinstsance()` can only be used to check that an object is an instance of a class ([documentation](https://docs.python.org/3/library/functions.html#isinstance))

Is it possible that a user passes in a type that's not a class or that will otherwise cause a false / throw an error in `isinstance()` even it really is the desired type?? after doing some research, it seems so:

- [PEP 484](https://peps.python.org/pep-0484/#instantiating-generic-classes-and-type-erasure) talks about runtime type erasure for generics
- there's also [reddit](https://www.reddit.com/r/learnpython/comments/1dz56ep/isinstancex_listlistint_doesnt_work/) and [stackoverflow](https://stackoverflow.com/questions/78703537/why-does-python-not-allow-generics-with-isinstance-checks) posts about for example it not working with `list[<some type>]` or `Dict[str, any]`

There's a bunch of solutions floating around on the internet / with AI, but I propose we ask Lahiru and/or Denver for advice

stephenarosaj (changes requested):
Further partial review

firebase_admin/dataconnect.py
mk2023: That's a good point! After taking a look at `AuthClaims` and `DecodedIdToken`, I used `typing.TypedDict` with `total=False` to mirror `Partial<DecodedIdToken>`, along with a nested `FirebaseClaim` type structure. Additionally, I annotated `Impersonation.authenticated` with `Union[AuthClaims, Dict[str, Any]]` so that type checkers provide full auto-completion for standard claims while still supporting arbitrary custom claim dictionaries:

class FirebaseClaim(TypedDict, total=False):
    """Provider-specific identity details and sign-in info within a decoded ID token."""
    identities: Dict[str, Any]
    sign_in_provider: str
    sign_in_second_factor: str
    second_factor_identifier: str
    tenant: str


class AuthClaims(TypedDict, total=False):
    """Partial claims of a Firebase Auth token used for Data Connect policy evaluation."""
    aud: str
    auth_time: int
    email: str
    email_verified: bool
    exp: int
    firebase: FirebaseClaim
    iat: int
    iss: str
    phone_number: str
    picture: str
    sub: str
    uid: str


class Impersonation:
    """Represents impersonation configuration for DataConnect requests."""

    @staticmethod
    def unauthenticated() -> Dict[str, bool]:
        """Returns impersonation configuration for unauthenticated requests."""
        return {"unauthenticated": True}

    @staticmethod
    def authenticated(auth_claims: Union[AuthClaims, Dict[str, Any]]) -> Dict[str, Any]:
        """Returns impersonation configuration for authenticated requests."""
        return {"authClaims": auth_claims}`

firebase_admin/dataconnect.py
mk2023: Unfortunately, if I change it to T_Data and T_Variables, the linter says: 
firebase_admin/dataconnect.py:54:0: C0103: Type variable name "T_Variables" doesn't conform to predefined naming style (invalid-name)

Should I leave it as _Data and _Variables?

firebase_admin/dataconnect.py
mk2023: That's a great point I should have considered. One thing I'm worried about though is in `firebase_admin/functions.py`, they also use the same `_get_emulator_host()` so should I make a custom implementation specifically for `dataconnect.py,` or just drop it entirely to avoid the inconsistency?

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+    Attributes:
+        connector_config: The connector configuration specifying the service,
+            location, and connector name.
+        app: The Firebase App instance associated with this client.
+    """
+
+    def __init__(self, connector_config: ConnectorConfig, app: App) -> None:
+        if not isinstance(app, App):
+            raise ValueError(
+                'First argument passed to DataConnectApiClient must be a valid '
```
itsrakhil: App looks like it is actually the second argument here (since connector_config is the first). We should probably update this to say 'The app argument passed to...' or 'Second argument...' to avoid confusing users who encounter this.

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+            version=_API_VERSION,
+            project_id=project_id,
+            location_id=location,
+            service_id=service_id,
+            endpoint_id=method_name
+        )
+
+    def _get_headers(self) -> Dict[str, str]:
+        """Build and return the headers for a Firebase Data Connect API call."""
+        return{
```
itsrakhil: Nit: Looks like we're missing a space after return?

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+    def authenticated(auth_claims: Dict[str, Any]) -> Dict[str, Any]:
+        """Returns impersonation configuration for authenticated requests."""
+        return {"authClaims": auth_claims}
+
+
+@dataclass
+class GraphqlOptions(Generic[_V]):
+    variables: Optional[_V] = None
+    operation_name: Optional[str] = None
+    impersonate: Optional[Impersonation] = None
```
itsrakhil: Looks like there is a type mismatch here. `GraphqlOptions.impersonate` expects an instance of Impersonation, but Impersonation.unauthenticated() and authenticated() both return dictionaries. This will probably cause type checkers like mypy to complain if users try to do impersonate=Impersonation.unauthenticated(). We might want to either make Impersonation an actual instantiable class that wraps the dictionary, or change the type hint in GraphqlOptions to expect a Dict or a TypeAlias.

firebase_admin/dataconnect.py
itsrakhil: If `functions.py` has this identical implementation, let's extract this out into a shared utility function inside firebase_admin/_utils.py. You could make it something like `def get_emulator_host(env_var_name: str) -> Optional[str]:` and then reuse it in both functions.py and dataconnect.py.

itsrakhil (commented):
Left some comments.

firebase_admin/dataconnect.py
mk2023: Great idea! That keeps the function focused and makes it much more reusable for future methods. I’ll rename it to `_validate_graphql_options` and move the query string validation back inline within `execute_graphql`.

firebase_admin/dataconnect.py
mk2023: Oh, that's a good point! I still do want to be able to catch strings with multiple blank spaces, so I added a step where we save the new edited `.strip()` string and check whether that is empty. 
I'll also keep this in consideration when I check the query specifically in `executegraphql`. As of right now, I am doing this when I check the `operation_name` in the newly renamed `_validate_graphql_options.`

firebase_admin/dataconnect.py
mk2023: Done, and addressed!

firebase_admin/dataconnect.py
mk2023: That's a great point, my mistake! I changed it to: 
`
 if not isinstance(app, App):
            raise ValueError(
                'Second argument passed to DataConnectApiClient must be a valid '
                'Firebase app instance.'
            )
`

firebase_admin/dataconnect.py
mk2023: I added the space! Sorry about that!

firebase_admin/dataconnect.py
mk2023: That's a great point! I made the following changes:

1. `Impersonation` class inherits from `dict` +  its static methods to return `Impersonation` instances. This satisfies static type checkers while maintaining full dictionary compatibility at runtime.

2. Updated the type hint in `GraphqlOptions` to `Optional[Union[Impersonation, Dict[str, Any]]] = None` ==> this would not only allow developers construct `impersonate` using the built-in method `Impersonation.unauthenticated()`, but also through using a raw dictionary (e.g. `{"unauthenticated": True}`).

Since `Impersonation` is now a `dict` subclass, the existing `isinstance(impersonate, dict)` runtime validation in `_validate_graphql_options` remains perfectly valid.

While you suggested either an instantiable class or a Dict/TypeAlias, I chose the class approach. Relying solely on dictionaries degrades IDE discoverability and autocomplete. This hybrid approach gives us the best of both!

firebase_admin/dataconnect.py
mk2023: Great catch! You're completely right; passing something like list[User] or Dict[str, Any] will crash` isinstance()` with a TypeError because of type erasure in Python. I've been seeing a bunch of solutions as well, like utilizing `typing.get_origin()` to extract the base class from a generic type, and then using `typing.get_args()`to complete a deep-check of the inner types. I think it's a great idea to ask Lahiru and/or Denver for advice on how to go about this matter.

firebase_admin/dataconnect.py
stephenarosaj: SGTM - let's update the API proposal to match

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+                if not isinstance(variables, variable_type):
+                    raise ValueError(f"variables must be of type {variable_type.__name__}")
+
+            # Validate Operation Name (if it exists)
+            operation_name = graphql_options.operation_name
+            if operation_name is not None:
+                if not isinstance(operation_name, str) or not operation_name.strip():
+                    raise ValueError('operation_name must be a non-empty string')
+
+            # Validate Impersonation (if it exists)
```
stephenarosaj, line 277: Similar to other comment about pulling out the validation logic - validating impersonation options will probably be needed by other functions which don't have things like `query_str` or even the same `GraphqlOptions` - for example, `ImpersonateAuthenticated` is used in the Node SDK's arbitrary GraphQL API as well as our deployed operation APIs

Let's pull out the impersonation validation logic to a standalone helper function so that it can be reused where needed

firebase_admin/dataconnect.py
```diff
@@ -122,3 +147,178 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+                    if not isinstance(impersonate['unauthenticated'], bool):
+                        raise ValueError("'unauthenticated' claim must be a boolean")
+                if 'authClaims' in impersonate:
+                    if not isinstance(impersonate['authClaims'], dict):
+                        raise ValueError("'authClaims' claim must be a dictionary")
+
+    def _prepare_graphql_payload(
+        self,
+        graphql_query: str,
+        graphql_options: Optional[GraphqlOptions[Any]]
```
stephenarosaj: is it `GraphqlOptions[Any]`? shouldn't it be `GraphqlOptions[_Variables]`?

stephenarosaj (changes requested):
partial review of main file - now for tests

firebase_admin/dataconnect.py
mk2023: Done!

firebase_admin/dataconnect.py
mk2023: Separated `_validate_graphql_options` to `_validate_variables_type` and `_validate_impersonation_options` and called both within `_validate_graphql_options`!

firebase_admin/dataconnect.py
mk2023: That's a great idea! I extracted the duplicated emulator host extraction functions from `dataconnect.py` and `functions.py` into a single, shared helper function named `get_emulator_host` in `_utils.py` This new helper function also addresses Rosa's comment and uses a regex expression for checking.

firebase_admin/dataconnect.py
mk2023: I don't think we actually need `_Variables` here. Since `_prepare_graphql_payload` just serializes the options to a standard `Dict[str, Any]` and doesn't propagate the type to the return type, the generic type variable doesn't have an effect. I'm totally happy to switch it back if you'd prefer, though!

firebase_admin/dataconnect.py
stephenarosaj: **Note**: this is a point of [active discussion](https://docs.google.com/document/d/144eeuSdj-Q0OxyHtDAqf2k1uHJikOA0IkrNppNz-FGo/edit?resourcekey=0-VXQH1ZGHZciJol_tfFjJvw&tab=t.uyfs0h262in6#heading=h.980r5dpbj09j) and there may be some API changes - for now, let's continue with the rest of the PR and project, and we'll make adjustments later as needed

firebase_admin/dataconnect.py
```diff
@@ -122,3 +145,183 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+        else:
+            self._credential = app.credential.get_credential()
+
+        self._http_client = _http_client.JsonHttpClient(credential=self._credential)
+
+    def _validate_variables_type(self, variables: Any, variable_type: Any) -> None:
+        """Validates variables against expected type."""
+        if variables is not None and variable_type is not None:
+            if not isinstance(variables, variable_type):
+                raise ValueError(f"variables must be of type {variable_type.__name__}")
```
itsrakhil: If a user passes something like a tuple of types (e.g., (list, dict)) or certain generics from the typing module, Python will crash with an AttributeError when trying to construct this string, hiding the actual ValueError.
To make the error reporting completely safe, we can use getattr with a fallback to str():
`raise ValueError(f"variables must be of type {getattr(variable_type, '__name__', str(variable_type))}")`

tests/test_data_connect.py
```diff
@@ -331,3 +333,347 @@ def test_overall_client_retrieval_and_caching(self):
...
+    def teardown_method(self, method):
+        del method
+        testutils.cleanup_apps()
+
+    def test_validate_graphql_options_valid(self):
+        # Valid with no options
+        self.api_client._validate_graphql_options(None)
+
+        # Valid with options
+        options = dataconnect.GraphqlOptions(variables={"foo": "bar"})
```
stephenarosaj: seems like you're testing with variables on this line, which is a test case down below - so let's remove this line

perhaps instead, we could add a line to this test verifying that `GraphqlOptions` with no constructor arguments works fine too?

tests/test_data_connect.py
```diff
@@ -331,3 +333,347 @@ def test_overall_client_retrieval_and_caching(self):
...
+
+    def test_validate_graphql_options_valid_variables(self):
+        @dataclass
+        class User:
+            user_id: str
+            name: str
+            address: str
+
+        @dataclass
+        class UsersResponse:
```
stephenarosaj: nit: this is a `Response` type and matches the shape of data returned from the server for some `Users` table query - but we are checking for it's validity when used as variables. it's not a big deal, but we should use realistic data for the tests - can you make the type more like real variables?

tests/test_data_connect.py
```diff
@@ -331,3 +333,347 @@ def test_overall_client_retrieval_and_caching(self):
...
+
+    def test_validate_graphql_options_invalid_variables(self):
+        @dataclass
+        class User:
+            user_id: str
+            name: str
+            address: str
+
+        @dataclass
+        class UsersResponse:
```
stephenarosaj: same variables / response comment as above

tests/test_data_connect.py
```diff
@@ -331,3 +333,347 @@ def test_overall_client_retrieval_and_caching(self):
...
+        }
+
+    def test_prepare_graphql_payload_with_dataclass_variables(self):
+        @dataclass
+        class User:
+            user_id: str
+            name: str
+            address: str
+
+        @dataclass
```
stephenarosaj, line 575: same variables / response comment as above

tests/test_data_connect.py
```diff
@@ -331,3 +333,347 @@ def test_overall_client_retrieval_and_caching(self):
...
+
+    def test_prepare_graphql_payload_with_all_fields(self):
+        @dataclass
+        class User:
+            user_id: str
+            name: str
+            address: str
+
+        @dataclass
+        class UsersResponse:
```
stephenarosaj: same variables / response comment as above

stephenarosaj (changes requested):
some minor changes but i think this looks good!

tests/test_data_connect.py
mk2023: Yes, sounds good! Made that change!

firebase_admin/dataconnect.py
stephenarosaj: re: our discussion at lunch today - let's update the code to the following:
- the generic variables type should be limited to being either `collections.abc.Mapping` or a `dataclass`
- `_validate_variables_type()` should validate that if the variables are not `None`, they must either be a `collections.abc.Mapping` or a `dataclass`
- `variables_type` should be type-annotated as an optional `dataclass` since we can't type-check a `TypedDict` (we may end up removing `variables_type` after further design - for now let's keep it in to follow the API proposal, and because it's easier to remove than to add :P)

tests/test_data_connect.py
mk2023: That sounds more reasonable! I changed it to:
@dataclass
        class UserProfile:
            address: str
            phone: str

        @dataclass
        class CreateUserVariables:
            user_id: str
            name: str
            profile: UserProfile

firebase_admin/dataconnect.py
mk2023: Sounds good, I updated `_validate_variables_type()` and `_validate_graphql_options()`!

firebase_admin/dataconnect.py
mk2023: Unlike the Node.js Admin SDK, which exports DecodedIdToken from auth/token-verifier.ts, the Python Admin SDK does not define or export any typed dictionary or class representation for verified ID tokens (the auth module returns raw dicts). Thus, for right now, we just left `AuthClaims`as a raw dictionary.

firebase_admin/dataconnect.py
mk2023: That's a great point--just changed it!

firebase_admin/dataconnect.py
stephenarosaj: If there's no reason _not_  to use `_Variables`, let's do it, since that's descriptive of what the type actually is - we use it everywhere else too

mk2023: Okay, sounds good, I changed it! Thanks!

firebase_admin/dataconnect.py
stephenarosaj: SGTM - if we want to in the future we can more strongly type it, but let's just match what Auth does - we should not try to define their API

firebase_admin/dataconnect.py
stephenarosaj: SGTM - leaving this comment open for Lahiru to see when he reviews

stephenarosaj (approved):
LGTM - ready for Lahiru

firebase_admin/dataconnect.py
lahirumaramba: Using `typing.get_origin(variable_type) or variable_type` to unwrap the generics sounds good to me!
I think this unwraps subscripted generics (`Dict[str, Any] -> dict`, `Mapping[str, Any] -> collections.abc.Mapping`) and leave the plain dataclass classes untouched.

For inner type validation (`get_args`), WDYT about keeping runtime validation focused on top-level structure and leave deep type-checking to static type checkers (mypy/pyright) to avoid unnecessary runtime overhead?

firebase_admin/_utils.py
```diff
@@ -345,3 +347,16 @@ def __init__(self):
 
     def refresh(self, request):
         pass
+
+
+def get_emulator_host(env_var_name: str) -> Optional[str]:
+    """Retrieves and validates the host from the specified emulator environment variable."""
+    emulator_host = os.environ.get(env_var_name)
+    if emulator_host:
+        if not re.match(r'^(?:\[[a-fA-F0-9:]+\]|[a-zA-Z0-9.-]+):[0-9]+$', emulator_host):
```
lahirumaramba: This regex rejects hostnames with underscores (_). I am not sure how frequently hostnames with `_` are used but think that's common in Docker containers. Consider changing it to: `if not re.match(r'^(?:\[[a-fA-F0-9:]+\]|[a-zA-Z0-9._-]+):[0-9]+$', emulator_host):`

This will also make sure the new check is non-breaking. The existing check only look for `//`, so we need to make sure the new check is non-breaking for existing users.

firebase_admin/dataconnect.py
```diff
@@ -122,3 +146,191 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
...
+
+            # Validate Variables against expected variable_type
+            self._validate_variables_type(graphql_options.variables, variable_type)
+
+            # Validate Operation Name (if it exists)
+            operation_name = graphql_options.operation_name
+            if operation_name is not None:
+                if not isinstance(operation_name, str):
+                    raise ValueError('operation_name must be a string')
+                operation_name = operation_name.strip()
```
lahirumaramba: Probably harmless but this modifies the user provided `GraphqlOptions` object. If that is intended, leave it as it is. Otherwise, change it to `
```py
    if not operation_name.strip():
        raise ValueError('operation_name must be a non-empty string')
```

and apply `.strip()` inside `_prepare_graphql_payload()`
```py
    payload["operationName"] = graphql_options.operation_name.strip()
```

This will prevent any potential side effects of modifying user input.

firebase_admin/_utils.py
```diff
@@ -345,3 +347,16 @@ def __init__(self):
 
     def refresh(self, request):
         pass
+
+
+def get_emulator_host(env_var_name: str) -> Optional[str]:
```
lahirumaramba, line 352: Can we add unit tests to cover this new helper function?

lahirumaramba (changes requested):
Thank you! Overall LGTM!
Added a few comments/questions.

firebase_admin/_utils.py
mk2023: Yes, that's a great point! I changed it!

firebase_admin/_utils.py
mk2023: Yes, that's a great idea! I will create a new file called `test_utils.py` in the `tests` folder!

firebase_admin/dataconnect.py
mk2023: That sounds good! I changed both `_validate_graphql_options` and `_prepare_graphql_payload()`!

firebase_admin/dataconnect.py
mk2023: Yes, I agree with that. I changed `_validate_variables_type` to first check whether it's a mapping or a dataclass, and then unwrapped the subscripted generics to validate against the passed variables. Like you mentioned, I didn't run a deep check on the inner types!

lahirumaramba (approved):
LGTM! Thanks!

PR #969, opened 2026-07-20
feat(fdc): Add request execution and response parsing to _DataConnectApiClient

mk2023:
## Overview
✨ Implements internal GraphQL HTTP request execution (`_make_gql_request`), response parsing (`_parse_graphql_response`), error checking (`_check_graphql_errors`), and introduces `QueryError` exception handling for Firebase Data Connect.

## Highlights
✨ Key changes:
* **Request Execution (`_DataConnectApiClient._make_gql_request`)**: Added low-level HTTP POST request helper using `JsonHttpClient.body_and_response` to handle Data Connect GraphQL API calls.
* **Error Handling (`QueryError` & `_check_graphql_errors`)**: Introduced `QueryError` (subclass of `FirebaseError`) to represent Data Connect query/mutation errors and added error formatting for non-empty GraphQL error payloads.
* **Response Parsing (`_parse_graphql_response`)**: Added `ExecuteGraphqlResponse` parsing to return raw JSON response dictionaries, aligning Data Connect with Firestore and Realtime Database patterns.
* **Client Instantiation (`DataConnect`)**: Updated `DataConnect.__init__` to immediately instantiate `_DataConnectApiClient` upon creation.
* **Unit Tests (`tests/test_data_connect.py`)**: Added unit tests covering HTTP request execution, error parsing, non-dictionary response handling, and immediate client instantiation.

<details>

### Core SDK Changes
  - Implemented `_make_gql_request` for executing HTTP POST requests.
  - Implemented `_check_graphql_errors` to parse and format backend GraphQL errors into `QueryError`.
  - Implemented `_parse_graphql_response` to extract `data` into `ExecuteGraphqlResponse`.
  - Added `QueryError` exception class and exported it in `__all__`.
  - Updated `DataConnect.__init__` to instantiate `_DataConnectApiClient`.

### Unit Test Changes
- **`tests/test_data_connect.py`**:
  - Added unit test cases for `_make_gql_request` (success, missing parameters, HTTP errors, server GraphQL errors).
  - Added unit test cases for `_parse_graphql_response` (dictionary data, None data, malformed payloads).
  - Added unit test cases for `QueryError` formatting and property assignment.

</details>

---

### Testing Strategy
- **Unit Tests**: `pytest tests/test_data_connect.py` (PASSED)

---

Context Sources Used:
- id: firebase-admin-python

### DIFF
--- a/firebase_admin/dataconnect.py
+++ b/firebase_admin/dataconnect.py
@@ -22,8 +22,11 @@ from collections.abc import Mapping
 from dataclasses import dataclass, asdict, is_dataclass
 import typing
 from typing import Any, Dict, Generic, Optional, Type, TypeVar, Union
+
+import requests
+
 import firebase_admin
-from firebase_admin import _utils, _http_client, App
+from firebase_admin import _utils, _http_client, App, exceptions
 
 __all__ = [
     'ConnectorConfig',
@@ -334,3 +337,39 @@ class _DataConnectApiClient:
             "X-Firebase-Client": f"fire-admin-python/{firebase_admin.__version__}",
             "x-goog-api-client": _utils.get_metrics_header(),
         }
+
+    def _make_gql_request(
+        self,
+        url: str,
+        headers: Dict[str, str],
+        payload: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        """Make a GraphQL request to the Data Connect service."""
+        if url is None or headers is None or payload is None:
+            raise ValueError("url, headers, and payload must all be specified.")
+
+        try:
+            resp_dict, resp = self._http_client.body_and_response(
+                'post',
+                url=url,
+                headers=headers,
+                json=payload
+            )
+        except requests.exceptions.RequestException as error:
+            raise _utils.handle_platform_error_from_requests(error)
+
+        if resp_dict and "errors" in resp_dict:
+            errors = resp_dict["errors"]
+            if isinstance(errors, list) and len(errors) > 0:
+                messages = [
+                    err.get("message") for err in errors
+                    if isinstance(err, dict) and err.get("message")
+                ]
+                all_messages = " ".join(messages)
+                raise exceptions.FirebaseError(
+                    code="query-error",
+                    message=all_messages,
+                    http_response=resp
+                )
+
+        return resp_dict
--- a/firebase_admin/dataconnect.py
+++ b/firebase_admin/dataconnect.py
@@ -20,18 +20,14 @@ Firebase apps.
 
 from collections.abc import Mapping
 from dataclasses import dataclass, asdict, is_dataclass
-import enum
 import typing
 from typing import Any, Dict, Generic, Optional, Type, TypeVar, Union
 
+
 import requests
 
-try:
-    from types import UnionType  # pylint: disable=no-name-in-module
-except ImportError:
-    UnionType = None
-
 import firebase_admin
+
 from firebase_admin import _utils, _http_client, App, exceptions
 
 __all__ = [
@@ -121,6 +117,7 @@ class DataConnect:
         """Initializes a DataConnect client instance. """
         self._app: App = app
         self._config = config
+        self._client = _DataConnectApiClient(connector_config=config, app=app)
 
     @property
     def app(self) -> App:
@@ -172,7 +169,6 @@ def client(config: ConnectorConfig, app: Optional[App] = None) -> DataConnect:
     dc_service = _utils.get_app_service(app, _DATA_CONNECT_ATTRIBUTE, _DataConnectService)
 
     return dc_service.get_client(config)
-
 
 
 class Impersonation(dict):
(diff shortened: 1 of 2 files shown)


firebase_admin/dataconnect.py
```diff
@@ -334,3 +337,39 @@ def _get_headers(self) -> Dict[str, str]:
...
+                messages = [
+                    err.get("message") for err in errors
+                    if isinstance(err, dict) and err.get("message")
+                ]
+                all_messages = " ".join(messages)
+                raise exceptions.FirebaseError(
+                    code="query-error",
+                    message=all_messages,
+                    http_response=resp
+                )
```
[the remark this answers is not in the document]
mk2023: That makes sense! To guarantees that an exception will be raised when "errors" is present in the response dictionary, I changed it to:
` 
 if isinstance(resp_dict, dict) and "errors" in resp_dict:
            errors = resp_dict["errors"]
            all_messages = ""
            if isinstance(errors, list):
                messages = [
                    err.get("message") for err in errors
                    if isinstance(err, dict) and err.get("message")
                ]
                all_messages = " ".join(messages)
            if not all_messages:
                all_messages = (
                    f"GraphQL execution failed: {errors}" if errors
                    else "GraphQL execution failed."
                )
            raise exceptions.FirebaseError(
                code="query-error",
                message=all_messages,
                http_response=resp
            )
`

tests/test_data_connect.py
```diff
@@ -724,3 +724,59 @@ def test_get_headers(self):
...
+
+    @mock.patch.object(_http_client.JsonHttpClient, "body_and_response")
+    def test_make_gql_request_error(self, mock_body_and_response):
+        mock_body_and_response.side_effect = requests.exceptions.RequestException()
+        url = "https://example.com/endpoint"
+        headers = {"key": "val"}
+        payload = {"query": "foo"}
+
+        with pytest.raises(exceptions.FirebaseError):
+            self.api_client._make_gql_request(url, headers, payload)
```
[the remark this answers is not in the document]
mk2023: Added!

firebase_admin/dataconnect.py
```diff
@@ -20,10 +20,19 @@
...
+import enum
 import typing
 from typing import Any, Dict, Generic, Optional, Type, TypeVar, Union
+
+import requests
+
+try:
+    from types import UnionType  # pylint: disable=no-name-in-module
+except ImportError:
+    UnionType = None
```
stephenarosaj: My first reaction is to be hesitant about this pattern. It looks like `UnionType` was only introduced in Python 3.10 - if we're running on 3.9, what happens? Does our type checking break?

Is there a way to support legacy and modern union types all at once with something like `typing.Union`?

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+                json=payload
+            )
+        except requests.exceptions.RequestException as error:
+            raise _utils.handle_platform_error_from_requests(error)
+
+        if isinstance(resp_dict, dict) and "errors" in resp_dict:
+            errors = resp_dict["errors"]
+            all_messages = ""
+            if isinstance(errors, list):
+                messages = [
```
stephenarosaj: i feel the definition of this variable is a bit hard to reason about. could we unravel the inline definition to make it easier to understand how it's defined?

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+                    if isinstance(err, dict) and err.get("message")
+                ]
+                all_messages = " ".join(messages)
+            if not all_messages:
+                all_messages = (
+                    f"GraphQL execution failed: {errors}" if errors
+                    else "GraphQL execution failed."
+                )
+            raise exceptions.FirebaseError(
+                code="query-error",
```
stephenarosaj: let's define these codes as an enum so that there's no hard-coded values for them. that way we don't have to update every hardcoded use in the future if a change is needed

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+                message=all_messages,
+                http_response=resp
+            )
+
+        return resp_dict
+
+    @staticmethod
+    def _extract_actual_type(field_type: Any) -> Any:
+        origin = typing.get_origin(field_type)
+        if origin is Union or (UnionType is not None and origin is UnionType):
```
stephenarosaj: Re: [this comment](https://github.com/firebase/firebase-admin-python/pull/969/changes#r3624607531) - this line of code seems to imply that something can be either a `Union` or a `UnionType` and therefore we have to check for both of them - is that true? If not, I suggest we just use the backwards compatible type to keep things simple

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+            raise exceptions.FirebaseError(
+                code="query-error",
+                message=all_messages,
+                http_response=resp
+            )
+
+        return resp_dict
+
+    @staticmethod
+    def _extract_actual_type(field_type: Any) -> Any:
```
stephenarosaj: My understanding of this method is that, for example if we have a type union `X | None`, it should return `X` - but what about when we have a type union `X | Y | None`? it will just end up returning that whole union right? Is that intentional? 

I don't think the multi-member union is even possible for FDC's schema type semantics - but this is user input, so I think it makes sense to guard for this

Also, I think the name `_extract_actual_type` may not be descriptive / accurate for this behavior - how about something like `_reduce_optiona_type`?

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+        if data is None or type_hint is Any:
+            return data
+
+        actual_type = _DataConnectApiClient._extract_actual_type(type_hint)
+
+        origin = typing.get_origin(actual_type)
+        args = typing.get_args(actual_type)
+
+        # Union (e.g. Union[int, str] or Union[int, List[str]])
+        if origin is Union or (UnionType is not None and origin is UnionType):
```
stephenarosaj: can you help me understand this code? i get a little lost in the innermost parts
(can chat synchronously offline about it)

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+                return non_none_args[0]
+        return field_type
+
+    @staticmethod
+    def _deserialize_type(type_hint: Any, data: Any) -> Any:
+        """Recursively deserializes data into the specified type hint."""
+        if data is None or type_hint is Any:
+            return data
+
+        actual_type = _DataConnectApiClient._extract_actual_type(type_hint)
```
stephenarosaj: Re: [this comment](https://github.com/firebase/firebase-admin-python/pull/969/changes#r3624739637) - could we change this variable name to `optional_type`?

firebase_admin/dataconnect.py
```diff
@@ -334,3 +343,161 @@ def _get_headers(self) -> Dict[str, str]:
...
+        origin = typing.get_origin(field_type)
+        if origin is Union or (UnionType is not None and origin is UnionType):
+            args = typing.get_args(field_type)
+            non_none_args = [arg for arg in args if arg is not type(None)]
+            if len(non_none_args) == 1:
+                return non_none_args[0]
+        return field_type
+
+    @staticmethod
+    def _deserialize_type(type_hint: Any, data: Any) -> Any:
```
stephenarosaj: great job with this function, it makes me appreciate how complex type deserialization is in python lol

QQ: should we just use a 3rd party library for this (something like the [`TypeAdapter`](https://pydantic.dev/docs/validation/latest/concepts/type_adapter/) from [pydantic](https://pydantic.dev/docs/validation/latest/get-started/))? i think that it would be preferred that we don't have to maintain / verify this complexity ourselves (especially considering in the future we're going to add generated types). Letting a third party library handle it would be ideal IMO

i don't want to bloat the other packages with another library, but thankfully it seems with `setup.py` we can actually do `extras_require` so that this is only required for FDC and not other packages -WDYT @lahirumaramba ?

stephenarosaj (changes requested):
Some comments - haven't looked at tests yet

firebase_admin/dataconnect.py
stephenarosaj: Discussed offline!

firebase_admin/dataconnect.py
stephenarosaj: Discussed offline - returning `X | Y | Z` is intentional and it's used properly in the code down below. But we should probably add some comments / documentation to this function to make it clearer

firebase_admin/dataconnect.py
mk2023: That's a good point! It's important to write code in a way that is easily understandable for everyone! I replaced the inline comprehension to an explicit loop!

firebase_admin/dataconnect.py
mk2023: That is a great point, and I completely agree with the reasoning for avoiding hard-coded error strings across the codebase. Looking deeper into the SDK's existing patterns, other service modules (such as `db.py` with `TransactionAbortedError`) define dedicated error classes by subclassing `FirebaseError` rather than using an `Enum`. Thus, what do you think about doing this for dataconnect as well:

class QueryError(exceptions.FirebaseError):
    """Raised when a GraphQL query or mutation execution fails."""

    def __init__(self, message: str, http_response: Any = None) -> None:
        super().__init__(
            code=_QUERY_ERROR_CODE,
            message=message,
            http_response=http_response
        )

and call it through:

`raise QueryError(
                message=all_messages,
                http_response=resp
            )`

Let me know if you want to stick to the `Enum` option!

firebase_admin/dataconnect.py
mk2023: Great catch! However, we decided to pivot and drop the internal deserialization completely to return a raw dict instead. This code is removed!

firebase_admin/dataconnect.py
mk2023: We decided to pivot and drop the internal deserialization completely to return a raw dict instead. This code is removed!

firebase_admin/dataconnect.py
mk2023: We decided to pivot and drop the internal deserialization completely to return a raw dict instead. This code is removed!

firebase_admin/dataconnect.py
mk2023: Decided on removing the internal deserialization logic and the `data_type` parameter from the `execute_graphql` API after in-depth research and multiple discussions. Because Python lacks built-in deserialization, returning a raw dictionary prevents forcing developers into specific third-party dependencies. Instead, they will be able to easily drop data into their preferred libraries. This also aligns with how other Firebase Python SDK services with user-defined schemas, like Realtime Database and Cloud Firestore.

itsrakhil (approved):
lgtm

firebase_admin/dataconnect.py
```diff
@@ -334,3 +355,62 @@ def _get_headers(self) -> Dict[str, str]:
...
+    def _parse_graphql_response(
+        resp_dict: Dict[str, Any]
+    ) -> ExecuteGraphqlResponse[Any]:
+        """Parses a raw GraphQL response payload into ExecuteGraphqlResponse."""
+        if not isinstance(resp_dict, dict):
+            raise exceptions.InternalError(
+                message="Response payload is not a valid JSON dictionary."
+            )
+
+        return ExecuteGraphqlResponse(data=resp_dict.get("data"))
```
stephenarosaj, line 423: side note: can you add a `TODO` comment here, and in the class definition of `ExecuteGraphqlResponse` mentioning that we need to add support for "partial errors"? Link to this bug: [b/406281627](https://b/406281627)

here's a `TODO` style guide: [go/todo-style](https://go/todo-style)

stephenarosaj (approved):
LGTM! the simplification of removing deserialization is a gr8 :P

firebase_admin/dataconnect.py
mk2023: Done!

firebase_admin/dataconnect.py
```diff
@@ -334,3 +355,62 @@ def _get_headers(self) -> Dict[str, str]:
...
+
+    @staticmethod
+    def _parse_graphql_response(
+        resp_dict: Dict[str, Any]
+    ) -> ExecuteGraphqlResponse[Any]:
+        """Parses a raw GraphQL response payload into ExecuteGraphqlResponse."""
+        if not isinstance(resp_dict, dict):
+            raise exceptions.InternalError(
+                message="Response payload is not a valid JSON dictionary."
+            )
```
jonathanedey, line 420: Would we also want to provide the response here for debug purposes?

firebase_admin/dataconnect.py
```diff
@@ -22,8 +22,13 @@
 from dataclasses import dataclass, asdict, is_dataclass
 import typing
 from typing import Any, Dict, Generic, Optional, Type, TypeVar, Union
+
```
jonathanedey, line 25: super nit: spacing

jonathanedey (approved):
LGTM with a few comments, Thanks!

firebase_admin/dataconnect.py
mk2023: Fixed!

firebase_admin/dataconnect.py
mk2023: Yes, that's a good point! I will add this change!

PR #970, opened 2026-07-27
feat(fdc): Add execute_graphql and execute_graphql_read support with Pythonic impersonation

mk2023:
## Overview
✨ Adds public `execute_graphql` and `execute_graphql_read` methods to `DataConnect`, refactors `Impersonation` to use Pythonic `auth_claims` (PEP 8 snake_case) in Python land while translating to `authClaims` during JSON payload serialization, and includes comprehensive unit and emulator integration test suites.

## Highlights
✨ Key changes:
* **Public API (`firebase_admin.dataconnect`)**: Added `execute_graphql` for query/mutation execution and `execute_graphql_read` for read-only queries with mutation validation.
* **Pythonic Impersonation**: Updated `Impersonation` to store `auth_claims` in Python land, with automatic `auth_claims` -> `authClaims` key translation in `_prepare_graphql_payload` for network serialization.
* **Options Validation**: Updated `_validate_impersonation_options` to validate `auth_claims` in Python land.
* **Unit Tests (`tests/test_data_connect.py`)**: Added 82 unit tests covering method execution, payload serialization, dataclass variables, impersonation options, and error parsing.
* **Integration Tests & Emulator Seeding (`integration/`)**: Added 20 integration tests against the Data Connect emulator, including `seed.sh` data seeding (`fred_id`, `jeff_id`, `email_id`) and state cleanup (`UPSERT_FRED_EMAIL`).
* **CI Workflow (`.github/workflows/ci.yml`)**: Added `--cert` flag and emulator seeding step for CI integration testing.

<details>

### Core SDK Changes
  - Implemented `execute_graphql` and `execute_graphql_read` on `DataConnect`.
  - Added `Impersonation` class with `unauthenticated()` and `authenticated(auth_claims=...)`.
  - Updated `_prepare_graphql_payload` to translate `auth_claims` to `authClaims` via dictionary key popping.
  - Added `_validate_impersonation_options` and `ExecuteGraphqlResponse` dataclass.

### Integration & Emulator Changes
- **`integration/test_data_connect.py`**: Added 20 integration tests covering queries, mutations, multi-operation documents, impersonated requests, and mutation state cleanup.
- **`integration/emulators/seed.sh`**: Added script to seed initial test data (`fred_id`, `jeff_id`, `email_id`) into the emulator.
- **`.github/workflows/ci.yml`**: Added `--cert` flag and emulator seeding to integration test workflow.

### Unit Test Changes
- **`tests/test_data_connect.py`**: Added 82 unit tests, refactoring shared constants (`TEST_URL`, `TEST_HEADERS`, `TEST_PAYLOAD`, `TEST_AUTH_CLAIMS`, `TEST_VARIABLES`) and dataclasses (`UserProfile`, `CreateUserVariables`, `User`).

</details>

---

### Testing Strategy
- **Unit Tests**: `pytest tests/test_data_connect.py` (82/82 PASSED)
- **Integration Tests**: `DATA_CONNECT_EMULATOR_HOST=127.0.0.1:9399 pytest integration/test_data_connect.py --cert tests/data/service_account.json` (20/20 PASSED)

---

Context Sources Used:
- id: firebase-admin-python


tests/test_data_connect.py
```diff
@@ -881,3 +881,219 @@ def test_parse_graphql_response_non_dict_error(self):
...
+            self.api_client.execute_graphql("query { foo }", options="not-graphql-options")
+
+    def test_execute_graphql_invalid_variables_type(self):
+        @dataclass
+        class User:
+            name: str
+
+        options = dataconnect.GraphqlOptions(variables={"name": "Fred"})
+        with pytest.raises(ValueError, match="Expected variables of type User"):
+            self.api_client.execute_graphql("query { foo }", options=options, variables_type=User)
```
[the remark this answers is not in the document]
mk2023: That's a good catch! Edited the test to validate against the proper message!

firebase_admin/dataconnect.py
```diff
@@ -102,7 +105,42 @@ def __post_init__(self):
             raise ValueError("connector cannot be empty")
 
 
+class Impersonation(dict):
```
stephenarosaj, line 109: just realizing this now but IIUC, because there's no constructor defined here a default one is available for users to call directly

we should make sure that:
1) we call out in documentation it's preferred that you use the static factory methods
2) if they DO call the constructor, we should strictly enforce that they choose a single configuration - they can either provide `unauthenticated=True` or `auth_claims=auth_claims`

firebase_admin/dataconnect.py
```diff
@@ -127,6 +165,66 @@ def app(self) -> App:
...
+            ValueError: If the arguments are invalid from the local inputs side.
+            InvalidArgumentError: If GraphQL syntax validation fails on the server.
+            PermissionDeniedError: If an @auth policy directive blocks execution due to
+                insufficient permission.
+            NotFoundError: If a specified resource is not found, or the request is rejected
+                by undisclosed reasons, such as whitelisting.
+            InternalError: If the server response payload is invalid or malformed.
+            FirebaseError: The base platform exception.
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
stephenarosaj: this is intentional, it's a draft PR.

firebase_admin/dataconnect.py
```diff
@@ -127,6 +165,66 @@ def app(self) -> App:
...
```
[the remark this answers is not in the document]
stephenarosaj: intentional, draft PR

firebase_admin/dataconnect.py
```diff
@@ -414,3 +483,31 @@ def _parse_graphql_response(
...
+
+    def execute_graphql_read(
+        self,
+        query: str,
+        options: Optional[GraphqlOptions[_Variables]] = None,
+        variables_type: Type[_Variables] = Any,
+    ) -> ExecuteGraphqlResponse[Any]:
+        """Executes a read-only GraphQL query and returns the result."""
+        raise NotImplementedError
```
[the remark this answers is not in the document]
stephenarosaj: intentional, draft PR

integration/test_data_connect.py
```diff
@@ -0,0 +1,337 @@
...
+    impersonate=dataconnect.Impersonation.authenticated({
+        'sub': 'non-existing-id',
+        'email_verified': True
+    })
+)
+
+
+@pytest.fixture(autouse=True)
+def setup_teardown_db(default_app):
+    """Sets initial database state before each test and cleans up after."""
```
stephenarosaj: ah, this was a bad practice that we did in the admin node SDK - we haven't tested these functions yet, so how could we possibly use them to initialize the database?

instead, WDYT of this - we should seed the database using some already-tested thing... maybe the firebase CLI? i'll look more into this, feel free to as well and drop some ideas

stephenarosaj (changes requested):
in progress, leaving comments early - have to look at test cases still

firebase_admin/dataconnect.py
mk2023: Good catch! `_validate_impersonation_options` extensively validates raw dictionary inputs passed directly to `GraphqlOptions(impersonate={...})`, but adding validation in `Impersonation.__init__` handles direct class instantiation early.

I've updated `Impersonation`:
- Updated docstring to highlight static factory methods (`unauthenticated()` / `authenticated()`).
- Added an explicit `__init__` constructor validating `unauthenticated` vs `auth_claims` (and handling both `auth_claims` and `authClaims`).
- Added unit tests in `TestImpersonation` covering all constructor validation scenarios.

integration/test_data_connect.py
mk2023: Yes, I understand why that isn't the most honest testing method! 

If we use the Node.js SDK (or Firebase CLI) to seed the database, we would have to introduce external Node.js subprocess dependencies into `pytest`. 

Instead, what do you think about using raw `requests.post()` calls directly against the emulator URL inside `setup_teardown_db`? This way we can seed the database via raw HTTP POST requests without invoking any `firebase-admin-python` SDK methods while keeping the test suite 100% Python-native.

mk2023: After looking into this and getting advice, I removed the setup/teardown logic from pytest, created a `seed.sh` script for the initial data, and updated our GitHub Actions workflow to run the script using firebase emulators:exec right before executing the tests!

firebase_admin/dataconnect.py
stephenarosaj: think i may have missed something - why do we have `auth_claims` and `authClaims`?

firebase_admin/dataconnect.py
```diff
@@ -315,12 +353,18 @@ def _validate_variables_type(
         if variables is not None:
             if not (isinstance(variables, Mapping) or is_dataclass(variables)):
                 raise ValueError("variables must be a collections.abc.Mapping or a dataclass")
-            if variable_type is not None:
+            if (
+                variable_type is not None
+                and variable_type is not Any
+                and variable_type is not typing.Any
+            ):
```
stephenarosaj, line 353: i'm curious - why add these additonal Any checks?

firebase_admin/dataconnect.py
```diff
@@ -492,7 +539,20 @@ def _execute_graphql_helper(
         variables_type: Type[_Variables] = Any,
     ) -> ExecuteGraphqlResponse[Any]:
         """Helper method to execute GraphQL queries or mutations against a specified endpoint."""
-        raise NotImplementedError
+        if not isinstance(query, str):
+            raise ValueError("query must be a string")
+        if not query.strip():
```
stephenarosaj: suggestion: `strip()` doesn't edit the string in-place, it returns a new string - the usage is correctly here, but could we also strip the string before we use it in the rest of the function?

integration/emulators/seed.sh
```diff
@@ -0,0 +1,28 @@
...
+  -H "Content-Type: application/json" \
+  -d '{"query": "mutation { user_upsert(data: { id: \"jeff_id\", address: \"99 Oak St.\", name: \"Jeff\" }) }"}' > /dev/null
+
+# 3. Seed Fred Email
+curl -s -f -X POST "${ENDPOINT}" \
+  -H "Authorization: Bearer owner" \
+  -H "Content-Type: application/json" \
+  -d '{"query": "mutation { email_upsert(data: { id: \"email_id\", subject: \"free bitcoin inside\", date: \"1999-12-31\", text: \"get pranked! LOL!\", fromId: \"fred_id\" }) }"}' > /dev/null
+
+echo "Successfully seeded initial test data (2 users, 1 email)!"
```
stephenarosaj: i think this may fail silently for GraphQL errors. IIUC, `set -e` and using `curl -f` will fail the script if `curl` returns a non-OK status code, but if we successfully send this request and receive a proper GQL response from the backend that contains errors, it will still report `200` and the errors won't be caught/reported, and then we'll report a success on the last line

integration/test_data_connect.py
```diff
@@ -19,11 +19,12 @@
 from firebase_admin import dataconnect, exceptions
 
 CONNECTOR_CONFIG = dataconnect.ConnectorConfig(
-    location='us-west2',
+    location='us-east4',
```
stephenarosaj: why `east4`? the config you checked in uses `west2`

integration/test_data_connect.py
```diff
@@ -156,15 +145,14 @@ def test_execute_graphql_mutation(self):
...
-        delete_resp = dc_client.execute_graphql(DELETE_ALL)
-        assert delete_resp.data['email_deleteMany'] == 1
-        assert delete_resp.data['user_deleteMany'] == 2
 
     def test_execute_graphql_query(self):
         """Tests executing a query via execute_graphql."""
         dc_client = dataconnect.client(CONNECTOR_CONFIG)
         resp = dc_client.execute_graphql(QUERY_LIST_USERS)
-        assert resp.data['users'] == INITIAL_STATE['users']
+        assert sorted(resp.data['users'], key=lambda x: x['id']) == sorted(
```
stephenarosaj: nit: instead of `x` can we use `user`

integration/test_data_connect.py
stephenarosaj: let's test queries first, because it seems we are querying in this function before having tested it

integration/test_data_connect.py
```diff
@@ -156,15 +145,14 @@ def test_execute_graphql_mutation(self):
         )
         assert query_email_resp.data['email'] == FRED_EMAIL
```
stephenarosaj: is the initial data seeded for `FRED_EMAIL` the same as what we upsert it to be? because if so, it's a bit hard to see if our `UPSERT_FRED_EMAIL` had any effect

integration/test_data_connect.py
stephenarosaj: i think you might be missing an test case analogous to this one but for `read`

tests/test_data_connect.py
```diff
@@ -879,7 +877,290 @@ def test_parse_graphql_response_non_dict_error(self):
...
+        self.dc_client = dataconnect.DataConnect(self.app, BASE_CONFIG)
+
+    def teardown_method(self, method):
+        del method
+        testutils.cleanup_apps()
+
+    @mock.patch.object(dataconnect._DataConnectApiClient, "execute_graphql")
+    def test_execute_graphql_delegates_default(self, mock_execute_graphql):
+        mock_execute_graphql.return_value = dataconnect.ExecuteGraphqlResponse(data={"key": "val"})
+        res = self.dc_client.execute_graphql("query { hello }")
```
stephenarosaj: can we factor out the string `"query { hello }"` to a variable which we use throughout the different tests, that way it's easier to read / verify the behavior of this and other tests?

there are some other places we can make the same kind of improvement in the rest of this file (with query strings and with returned data)

tests/test_data_connect.py
```diff
@@ -879,7 +877,290 @@ def test_parse_graphql_response_non_dict_error(self):
...
+        with pytest.raises(ValueError, match="'unauthenticated' must be a boolean."):
+            dataconnect.Impersonation(unauthenticated="not-a-bool")
+
+    def test_constructor_invalid_auth_claims_type(self):
+        """Tests non-dict auth_claims raises ValueError."""
+        with pytest.raises(ValueError, match="'auth_claims' must be a dictionary."):
+            dataconnect.Impersonation(auth_claims="not-a-dict")
+
+
+class TestDataConnectExecuteGraphql:
```
stephenarosaj: the tests in this class are fairly trivial - we're just testing "does it call the wrapper function" when that's literally _all_ the execute functions do - it might be better to omit them, because ideally we should be testing behaviors not implementations - the internal client tests should suffice

tests/test_data_connect.py
```diff
@@ -879,7 +877,290 @@ def test_parse_graphql_response_non_dict_error(self):
...
+        class User:
+            name: str
+
+        options = dataconnect.GraphqlOptions(variables={"name": "Fred"})
+        with pytest.raises(ValueError, match="variables must be of type User"):
+            self.api_client.execute_graphql("query { foo }", options=options, variables_type=User)
+
+
+    @mock.patch.object(dataconnect._DataConnectApiClient, "_make_gql_request")
+
```
stephenarosaj: nit: extra newline here?

tests/test_data_connect.py
```diff
@@ -879,7 +877,290 @@ def test_parse_graphql_response_non_dict_error(self):
...
+    def test_execute_graphql_success(self, mock_make_gql_request):
+        mock_make_gql_request.return_value = {"data": {"foo": "bar"}}
+        res = self.api_client.execute_graphql("query { foo }")
+        assert res.data == {"foo": "bar"}
+        mock_make_gql_request.assert_called_once()
```
stephenarosaj, line 922: just realizing this now, but for all of the tests let's put the `assert_called_once` check before the data equality check 

we probably care first that this was properly called, and second that it returned the expected data (because if it didn't return the right data it would fail right at that line, and then it's unclear if it even called the helper function which might be the underlying reason the equality check failed)

stephenarosaj (approved):
LGTM with some changes requested - mostly small stuff, only a few blocking comments

stephenarosaj (changes requested):
Accidentally selected `Approve` but mean `Request changes` - re-submitting review

LGTM with some changes requested - mostly small stuff, only a few blocking comments

firebase_admin/dataconnect.py
mk2023: I added them because Python's `isinstance()` raises a `TypeError` when passed `typing.Any`, which is our default `variables_type` parameter!

firebase_admin/dataconnect.py
mk2023: That's a good idea! It would allow us to strip GraphQL queries that are written using triple quotes with leading/trailing newlines and prevent messy logs & traces! Fixed it!

firebase_admin/dataconnect.py
mk2023: I included both intentionally to balance Python conventions with developer experience. `auth_claims` provides an idiomatic, PEP 8-compliant interface for Python helpers like `Impersonation.authenticated(`). However, I also wanted to support `authClaims` so that developers migrating from Node.js or pasting raw REST JSON payloads can pass camelCase dictionaries natively without hitting arbitrary breaking changes!

tests/test_data_connect.py
mk2023: Fixed!

integration/emulators/seed.sh
mk2023: Oh, I didn't know that! Thanks so much for catching it! To fix this I updated `seed.sh `with a helper method that captures the response body and explicitly checks for a non-empty "errors" field in the JSON payload. And if there is an error message, the script fails!

integration/test_data_connect.py
mk2023: That was a mismatch on my part! I forgot to switch the integration file when I added the seeding file! Switched it!

integration/test_data_connect.py
mk2023: I intentionally didn't duplicate every edge case for read because `execute_graphql_read` shares the exact same payload serialization and impersonation header logic as `execute_graphql`.  I figured it would only be important to include a set of execute_graphql_read tests to verify the endpoint routing and mutation-rejection behavior. To keep the test suite concise, I'm thinking about even deleting `test_execute_graphql_read_impersonated_unauthenticated_fails`, and only keep `test_execute_graphql_read_impersonated_authenticated` to verify that `execute_graphql_read` + impersonation works! However, if you think I should include all respective tests using read, then I will!

tests/test_data_connect.py
mk2023: Great suggestion! I've factored out `"query { hello }"` and `{"key": "val"}` into top-level test constants (`TEST_QUERY` and `TEST_RESPONSE_DATA`) and updated all the delegation and payload serialization tests to use them consistently!

To further make the testing file concise, I also factored out duplicate test data constants (`TEST_URL, TEST_HEADERS, TEST_PAYLOAD, TEST_AUTH_CLAIMS, TEST_VARIABLES`) and consolidated the `UserProfile` and `CreateUserVariables` test dataclasses to module level in `tests/test_data_connect.py`.

tests/test_data_connect.py
mk2023: That's a great point! Fixed!

tests/test_data_connect.py
mk2023: That makes sense! Since `TestDataConnectApiClientExecuteGraphql` already thoroughly tests validation, execution, and error handling, I've removed `TestDataConnectExecuteGraphql` and its delegation tests.

firebase_admin/dataconnect.py
mk2023: After discussion, we decided to remove authClaims to keep the code Pythonic!

integration/test_data_connect.py
mk2023: To keep the integration test suite concise, I removed duplicate edge-case tests for `execute_graphql_read` and retained `test_execute_graphql_read_impersonated_authenticated` to verify that impersonation works with `execute_graphql_read`.

firebase_admin/dataconnect.py
```diff
@@ -115,29 +115,23 @@ def __init__(
...
         if unauthenticated is not None:
             if not isinstance(unauthenticated, bool):
                 raise ValueError("'unauthenticated' must be a boolean.")
             super().__init__(unauthenticated=unauthenticated)
         else:
-            if not isinstance(claims, dict):
+            if not isinstance(auth_claims, dict):
                 raise ValueError("'auth_claims' must be a dictionary.")
-            super().__init__(authClaims=claims)
+            super().__init__(authClaims=auth_claims)
```
stephenarosaj: think there might still be a typo here - should be `auth_claims=auth_claims` right?

stephenarosaj (changes requested):

tests/test_data_connect.py
```diff
@@ -28,13 +27,39 @@
...
     connector="my_connector",
 )
 
+TEST_QUERY = "query { hello }"
+TEST_RESPONSE_DATA = {"foo": "bar"}
+TEST_URL = "https://example.com/endpoint"
+TEST_HEADERS = {"key": "val"}
+TEST_PAYLOAD = {"query": TEST_QUERY}
+TEST_AUTH_CLAIMS = {"sub": "user_123"}
+TEST_VARIABLES = {"foo": "bar"}
```
stephenarosaj: nit: let's make the value for `TEST_VARIABLES` different from `TEST_RESPONSE_DATA`

tests/test_data_connect.py
```diff
@@ -571,37 +563,22 @@ def teardown_method(self, method):
         testutils.cleanup_apps()
 
     def test_prepare_graphql_payload_only_query(self):
-        payload = self.api_client._prepare_graphql_payload("query { hello }", None)
-        assert payload == {"query": "query { hello }"}
+        payload = self.api_client._prepare_graphql_payload(TEST_QUERY, None)
+        assert payload == {"query": TEST_QUERY}
```
stephenarosaj: nit: instead of `{"query": TEST_QUERY}` you could use `TEST_PAYLOAD`

tests/test_data_connect.py
```diff
@@ -1117,19 +962,19 @@ class User:
...
     def test_execute_graphql_payload_omits_empty_fields(self, mock_make_gql_request):
-        mock_make_gql_request.return_value = {"data": {"foo": "bar"}}
-        self.api_client.execute_graphql("query { foo }")
+        mock_make_gql_request.return_value = {"data": TEST_RESPONSE_DATA}
+        self.api_client.execute_graphql(TEST_QUERY)
         mock_make_gql_request.assert_called_once_with(
             url=mock.ANY,
             headers=mock.ANY,
-            payload={"query": "query { foo }"}
+            payload={"query": TEST_QUERY}
```
stephenarosaj: nit: instead of `{"query": TEST_QUERY}` you could use `TEST_PAYLOAD`

stephenarosaj (changes requested):

integration/test_data_connect.py
mk2023: Yes, fixed it!

integration/test_data_connect.py
mk2023: Good catch! I updated the test to mutate using `UPSERT_UPDATED_FRED_EMAIL` (which contains modified fields) to ensure we're testing a real state change, then reset it back to the initial state at the end of the test.

integration/test_data_connect.py
mk2023: That makes sense! Changed it!

tests/test_data_connect.py
mk2023: Okay! That's better!

tests/test_data_connect.py
mk2023: Yes, that's probably easier to understand when reading the code!

tests/test_data_connect.py
mk2023: Changed!

firebase_admin/dataconnect.py
mk2023: I was also confused about this, but the snake_case `auth_claims` is used in the `__init__` signature and the `authenticated()` factory method so the public API remains purely Pythonic. However, because this class inherits from dict, the `super().__init__(authClaims=...)` is required to translate that pythonic argument into the exact camelCase `authClaims` JSON key that the Firebase Data Connect API expects. If we change the `super().__init__` call to snake_case, the backend will reject the payload.

stephenarosaj: This is a great call out - I think we should remedy this, and it shouldn't be too complex. This is how I'm thinking of it:
- `authClaims` is what the backend API expects - this is how the data should be represented when it's turned into JSON and sent over the wire
- `auth_claims` is the Pythonic, user-facing representation of that same data
- If users were to print an instance of the `Impersonation` class at runtime, they would see `authClaims` - but this is the JSON representation, not the Python representation! We should instead be using the `auth_claims` representation here, since that's what users will interact with and it matches the platform via which they are interacting (Python)
- We should keep the Python representation when in user-facing Python land, and use the JSON wire representation when communicating with the backend (instead of communicating with users) - this implies that we need some "translation" between the two
  - Lucky for us, we already have a `_prepare_graphql_payload` method which "Serializes input query and options to JSON-compatible dictionary" ;)

mk2023: Great feedback! After discussion, we decided to not expect `authClaims` at all. Thus, the `validate_impersonation_options` will not be considering `authClaims` as a valid input. 

`_prepare_graphql_payload` now translates `auth_claims` to `authClaims` before building the request payload via a dictionary key popping mechanism.

stephenarosaj (approved):
LGTM

itsrakhil (approved):

firebase_admin/dataconnect.py
```diff
@@ -101,7 +104,70 @@ def __post_init__(self):
...
+
+    Attributes:
+        data: The raw JSON dictionary returned by the GraphQL execution.
+    """
+    data: _Data
+
+
 class DataConnect:
+
+
```
jonathanedey: nit: spacing

.github/workflows/ci.yml
```diff
@@ -54,6 +54,10 @@ jobs:
       run: firebase emulators:exec --only database --project fake-project-id 'pytest integration/test_db.py'
     - name: Run Functions emulator tests
       run: firebase emulators:exec --config integration/emulators/firebase.json --only tasks,functions --project fake-project-id 'CLOUD_TASKS_EMULATOR_HOST=localhost:9499 pytest integration/test_functions.py'
+    - name: Run Data Connect emulator tests
+      run: firebase emulators:exec --config integration/emulators/firebase.json --only dataconnect --project fake-project-id './integration/emulators/seed.sh && DATA_CONNECT_EMULATOR_HOST=localhost:9399 pytest integration/test_data_connect.py --cert=tests/data/service_account.json'
```
jonathanedey: We shouldn't need to pass credentials here since this is only running against emulator but i'm guessing the test suite was complaining for one.

To get around that we override that check in the integration test file if the emulator host is set. For reference see: https://github.com/firebase/firebase-admin-python/blob/b40e7388053428be4da3ed3adf7f74c0c4a57398/integration/test_functions.py#L27-L44

integration/test_data_connect.py
```diff
@@ -0,0 +1,346 @@
...
+
+OPTS_NON_EXISTING_CLAIMS = dataconnect.GraphqlOptions(
+    impersonate=dataconnect.Impersonation.authenticated({
+        'sub': 'non-existing-id',
+        'email_verified': True
+    })
+)
+
+
+class TestExecuteGraphql:
```
jonathanedey, line 178: Since these integration tests are ran against the same project and config that we use for our Admin Node tests we should ensure that they are compatible. The tests can use the same data but we should ensure we populate and clean up data so that they are idempotent similar to node
https://github.com/firebase/firebase-admin-node/blob/f9bff6b2db209edc182fbae27701e127ec5d71b6/test/integration/data-connect.spec.ts#L162-L176

firebase_admin/dataconnect.py
```diff
@@ -101,7 +104,70 @@ def __post_init__(self):
...
+            if not isinstance(unauthenticated, bool):
+                raise ValueError("'unauthenticated' must be a boolean.")
+            super().__init__(unauthenticated=unauthenticated)
+        else:
+            if not isinstance(auth_claims, dict):
+                raise ValueError("'auth_claims' must be a dictionary.")
+            super().__init__(auth_claims=auth_claims)
+
+    @staticmethod
+    def unauthenticated() -> 'Impersonation':
```
jonathanedey: Slight nit: Can we use `from __future__ import annotations` instead of a string here?

firebase_admin/dataconnect.py
```diff
@@ -101,7 +104,70 @@ def __post_init__(self):
...
+    """Represents impersonation configuration for DataConnect requests.
+
+    It is recommended to construct instances using the static factory methods
+    :meth:`unauthenticated` or :meth:`authenticated`.
+    """
+
+    def __init__(
+        self,
+        *,
+        unauthenticated: Optional[bool] = None,
```
jonathanedey, line 119: No action item here just thinking aloud, i know we are discouraging the constructor uses here but should we have a default state of unauthenticated=True here instead?

jonathanedey (changes requested):
Thanks @mk2023! This overall looks great, with a few comments mainly on testing! I've also added the integration test tag so those tests should run from your next commit.

firebase_admin/dataconnect.py
mk2023: Fixed!

firebase_admin/dataconnect.py
mk2023: I'm not sure if that's a good idea--defaulting `unauthenticated=True` in `Impersonation.__init__` creates a risky API because calling `Impersonation()` without arguments would silently strip authentication credentials without explicit developer intent. Technically, if `unauthenticated` defaulted to `True`, calling `Impersonation(auth_claims={...})` would also implicitly trigger a validation error for providing mutually exclusive parameters (both `unauthenticated` and `auth_claims`)!

firebase_admin/dataconnect.py
mk2023: Done!

.github/workflows/ci.yml
mk2023: Thanks for letting me know!
I removed the credentials flag from `ci.yml`! Unlike `test_functions.py` which explicitly passes a named app instance to every call, `test_data_connect.py` relies on `dataconnect.client()` resolving the default app, so `default_app` initializes the default app directly instead of using pass. Let me know if this isn't optimal!

firebase_admin/dataconnect.py
jonathanedey: Sounds good, lets leave it as is then.

.github/workflows/ci.yml
jonathanedey: Yeah, that sounds like it should be fine as long as we define the `default_app(request)` fixture to use the same `integration_conf(request)` fixture override.

integration/test_data_connect.py
mk2023: Done! I added a `setup_and_cleanup_database` pytest fixture (`autouse=True`) to match Node's beforeEach / afterEach lifecycle. To avoid calling `execute_graphql()` during setup and teardown before tests run, the fixture executes raw HTTP bash scripts (`seed.sh` and `cleanup.sh` via curl) before and after each test. This populates initial state (fred_id, jeff_id, email_id) before each test and wipes the database clean (email_deleteMany, user_deleteMany) afterwards, ensuring 100% test idempotency! Let me know if you think this is okay!

.github/workflows/ci.yml
mk2023: Awesome, thanks! That's how I defined `default_app(request)` in integration/test_data_connect.py!

mk2023: Update: Defining `default_app` in `test_data_connect.py` caused session app collisions (`ValueError: The default Firebase app already exists`) during full integration test runs.
Thus, I aligned the code with existing codebase conventions to use a named app fixture and set `default_app` to `pass`. I also had to add a `dc_client(app)` fixture to inject the client into tests rather than instantiating `dataconnect.client()` in every method. Now, my code adheres to established project conventions and significantly reduces boilerplate code!

Additionally, the `stage_release` check revealed tests were attempting to run against live GCP, so I added a `check_emulator` fixture to safely skip execution when `DATA_CONNECT_EMULATOR_HOST` is absent.

integration/emulators/cleanup.sh
```diff
@@ -0,0 +1,22 @@
...
+  local response
+  response=$(curl -s -X POST "${ENDPOINT}" \
+    -H "Authorization: Bearer owner" \
+    -H "Content-Type: application/json" \
+    -d "${payload}")
+
+  if [ -z "${response}" ]; then
+    echo "Failed to receive response from Data Connect emulator at ${ENDPOINT}" >&2
+    exit 1
+  fi
```
stephenarosaj: this one should also have an `errors` check just like the other one

actually - we shouldn't duplicate this code at all. how about we instead make a single `setup_teardown.sh` file, so the code can be shared between each use case?

.github/workflows/ci.yml
mk2023: Update: `stage_release` check is supposed to fail because we do not have a prod project set up to run tests against. Deleted `check_emulator`!

integration/test_data_connect.py
```diff
@@ -14,9 +14,42 @@
...
+
+@pytest.fixture(scope='module', autouse=True)
+def default_app():
+    # Overwrites the default_app fixture in conftest.py.
+    # This test suite should not use the default app. Use the app fixture instead.
+    pass
+
+@pytest.fixture(scope='module', autouse=True)
+def check_emulator():
+    """Skips Data Connect integration tests if emulator host is not set."""
```
stephenarosaj: we shouldn't be skipping the integration tests if the emulator host isn't set - in that case, we should be running against production!

for this, i think we'll need to initialize the FDC service in the project for the service key we use for github actions test runs - and then deploy the schema + connectors we have in this repo to that project. i believe only Lahiru and @jonathanedey can do this

i think we should either remove the `stage release` tag or get that project set up

EDIT: according to [this comment](https://github.com/firebase/firebase-admin-python/pull/970#discussion_r3715718969), the project used here and in the node admin SDK are actually the same - so there's no need to set anything up!

stephenarosaj: for now, we can test against production manually using the same project we use for manual testing in the admin node SDK - i'll share the cert files with you

integration/emulators/cleanup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
stephenarosaj: something i forgot to think about: placement of these files. they should probs live in the `integration/dataconnect` folder, so it's clear these are ONLY for fdc

stephenarosaj (changes requested):

integration/test_data_connect.py
stephenarosaj: oops i missed this comment - this changes what i originally said in [this comment](https://github.com/firebase/firebase-admin-python/pull/970#discussion_r3717177377) - if both SDKs are using the same project, we shouldn't have to set up any fdc service or deploy anything - they should already be set up from the node SDK!

integration/test_data_connect.py
jonathanedey: Good point, sorry about that folks, I'll remove the tag to unblock this until we can get that setup correctly.

integration/emulators/cleanup.sh
mk2023: That makes sense! I moved the new seeding file, setup_teardown.sh, into integration/emulators/dataconnect/!

integration/emulators/cleanup.sh
mk2023: Done! I consolidated `seed.sh` and `cleanup.sh` into `integration/emulators/dataconnect/setup_teardown.sh`. It shares the `send_gql_mutation` helper with full error checking for both setup and teardown actions.

integration/test_data_connect.py
mk2023: Thank you so much!

integration/test_data_connect.py
```diff
@@ -122,6 +149,16 @@
...
+@pytest.fixture(autouse=True)
+def setup_and_cleanup_database():
+    """Initializes database via setup_teardown.sh before each test and wipes it afterwards."""
+    script_dir = os.path.dirname(__file__)
+    script = os.path.join(script_dir, 'emulators', 'dataconnect', 'setup_teardown.sh')
+
+    subprocess.run(['bash', script, 'setup'], check=True)
+    yield
+    subprocess.run(['bash', script, 'teardown'], check=True)
+
```
jonathanedey, line 159: Can we use existing SDK methods instead of a script here:

```py
@pytest.fixture(autouse=True)
def setup_and_cleanup_database(dc_client):
    dc_client.execute_graphql(UPSERT_FRED_USER)
    dc_client.execute_graphql(UPSERT_JEFF_USER)
    dc_client.execute_graphql(UPSERT_FRED_EMAIL)
    yield
    dc_client.execute_graphql(DELETE_ALL)
```

I think I misunderstood the purpose of the script originally since I assumed that it was emulator specific setup but that isn't the case and the above should should satisfy both. We should avoid accessing credentials through scripts like these. This might explain the failing integration tests since credential setup is handled differently on GitHub so the script won't actually pull credentials as expected.

mk2023: Yes, that makes sense! Just to make sure--should I be using `dc_client` inside the fixture (like calling `dc_client.execute_graphql(UPSERT_FRED_USER))`?

I wasn't sure if it's good testing practice to use the methods in `dataconnect.py` to set up and tear down test state before running the actual tests on them!

jonathanedey: Yup thats totally fine! The goal of these tests is catch end-to-end service and SDK issues early. If this setup or teardown were to fail in the future we would still be able to identify the cause through logs.

mk2023: Sounds good! I changed `setup_and_cleanup_database`!

jonathanedey (approved):
LGTM Thanks!
I'll leave the option open for triggering the integration tests (just let me know and I can re-add the tag) but this looks good to merge to `wine` from Admin SDK side.

stephenarosaj (approved):
LGTM - great job @mk2023 :)

integration/test_data_connect.py
mk2023: Update: Decided to go with calling `execute_graphql` here since it will still identify the cause of error (through logs).

[Link to different comment thread](https://github.com/firebase/firebase-admin-python/pull/970#discussion_r3723408974)

PR #976, opened 2026-08-17
feat(appcheck): Verify one-time tokens for replay protection

yvonnep165:
This PR adds support for App Check one-time token verification for replay protection by adding an optional `consume` parameter to `app_check.verify_token()`. The returned claims dictionary will contain an `already_consumed` boolean key indicating whether the token was previously consumed.

### DIFF
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -27,7 +27,7 @@ def _get_app_check_service(app) -> Any:
     return _utils.get_app_service(app, _APP_CHECK_ATTRIBUTE, _AppCheckService)
 
 def verify_token(token: str, app=None, consume: bool = False) -> Dict[str, Any]:
-    """Verifies a Firebase App Check token.
+    """Verifies a Firebase App Check token, optionally consuming limited-use tokens.
 
     Args:
         token: A token from App Check.
@@ -83,7 +83,7 @@ class _AppCheckService:
 
 
     def verify_token(self, token: str, consume: bool = False) -> Dict[str, Any]:
-        """Verifies a Firebase App Check token."""
+        """Verifies a Firebase App Check token, optionally consuming limited-use tokens."""
         _Validators.check_string("app check token", token)
 
         # Obtain the Firebase App Check Public Keys
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -107,7 +107,9 @@ class _AppCheckService:
             except requests.exceptions.RequestException as error:
                 raise _utils.handle_requests_error(error)
 
-            already_consumed = body.get('alreadyConsumed', False) if isinstance(body, dict) else False
+            already_consumed = False
+            if isinstance(body, dict):
+                already_consumed = body.get('alreadyConsumed', False)
             verified_claims['already_consumed'] = bool(already_consumed)
 
         return verified_claims
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -100,6 +100,8 @@ class _AppCheckService:
 
         verified_claims['app_id'] = verified_claims.get('sub')
 
+        if not isinstance(consume, bool):
+            raise ValueError('consume must be a boolean.')
         if consume:
             url = self._VERIFY_URL_FORMAT.format(project_id=self._project_id)
             try:
--- a/firebase_admin/app_check.py
+++ b/firebase_admin/app_check.py
@@ -15,38 +15,47 @@
 """Firebase App Check module."""
 
 from typing import Any, Dict
+import requests
 import jwt
 from jwt import PyJWKClient, ExpiredSignatureError, InvalidTokenError, DecodeError
 from jwt import InvalidAudienceError, InvalidIssuerError, InvalidSignatureError
-from firebase_admin import _utils
+from firebase_admin import _http_client, _utils
 
 _APP_CHECK_ATTRIBUTE = '_app_check'
 
 def _get_app_check_service(app) -> Any:
     return _utils.get_app_service(app, _APP_CHECK_ATTRIBUTE, _AppCheckService)
 
-def verify_token(token: str, app=None) -> Dict[str, Any]:
+def verify_token(token: str, app=None, consume: bool = False) -> Dict[str, Any]:
     """Verifies a Firebase App Check token.
 
     Args:
         token: A token from App Check.
         app: An App instance (optional).
+        consume: If set to ``True``, performs stateful verification with the App Check
+            backend to mark the token as consumed for replay protection. Defaults to ``False``.
 
     Returns:
-        Dict[str, Any]: The token's decoded claims.
+        Dict[str, Any]: The token's decoded claims. If ``consume`` is ``True``, the dictionary
+            also includes an ``already_consumed`` boolean key indicating whether the token was
+            previously consumed.
 
     Raises:
         ValueError: If the app's ``project_id`` is invalid or unspecified,
-        or if the token's headers or payload are invalid.
+            or if the token's headers or payload are invalid.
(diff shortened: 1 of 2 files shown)


firebase_admin/app_check.py
```diff
@@ -87,6 +99,19 @@ def verify_token(self, token: str) -> Dict[str, Any]:
                 ) from exception
 
         verified_claims['app_id'] = verified_claims.get('sub')
+
+        if consume:
```
jonathanedey, line 105: Let's validate `consume` is a bool here.

jonathanedey (approved):
LGTM with one comment, Thanks!

firebase_admin/app_check.py
yvonnep165: Thanks for taking a look at this! Validation is added.

weixifan (approved):
Thanks for this PR!

kevinthecheung (approved):

firebase_admin/app_check.py
```diff
@@ -87,6 +99,21 @@ def verify_token(self, token: str) -> Dict[str, Any]:
                 ) from exception
 
         verified_claims['app_id'] = verified_claims.get('sub')
+
+        if not isinstance(consume, bool):
```
lahirumaramba: Let's move this to the beginning of the function (along side of `check_string`) that way we fail fast for invalid inputs before obtaining the key sets. If you think this check will be reused, take a look at the `check_boolean`  helper in `firebase_admin/_messaging_encoder.py`

firebase_admin/app_check.py
```diff
@@ -87,6 +99,21 @@ def verify_token(self, token: str) -> Dict[str, Any]:
...
+            raise ValueError('consume must be a boolean.')
+        if consume:
+            url = self._VERIFY_URL_FORMAT.format(project_id=self._project_id)
+            try:
+                body = self._http_client.body('post', url, json={'app_check_token': token})
+            except requests.exceptions.RequestException as error:
+                raise _utils.handle_platform_error_from_requests(error)
+
+            already_consumed = False
+            if isinstance(body, dict):
```
lahirumaramba: What if `body` is not a `dictionary`, do we fail silently and return the token as not already consumed?
This could be an issue if the BE response is malformed for some reason and we silently falls back to `already_consumed = False`. WDYT about throwing an exception here?
```ts
if not isinstance(body, dict):
    raise exceptions.UnknownError(
        'Unexpected response from App Check service. '
        f'Expected a JSON object, but got {type(body).__name__}.')
```

firebase_admin/app_check.py
lahirumaramba: I think this also missing unit tests

firebase_admin/app_check.py
```diff
@@ -87,6 +99,21 @@ def verify_token(self, token: str) -> Dict[str, Any]:
                 ) from exception
 
         verified_claims['app_id'] = verified_claims.get('sub')
+
+        if not isinstance(consume, bool):
+            raise ValueError('consume must be a boolean.')
+        if consume:
+            url = self._VERIFY_URL_FORMAT.format(project_id=self._project_id)
```
lahirumaramba: nit: moving the replay verification to a separate helper method will make the main verify token function less cluttered. 

WDYT about something like:
```ts
def _verify_replay_protection(self, token: str) -> bool:
    url = self._VERIFY_URL_FORMAT.format(project_id=self._project_id)
    try:
        body = self._http_client.body('post', url, json={'app_check_token': token})
    except requests.exceptions.RequestException as error:
        raise _utils.handle_platform_error_from_requests(error)
    except ValueError as error:
        raise exceptions.UnknownError(f'Unexpected response from App Check service: {error}')

    if not isinstance(body, dict):
        raise exceptions.UnknownError(
            'Unexpected response from App Check service. '
            f'Expected a JSON object, but got {type(body).__name__}.')
    return bool(body.get('alreadyConsumed', False))
```

firebase_admin/app_check.py
```diff
@@ -87,6 +99,21 @@ def verify_token(self, token: str) -> Dict[str, Any]:
...
+            try:
+                body = self._http_client.body('post', url, json={'app_check_token': token})
```
lahirumaramba: Add a test to verify when `consume=False` (the default), no HTTP call is made to `_http_client.body`, and `already_consumed` is not present in the returned dictionary.

tests/test_app_check.py
```diff
@@ -273,3 +274,62 @@ def test_verify_token_with_incorrect_issuer_raises_error(self, mocker):
...
+        mocker.patch("jwt.decode", return_value=JWT_PAYLOAD_SAMPLE)
+        mocker.patch("jwt.PyJWKClient.get_signing_key_from_jwt", return_value=PyJWK(signing_key))
+        mocker.patch("jwt.get_unverified_header", return_value=JWT_PAYLOAD_SAMPLE.get("headers"))
+        app = firebase_admin.get_app()
+        app_check_service = app_check._get_app_check_service(app)
+
+        req_exc = requests.exceptions.RequestException("Backend error")
+        mocker.patch.object(app_check_service._http_client, "body", side_effect=req_exc)
+
+        with pytest.raises(exceptions.FirebaseError):
```
lahirumaramba: nit: can we test for more specific error types instead of checking the more broader FirebaseError?

tests/test_app_check.py
```diff
@@ -273,3 +274,62 @@ def test_verify_token_with_incorrect_issuer_raises_error(self, mocker):
 
         expected = 'Token does not contain the correct "iss" (issuer).'
         assert str(excinfo.value) == expected
+
+    def test_verify_token_with_consume_true_not_consumed(self, mocker):
+        mocker.patch("jwt.decode", return_value=JWT_PAYLOAD_SAMPLE)
+        mocker.patch("jwt.PyJWKClient.get_signing_key_from_jwt", return_value=PyJWK(signing_key))
+        mocker.patch("jwt.get_unverified_header", return_value=JWT_PAYLOAD_SAMPLE.get("headers"))
```
lahirumaramba: Can these `mocker.patch` calls be grouped into a pytest fixture and be reused in here, `... already_consumed `, and ` ...backend_error`?

firebase_admin/app_check.py
```diff
@@ -15,38 +15,47 @@
...
     Returns:
-        Dict[str, Any]: The token's decoded claims.
+        Dict[str, Any]: The token's decoded claims. If ``consume`` is ``True``, the dictionary
+            also includes an ``already_consumed`` boolean key indicating whether the token was
+            previously consumed.
 
     Raises:
         ValueError: If the app's ``project_id`` is invalid or unspecified,
-        or if the token's headers or payload are invalid.
+            or if the token's headers or payload are invalid.
```
lahirumaramba: `ValueError: If ``consume`` is not a boolean, or if the app's ``project_id``...`

lahirumaramba (changes requested):
Thanks! Looks great!
I added a few comments, let me know what you think.

firebase_admin/app_check.py
yvonnep165: Done! Added `check_boolean` to `_Validators` and moved the validation to the top before doing any JWKS/crypto operations. Also added unit tests for non-boolean consume arguments.

firebase_admin/app_check.py
yvonnep165: Great catch! I've updated the logic to raise `exceptions.UnknownError` if body is not a dict, and added unit test coverage for malformed/non-dict responses.

firebase_admin/app_check.py
yvonnep165: Added.

lahirumaramba (approved):
LGTM! Thank you! :)

PR #980, opened 2026-08-26
feat(fcm): Migrate topic management to FCM v1 API

lahirumaramba:
Migrates `subscribe_to_topic` and `unsubscribe_from_topic` in the messaging module from the legacy Instance ID (IID) API to the FCM v1 Topic Subscriptions API.

Key changes:
- Updated `subscribe_to_topic` and `unsubscribe_from_topic` to call the FCM v1 endpoints (`/v1/projects/{projectId}/registrations/{token}/topicSubscriptions`).
- Added asynchronous counterparts `subscribe_to_topic_async` and `unsubscribe_from_topic_async` utilizing `HttpxAsyncClient` (HTTP/2).
- Added deprecated legacy methods `subscribe_to_topic_legacy` and `unsubscribe_from_topic_legacy` with deprecation warnings.
- Bounded client-side concurrency to 100 simultaneous requests.
- Handled idempotency: `ALREADY_EXISTS` (HTTP 409) is treated as success for topic subscriptions, while `NOT_FOUND` (HTTP 404) is recorded as a failure for topic unsubscriptions.


firebase_admin/messaging.py
```diff
@@ -499,6 +611,225 @@ async def send_data(data):
...
+            return status_map[status_code]
+
+        if isinstance(error_data, dict):
+            message = error_data.get('message')
+            if message:
+                return message
+
+        return 'UNKNOWN_ERROR'
+
+    def _build_topic_subscription_result(self, response, is_subscribe):
```
jonathanedey, line 790: Not blocking but is there a reason we don't reuse the `_build_fcm_error` and `_build_fcm_error_httpx` methods to parse the responses here. I think I follow that there needs to be a pre-check based on `is_subscribe`. Is there anything else?

OrlandriaH-G (approved):

jonathanedey (approved):
LGTM with one comment on error parsing which isn't blocking.

yvonnep165 (approved):
LGTM!

firebase_admin/messaging.py
lahirumaramba: I think it was because we had to map the error codes to keep backward compatibility with the legacy subscription methods. I will add a task to clean this up in the next major version. Thanks!

PR #983, opened 2026-09-10
fix(rtdb): Preserve server-side key order in order_by_key() query results

### LINKED ISSUE #677: Order diff between backend and client with order_by_key()
### Describe your environment

  * Operating System version: macOS 12.4
  * Firebase SDK version: 6.1.0
  * Firebase Product: database
  * Python version: 3.7.9
  * Pip version: 21.3.1

### Describe the problem

When using order_by_key(), the order or the returned `OrderedDict` is not the same as the database making pagination very difficult

#### Steps to reproduce:

- Create a collection with this json in "/test" for example : 
```
{
  "123": {"myValue": true},
  "100001": {"myValue": true},
  "100002": {"myValue": true},
  "100003": {"myValue": true},
  "100004": {"myValue": true},
  "100005": {"myValue": true}
}
```
- Query it with :
```python
db.reference("/test").order_by_key().limit_to_first(4).get()
```
- Comparison : 

Current : 
```
OrderedDict([('100001', {'myValue': True}), ('100002', {'myValue': True}), ('100003', {'myValue': True}), ('123', {'myValue': True})])
```
Expected : 
```
OrderedDict([('123', {'myValue': True}), ('100001', {'myValue': True}), ('100002', {'myValue': True}), ('100003', {'myValue': True})])
```

#### Comments

So according to this [link](https://firebase.google.com/docs/database/admin/retrieve-data#orderbykey), order should be the same as the collection but in the corresponding `OrderedDict` the order is not the same so if you want make a pagination strategy, it just loops as getting the last key of the `OrderedDict` returned and using it in `start_at` will return the same result over and over again.

Sanjays2402:
## Problem

`Reference.order_by_key()` queries return children in the wrong order whenever keys are integer-like (e.g. `"123"`, `"100001"`). The RTDB server returns `orderBy="$key"` results in Firebase key order — integer-parseable keys first, numerically, then string keys lexicographically — but `Query.get()` re-sorted the payload client-side with a pure lexicographic comparison, scrambling the server order. This also breaks keyset pagination: feeding the last returned key back into `start_at()` returns the same page forever.

Fixes #677

## Solution

Skip the client-side sorter for `order_by_key()` queries and return the server-ordered payload as an `OrderedDict`, preserving the documented `Query.get()` return type ("a sorted list or an OrderedDict"). Ordering by child/value still goes through the client-side sorter, and `order_by='$priority'` behavior is unchanged. Also updated the `Query` class docstring, which previously claimed the server always returns an unordered collection.

## Testing

- Reproduced the bug from #677 with a standalone script simulating the server payload (`order_by_key().limit_to_first(4)` over `{"123", "100001", "100002", "100003"}`): before the fix the client returned `['100001', '100002', '100003', '123']`; after the fix it returns the server order `['123', '100001', '100002', '100003']`.
- Added two unit tests in `tests/test_db.py` (`TestQuery`): a regression test asserting `order_by_key().get()` preserves server key order, and a guard test asserting `order_by_child().get()` is still sorted client-side.
- Full `tests/test_db.py` suite: 476 passed (6 pre-existing SSE-listener errors also present on unmodified `main`; verified via `git stash`).
- `pylint --disable=all --enable=E` on both touched files: no errors.

Context Sources Used:
- id: firebase-admin-python (repo AGENTS.md)


Sanjays2402:
Closing as a duplicate of #982 — same fix for #677, opened in error. Please review #982 instead.

PR #984, opened 2026-09-10
fix(rtdb): Match backend key ordering in order_by_key() query results

### LINKED ISSUE #677: Order diff between backend and client with order_by_key()

Sanjays2402:
Fixes #677

### Problem
`db.reference(path).order_by_key().get()` returned children in plain lexicographic key order, while the Realtime Database backend orders keys that parse as 32-bit integers first (ascending numeric order) and the remaining keys lexicographically. The client/server order mismatch breaks key-based pagination: using the last key of a page in `start_at()` can return the same page repeatedly.

### Solution
Taught the client-side query sorter (`_SortEntry` in `firebase_admin/db.py`) the backend key ordering rules, mirroring the key comparator used by the other Firebase SDKs:
- keys parseable as signed 32-bit integers sort first, by numeric value ascending (numeric ties broken by key length, shorter first),
- all other keys follow in lexicographic order.

The change is scoped to `order_by == '$key'` with string keys; `$value`, child-path, and `$priority` ordering are untouched.

### Testing
- Reproduced the exact scenario from #677 before the fix (client returned `['100001', ..., '123']` vs backend `['123', '100001', ...]`); after the fix the client order matches the backend order.
- Extended `TestSorter::test_order_by_key` in `tests/test_db.py` with cases covering numeric ordering (`'2' < '10'`), negative keys, leading-zero ties, integer-before-string ordering, and keys outside the 32-bit range.
- `pytest tests/test_db.py`: 486 passed. `./lint.sh` (pylint): 10.00/10 on changed files.
- Full suite: one pre-existing, unrelated failure in `tests/test_http_client.py::TestHttpxAsyncClient::test_request` (httpx version issue in this environment; fails identically on the pristine tree).

Context Sources Used:
- id: firebase-admin-python


Sanjays2402:
Closing as a duplicate of #982 — same fix for #677, opened in error. Please review #982 instead.

PR #986, opened 2026-09-22
fix(fcm): reject empty message batches before dispatch

### LINKED ISSUE #792: Better Error handling for empty token list in Messaging send_each_multicast
**Problem**
Non-descriptive error message returned when sending multicast message with empty device token list. Currently this propagates throughout the code and results in an Unknown Error type with a vague error message as follows: "Unknown error while making remote service calls: max_workers must be greater than 0". This error seems to originate [here](https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/messaging.py#L443) as the Threadpool is created with max_workers based on the length of the messages parameter, which is ultimately dependent on the number of tokens sent out in the Message object.

**Describe the solution you'd like**
Check the length of device tokens directly in the send_each_multicast method and return a descriptive ValueError in the event of an empty list. [here](https://github.com/firebase/firebase-admin-python/blob/master/firebase_admin/messaging.py#L161)

**Describe alternatives you've considered**
Check the length of the message directly in the send_each class method [here](https://github.com/firebase/firebase-admin-python/blob/32b900b0b7f9d8f453f821bf684bf7023c799808/firebase_admin/messaging.py#L425).

Shubham-Padkonde:
Fixes #792.

An empty message list currently reaches a zero-worker thread pool and raises UnknownError in the synchronous sender, while the asynchronous sender returns an empty response. Reject empty batches with a clear ValueError before either dispatch path starts. This also covers multicast messages with no recipients; non-empty token and FID batches retain their existing behavior.

Added regression tests for direct and multicast sending in both sync and async APIs, and documented the non-empty requirement. All six collected regression cases fail on the original implementation.

Validation on Windows / Python 3.13:
- Full unit suite: 3,921 passed (200 deprecation warnings).
- Messaging module: 954 passed.
- Project-configured pylint on both changed files: 10.00/10, no findings.
- git diff --check passed.
- Live Firebase integration tests were not run (no configured project).

Prepared with Codex assistance.

Context Sources Used:
- id: firebase-admin-python; repository path: AGENTS.md


Shubham-Padkonde:
Addressed the multicast error-message feedback in ebf718a. Empty multicast calls now identify `multicast_message` and require at least one token or fid. Tests cover empty tokens, empty fids, and both, for synchronous and asynchronous methods. All 958 messaging tests pass; both modified modules pass pylint (10/10). Prepared with Codex assistance.

PR #987, opened 2026-09-22
fix(appcheck): Use v1beta endpoint for verifyAppCheckToken

lahirumaramba:
Regression from https://github.com/firebase/firebase-admin-python/pull/976

Updated `_AppCheckService._VERIFY_URL_FORMAT` in `firebase_admin/app_check.py` to use the `v1beta` endpoint (`https://firebaseappcheck.googleapis.com/v1beta/projects/{project_id}:verifyAppCheckToken`) when verifying token replay protection (`consume=True`), as `verifyAppCheckToken` is publicly exposed under `v1beta`.


yvonnep165 (approved):
LGTM! Thanks!

yvonnep165:
Also I want to confirm that the PR title should be feat(appcheck) or feat(fac) for the appcheck team related matters? Cause I saw a mix use of both

lahirumaramba:
> Also I want to confirm that the PR title should be feat(appcheck) or feat(fac) for the appcheck team related matters? Cause I saw a mix use of both

These were used by a previous release note generation tool that we no longer use, so it doesn't matter much. Either prefix is fine, but let's stick with "appcheck" moving forward for consistency :)

